PR #138, opened 2022-05-06
add and use API to check stack discipline in asm wrapper codegen

fw-immunant:


fw-immunant:
This is old and stale. Closing to declutter open PRs.

PR #150, opened 2022-08-12
Add TLS test and requisite support

fw-immunant:
This test creates two compartments and a thread-local variable in each. The compartments should be limited to accessing their own thread-locals, but each can presently access the other's as well. This will serve to validate (at least part of) our thread support when we implement it.


fw-immunant:
This is actually not dependent on supporting thread creation, so I've rebased onto `main`. After discussion with @rinon a couple weeks ago, it seems like storing `ia2_stackptrs` in TLS is actually the best way to scope stack pointers to threads (who'd've thunk) so #183 will get substantially reworked atop of this PR.

cmake/add-tls-padded-library.cmake
```diff
@@ -0,0 +1,54 @@
...
+        pkg_check_modules(LIB REQUIRED lib${WRAPPED_LIB})
+        set(LIB_DIR ${LIB_LIBDIR})
+    else()
+        set(LIB_DIR ${CMAKE_CURRENT_BINARY_DIR})
+    endif()
+
+    # Add command and target to generate the padded, sonamed library
+    add_custom_command(
+        OUTPUT ${OUTPUT_DIR}/lib${LIB_PADDED}.so
+        COMMAND cp ${LIB_DIR}/lib${LIB}.so ${OUTPUT_DIR}/lib${LIB_PADDED}.so
```
ahomescu, line 44: `${CMAKE_COMMAND} -E copy` instead of `cp` is more portable.

cmake/add-tls-padded-library.cmake
```diff
@@ -0,0 +1,54 @@
...
+        COMMAND pad-tls --allow-no-tls ${OUTPUT_DIR}/lib${LIB_PADDED}.so
+        COMMAND patchelf --set-soname lib${LIB_PADDED}.so ${OUTPUT_DIR}/lib${LIB_PADDED}.so
+        DEPENDS pad-tls ${LIB_DIR}/lib${LIB}.so
+        COMMENT "Padding TLS segment of wrapped library"
+    )
+    add_custom_target(${LIB_PADDED}-tgt DEPENDS "${OUTPUT_DIR}/lib${LIB_PADDED}.so")
```
ahomescu, line 50: `add_custom_target` can run the commands directly, do you need a separate `add_custom_command`?

cmake/define-test.cmake
```diff
@@ -103,4 +103,31 @@ function(define_test)
         ${WRAPPERS}
         partition-alloc)
     add_dependencies(check-ia2 ${MAIN})
+
+    # Find libc
+    list(APPEND LIBC_PATHS /lib/libc.so.6 /lib64/libc.so.6 /lib/x86_64-linux-gnu/libc.so.6)
```
ahomescu, line 108: `find_library(LIBC_PATH libc)`?

cmake/define-test.cmake
```diff
@@ -103,4 +103,31 @@ function(define_test)
...
+        endif()
+    endforeach()
+    if(NOT DEFINED LIBC_PATH)
+        string(REPLACE ";" "\t" PATHS "${LIBC_PATHS}")
+        message(FATAL_ERROR
+            "Could not find libc.so.6 in:\n${PATHS}\n")
+    endif()
+
+    # Generate libc with padded TLS segment
+    add_custom_command(
```
ahomescu, line 121: Same comments as above.

header-rewriter/tests/tls_protected/main.c
```diff
@@ -0,0 +1,62 @@
...
+  // Access to thread-local from the same compartment should work.
+  printf("main: main secret is %x\n", main_secret);
+  printf("errno=%d, pkru=%08x\n", errno, rdpkru());
+  lib_print_lib_secret();
+
+  printf("errno=%d, pkru=%08x\n", errno, rdpkru());
+  // If we have an argument, test the "main accessing lib" direction;
+  // otherwise test the "lib accessing main" direction. Both should
+  // exit with an MPK violation.
+  bool access_lib_secret = argc > 1;
```
ahomescu, line 45: Nit: future-proof this by checking the value of `argv[1]`? Pick a specific value to use here? E.g. `--access-lib-secret`?

Even better: check out `getopt` and `getopt_long`.

libia2/ia2.c
```diff
@@ -132,6 +134,37 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
+  for (size_t i = 0; i < info->dlpi_phnum; i++) {
+    Elf64_Phdr phdr = info->dlpi_phdr[i];
+    if (phdr.p_type != PT_TLS) {
+      continue;
+    }
+
+    void *start = info->dlpi_tls_data;
+    size_t len = phdr.p_memsz;
+
+    void *start_round_down = (void *)((uint64_t)start & ~0xFFF);
```
ahomescu, line 147: `~0xFFFULL` to avoid integer casting issues? Or maybe use `PAGE_SIZE - 1` for more clarity and portability?

pad-tls/pad-tls.c
```diff
@@ -0,0 +1,112 @@
+/* pad-tls: tool to pad TLS segment of ELF files so that a given compartment's
```
ahomescu, line 1: Nit: for tools like these, C++ or Rust might make your life easier (although C works fine too).

pad-tls/pad-tls.c
```diff
@@ -0,0 +1,112 @@
...
+    "invalid ELF magic number",
+    "ELF architecture is not x86-64",
+    "no TLS segment in ELF program headers",
+};
+
+enum patch_result patch_tls(void *elf, int64_t delta, uint64_t *size_out) {
+  Elf64_Ehdr *ehdr = elf;
+
+  if (memcmp(ehdr->e_ident,
+             "\x7f"
```
ahomescu, line 33: `elf.h` has `ELFMAG` and `SELFMAG` for exactly this purpose.

pad-tls/pad-tls.c
```diff
@@ -0,0 +1,112 @@
...
+      stderr,
+      "usage: %s [--allow-no-tls] <ELF>: adds %ld bytes to TLS segment size of ELF binary\n",
+      basename(argv[0]), delta);
+}
+
+int main(int argc, char **argv) {
+  char *filename;
+  bool allow_no_tls = false;
+
+  if (argc > 2 && !strcmp(argv[1], "--allow-no-tls")) {
```
ahomescu, line 74: Check out `getopt` and `getopt_long`, they're from POSIX.

pad-tls/pad-tls.c
```diff
@@ -0,0 +1,112 @@
...
+  Elf64_Ehdr *ehdr = elf;
+
+  if (memcmp(ehdr->e_ident,
+             "\x7f"
+             "ELF",
+             4)) {
+    return BAD_MAGIC;
+  }
+
+  if (ehdr->e_machine != EM_X86_64) {
```
ahomescu, line 39: Does this matter?

pad-tls/pad-tls.c
```diff
@@ -0,0 +1,112 @@
...
+      return PATCH_SUCCESS;
+    }
+
+    ph_addr += ehdr->e_phentsize;
+  }
+
+  return NO_TLS;
+}
+
+const int64_t delta = 4096;
```
ahomescu, line 61: This might get constant-folded if you made it `static const`. In C++ you'd just use `constexpr` but C doesn't have that.

Alternatively, could this be a macro?

pad-tls/pad-tls.c
```diff
@@ -0,0 +1,112 @@
...
+  if (ehdr->e_machine != EM_X86_64) {
+    return WRONG_ARCH;
+  }
+
+  char *ph_addr = ((char *)elf) + ehdr->e_phoff;
+  for (int i = 0; i < ehdr->e_phnum; i++) {
+    Elf64_Phdr *phdr = (Elf64_Phdr *)ph_addr;
+
+    if (phdr->p_type == PT_TLS) {
+      phdr->p_memsz += delta;
```
ahomescu, line 48: If this is always 4096, do you need it to be a function argument?

cmake/define-test.cmake
fw-immunant: This (specifically, `find_library(LIBC c REQUIRED)`) doesn't work, because `libc.so` itself is a linker script with glibc, and we need the actual ELF shared object so we can modify its program headers.

fw-immunant: I'll add a comment explaining the workaround.

header-rewriter/tests/tls_protected/main.c
fw-immunant: I agree this is bad, but it's what a bunch of our tests do. We should fix them all in a separate pass, imo. We've had some long-term interest in making our testing setup better (#154) that also specifically calls out this argument parsing.

pad-tls/pad-tls.c
fw-immunant: The whole project is x86_64-only, so using this tool on an ELF with a different architecture means something is wrong.

pad-tls/pad-tls.c
fw-immunant: It's somewhat likely that we'll need to change our padding scheme, possibly on a library-dependent basis, so I figured I should err toward making it easier to make this a command-line parameter eventually, just in case.

pad-tls/pad-tls.c
fw-immunant: It could, but none of this is performance-sensitive so I don't think it matters.

libia2/ia2.c
fw-immunant: The rest of the file uses [`~`]`0xFFFUL`; I'll make this match.

cmake/add-tls-padded-library.cmake
fw-immunant: `add_custom_target` creates a target that is always rebuilt because it has no output files. As such, to avoid always running the copying and patching of the ELF, we need to define those steps with `add_custom_command` so that CMake associates the output file with the commands. Then we can make the output file a dependency of the custom target, and hang the properties like `IMPORTED_LOCATION` off of the custom target to make it behave like a library target.

cmake/define-test.cmake
fw-immunant: The same logic applies here--we need both because we want to create a target that behaves like a library and also associate the output files with the command.

Issue #22, opened 2021-09-24
Support functions that return non-typedefed function pointers


ahomescu:
For now, untrusted functions that return function pointers are not supported. The main reason is that such pointers work in the opposite direction of the rest we handle: returned function pointers point to untrusted functions and therefore need to be wrapped with a transition to that compartment, while function pointer parameters point to the trusted compartment.

PR #151, opened 2022-08-17
Make typedef'd function pointer return types work

rinon:
The error printed seems to imply that we can support returning function
pointers that are a typedef, but we didn't actually check for that case.

### DIFF
--- a/header-rewriter/HeaderRewriter.cpp
+++ b/header-rewriter/HeaderRewriter.cpp
@@ -2,6 +2,7 @@
 #include "GenCallAsm.h"
 #include "TypeOps.h"
 #include "clang/AST/AST.h"
+#include "clang/AST/Type.h"
 #include "clang/ASTMatchers/ASTMatchFinder.h"
 #include "clang/ASTMatchers/ASTMatchers.h"
 #include "clang/Basic/FileManager.h"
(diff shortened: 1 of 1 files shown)


ayrtonm:
Thanks for pointing out `isTypedefedNameType`. I didn't see this function when I changed this from an error to a warning and thought that checking this condition would've been just as involved as fixing the problem with function pointers (#22).

fw-immunant:
What's the difference between [these](https://github.com/immunant/IA2-Phase2/compare/88d6d7131eb1b0b6f717f0d295a1a6b13a3d1102..13796c5a0ba32bb0a92885e8bef228385a76dff3) iterations of this change?

ayrtonm:
> What's the difference between [these](https://github.com/immunant/IA2-Phase2/compare/88d6d7131eb1b0b6f717f0d295a1a6b13a3d1102..13796c5a0ba32bb0a92885e8bef228385a76dff3) iterations of this change?

IIRC the first failed in CI because that function wasn't in whatever version of llvm it used.

rinon:
> > What's the difference between [these](https://github.com/immunant/IA2-Phase2/compare/88d6d7131eb1b0b6f717f0d295a1a6b13a3d1102..13796c5a0ba32bb0a92885e8bef228385a76dff3) iterations of this change?
> 
> IIRC the first failed in CI because that function wasn't in whatever version of llvm it used.

exactly, that's a 12+ API IIRC. The way I do it now should be equivalent for us (modulo a small difference in C++ type aliases).

Issue #18, opened 2021-09-21
Handle variadics


ahomescu:
Some libraries export variadics, which are a problem because we cannot easily copy variadic arguments through wrappers. Examples include:
- Logging/printing functions (example???)
- Polymorphism via untyped arguments (example???)

Potential solution: for wrapped functions that have a `va_list` equivalent, our wrappers can build one and call that function instead of the original.

PR #153, opened 2022-08-19
Do not remove variadic functions from the wrapped headers

rinon:
If we only build against the wrapped headers, without the originals, but
still need to use a variadic function, we need to preserve these
functions in the header. Until #18 is fixed we should still allow users
to call variadics from their current compartment, when possible.


fw-immunant (approved):
This seems fine to me--from a security perspective, if we simply don't wrap a compartment's variadic functions, calls to them from another compartment will fail when they attempt to access data the first compartment's data. Calls within the compartment will be unchanged as you say.

Issue #112, opened 2022-04-11
Run compartment destructors with compartment pkey


endbr64:
There is a crash after main in one of our tests that seems unrelated to pkey switching. This issue is for the second part of #95.

```
dmitriy@ubunut:~/git/IA2-Phase2$ gdb build/header-rewriter/tests/two_keys_minimal/two_keys_minimal-main
...
...
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7fe0f2b in ?? () from /lib64/ld-linux-x86-64.so.2
(gdb) bt
#0  0x00007ffff7fe0f2b in ?? () from /lib64/ld-linux-x86-64.so.2
#1  0x00007ffff7db28d7 in __run_exit_handlers (status=0, listp=0x7ffff7f58718 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true,
    run_dtors=run_dtors@entry=true) at exit.c:108
#2  0x00007ffff7db2a90 in __GI_exit (status=<optimized out>) at exit.c:139
#3  0x00007ffff7d900ba in __libc_start_main (main=0x55555555575c <main>, argc=2, argv=0x7fffffffe398, init=<optimized out>, fini=<optimized out>,
    rtld_fini=<optimized out>, stack_end=0x7fffffffe388) at ../csu/libc-start.c:342
#4  0x000055555555526e in _start ()
(gdb)
```

endbr64:
This no longer seems to be happening, now that the worker process is not crashing. Furthermore, shutting down the server with `./build/nginx -s quit` also exits cleanly.

ayrtonm:
Just to clarify, this no longer happens in nginx. The main process was previously exiting (probably because one of the child processes had an MPK violation) but now that the child process is working as expected, the main process just loops. This issue is still a problem in general when calling exit. Though I'm not exactly sure why `nginx -s quit` exits cleanly.

rinon:
Is this still an active issue? I think this was due to accessing read-only data that is no longer protected.

rinon:
TODO: write a test case with a destructor that accesses writable compartment data and fix that case.

fw-immunant:
GH autoclosed this erroneously.

rinon:
I think we need to rewrite the `.fini_array` destructors dynamically, so that we can handle unmodified, wrapped libraries. In this case we want to redirect the destructors to a function in the wrapper library that sets the pkey and then invokes any destructors in the unmodified library.

PR #156, opened 2022-08-23
Allow the DYNAMIC section to be shared to all compartments

rinon:
The dynamic loader expects that it can read the .fini_array from every
library, so we need to exclude that region from pkey protection. This
isn't ideal, but at least this region should be covered by relro and
thus is not trivially writable by an attacker.

### DIFF
--- a/include/ia2.h
+++ b/include/ia2.h
@@ -294,7 +294,7 @@ static const char *ignored_sections[][2] = {
 
 // The number of ranges that may be ignored by protect_pages, includes an extra
 // slot for the DYNAMIC section
-#define NUM_IGNORED_RANGES IGNORED_SECTION_COUNT + 1
+#define NUM_IGNORED_RANGES (IGNORED_SECTION_COUNT + 1)
 // The number of program headers to allocate space for in protect_pages. This is
 // only an estimate of the maximum value of dlpi_phnum below.
 #define NUM_PHDRS 20
--- a/include/ia2.h
+++ b/include/ia2.h
@@ -286,8 +286,15 @@ struct PhdrSearchArgs {
   const void *address;
 };
 
-// The number of ELF sections that may be ignored by protect_pages
-#define NUM_IGNORED_SECTIONS 2
+static const char *ignored_sections[][2] = {
+    {"__start_ia2_shared_data", "__stop_ia2_shared_data"},
+    {"__start_ia2_shared_rodata", "__stop_ia2_shared_rodata"},
+};
+#define IGNORED_SECTION_COUNT (sizeof(ignored_sections) / sizeof(ignored_sections[0]))
+
+// The number of ranges that may be ignored by protect_pages, includes an extra
+// slot for the DYNAMIC section
+#define NUM_IGNORED_RANGES IGNORED_SECTION_COUNT + 1
 // The number of program headers to allocate space for in protect_pages. This is
 // only an estimate of the maximum value of dlpi_phnum below.
 #define NUM_PHDRS 20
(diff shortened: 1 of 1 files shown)


rinon:
This doesn't fix #112 yet because we still crash when trying to run one of a library's destructors, but it's a step forward.

include/ia2.h
```diff
@@ -286,8 +286,15 @@ struct PhdrSearchArgs {
...
-#define NUM_IGNORED_SECTIONS 2
+static const char *ignored_sections[][2] = {
+    {"__start_ia2_shared_data", "__stop_ia2_shared_data"},
+    {"__start_ia2_shared_rodata", "__stop_ia2_shared_rodata"},
+};
+#define IGNORED_SECTION_COUNT (sizeof(ignored_sections) / sizeof(ignored_sections[0]))
+
+// The number of ranges that may be ignored by protect_pages, includes an extra
+// slot for the DYNAMIC section
+#define NUM_IGNORED_RANGES IGNORED_SECTION_COUNT + 1
```
fw-immunant, line 297: `
```suggestion
#define NUM_IGNORED_RANGES (IGNORED_SECTION_COUNT + 1)
```

include/ia2.h
```diff
@@ -384,6 +388,19 @@ static int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
+  // Add the ignored range for the DYNAMIC section. If the library is built with
+  // relro this section should be part of the relro region, so it at least won't
+  // be trivially writable without an mprotect call.
+  for (size_t i = 0; i < info->dlpi_phnum; i++) {
+    Elf64_Phdr phdr = info->dlpi_phdr[i];
+    if (phdr.p_type == PT_DYNAMIC) {
+      struct AddressRange *dynamic_range = &ignored_ranges[NUM_IGNORED_RANGES - 1];
+      dynamic_range->start = (info->dlpi_addr + phdr.p_vaddr) & ~0xFFFUL;
+      dynamic_range->end =
+          (dynamic_range->start + phdr.p_memsz + 0xFFFUL) & ~0xFFFUL;
```
fw-immunant, line 400: It would be clearer to at least use a `PAGE_SIZE` or `PAGE_SIZE_MASK` macro here.

fw-immunant: Then again, the other code here doesn't, so don't consider this a blocker.

include/ia2.h
```diff
@@ -384,6 +388,19 @@ static int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
   size_t extra_seg_info = info->dlpi_phnum;
 
+  // Add the ignored range for the DYNAMIC section. If the library is built with
+  // relro this section should be part of the relro region, so it at least won't
+  // be trivially writable without an mprotect call.
+  for (size_t i = 0; i < info->dlpi_phnum; i++) {
+    Elf64_Phdr phdr = info->dlpi_phdr[i];
+    if (phdr.p_type == PT_DYNAMIC) {
+      struct AddressRange *dynamic_range = &ignored_ranges[NUM_IGNORED_RANGES - 1];
+      dynamic_range->start = (info->dlpi_addr + phdr.p_vaddr) & ~0xFFFUL;
```
ayrtonm, line 398: We need to ensure that rounding the start down/end up doesn't include sections that should be protected. For ia2_shared_data we do this in libia2/padding.ld, but this is more finicky for some existing sections. There's a prototype that works with GNU ld [here](https://github.com/immunant/IA2-Phase2/pull/121/files#diff-9e28fda1df0859baaa4aed0488c407cba6bb5abac419956850641202e4f00766), but IIRC it doesn't work with lld and might not work with relro since it splits the relro sections have to be contiguous (though I agree that we should use relro).

fw-immunant:
General question on this approach: is the goal to obviate switching to a privileged pkey after `main`, or will we still need to do that, e.g. in light of https://github.com/immunant/IA2-Phase2/issues/112#issuecomment-1112679901?

rinon:
> General question on this approach: is the goal to obviate switching to a privileged pkey after `main`, or will we still need to do that, e.g. in light of [#112 (comment)](https://github.com/immunant/IA2-Phase2/issues/112#issuecomment-1112679901)?

We probably will still need to switch, but to switch before invoking anything in the library would have required dynamic loader changes. If we at least dispatch to fini functions in the library we can have a chance to hook them first.

ayrtonm:
I don't remember all the details of #112 off-hand, but I think there's some stuff added to .data by linking with libc/the linker that also needs to be shared to fix the problem (the issue mentions `__dso_handle`). We handled a similar issue for `environ` in #121, so we added it [here](https://github.com/immunant/IA2-Phase2/blob/am/nginx-rewrite/libia2/padding.ld#L147) and initialized like [this](https://github.com/immunant/IA2-Phase2/blob/am/nginx-rewrite/examples/nginx/src/core/nginx.c#L209-L221) on that branch.

endbr64:
I also remember `DYNAMIC` being the first step of many. I think we eventually ended up having to share `.data`, which we agreed was a nonstarter.

ayrtonm:
I don't think we ever tried moving just `__dso_handle` to ia2_shared_data.

rinon:
To do this right we really need to switch compartments or remove compartment protections, because destructors might do arbitrary things within their DSO. Just fixing `__dso_handle` seems like a hack that isn't going to get us very far.

rinon:
Subsumed by #170

PR #157, opened 2022-08-23
Move header #include generation into the FnPtrPrinter pass

endbr64:
After we removed the .symver wrapper, some headers
stopped being inserted properly. This change moves
the header logic into a pass that always runs.

### DIFF
--- a/header-rewriter/HeaderRewriter.cpp
+++ b/header-rewriter/HeaderRewriter.cpp
@@ -252,14 +252,6 @@ public:
     }
     clang::FileEntryRef header_ref = *header_ref_result;
 
-    // Add ia2.h and the output header to the header being rewritten
-    if (!isInitialized(header_ref)) {
-      if (addHeaderImport(header_name, OutputHeader)) {
-        return;
-      }
-      InitializedHeaders.push_back(header_ref);
-    }
-
     // Generate a wrapper for this function
     WrappedFunctions.push_back(FunctionWrapper(fn_decl));
   }
@@ -271,8 +263,6 @@ public:
 
 private:
   std::map<std::string, Replacements> &FileReplacements;
-  // Headers that have included the IA2 header
-  std::vector<clang::FileEntryRef> InitializedHeaders;
   // The wrappers for functions that have been wrapped so far.
   std::vector<FunctionWrapper> WrappedFunctions;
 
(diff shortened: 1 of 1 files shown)


ayrtonm:
@rinon @fw-immunant #158 provides some context on this PR. Basically the 5th point in that list was previously handled in the FnDecl rewriter pass and this moves it to the FnPtrPrinter pass. At some point while @endbr64 was rebasing the nginx branch, it seems like one of the rewriter invocations we used for nginx was causing the FnDecl pass to return before adding those `#include`s. This invocation used `--omit-wrappers` which caused it to skip all fn declarations, so the FnDecl pass was expected to return early.

Moving it to FnPtrPrinter fixes it because it ensures that ia2.h and the output header are always included if a header has any fn ptr types rewritten. For cases where there are only function declarations and no function pointers, this will not include those headers which shouldn't be an issue since function declarations don't trigger changes in the headers.

fw-immunant (commented):
> After we removed the .symver wrapper, some headers stopped being inserted properly. This change moves the header logic into a pass that always runs.

I was confused here for a couple reasons. The first is that there's the confusing terminology of a visitor `run`ning: both visitors unconditionally have their matchers applied to the `ASTMatchRefactorer`, so they do "run" over the AST of all files. But their `run` method does indeed only get executed when their matchers match.

The second is that (in the latter, "match" sense of `run`), the `FnPtrPrinter` visitor does not match all headers. It only matches anywhere in the headers that contain fn ptr params/returns/typedefs. Prior to this change, we assumed that the `FnDecl` visitor would trigger on a strict superset of the files that the `FnPtrPrinter` visitor would, but this isn't actually true (e.g. for files containing only a typedef declaration, and no function declarations).

That said, I don't think we actually want to include `ia2.h` in exactly the set of headers where the `FnPtr` visitor matches, do we? The `IA2_fnptr__<mangled type>` struct typedefs inserted by that visitor don't themselves require the header, and if we want to import `ia2.h`  in *every* header we process, then neither visitor we have is the right place to do so.

ayrtonm:
> The first is that there's the confusing terminology of a visitor running: both visitors unconditionally have their matchers applied to the ASTMatchRefactorer, so they do "run" over the AST of all files. But their run method does indeed only get executed when their matchers match.

Yeah by "run" I always meant the latter (when the matchers match part of the AST).

> The second is that the FnPtrPrinter visitor does not match all headers. ... Prior to this change, we assumed that the FnDecl visitor would trigger on a strict superset of the files that the FnPtrPrinter visitor would, but this isn't actually true

Yep, I'm aware FnPtr won't match all headers. I think we previously just hadn't considered in detail what happens when FnDecl doesn't run.

> That said, I don't think we actually want to include ia2.h in exactly the set of headers where the FnPtr visitor matches, do we? The IA2_fnptr__<mangled type> struct typedefs inserted by that visitor don't themselves require the header, and if we want to import ia2.h in every header we process, then neither visitor we have is the right place to do so.

Right, the mangled type structs only need the output header, but using them requires ia2.h so I think we do want to include it in the headers that FnPtr visitor matches. FnDecl doesn't modify the headers it matches so I don't see a reason to include ia2.h in those.

fw-immunant:
Sounds good. Let's clarify the situation in the commit message body and then merge this.

PR #159, opened 2022-08-25
Move common wrappers to their own cmake file 

rinon:
Moves the ffmpeg wrapper definitions out of the ffmpeg test and into a
common area for use by other binaries in tree.


cmake/define-ia2-wrapper.cmake
```diff
@@ -163,9 +163,10 @@ function(define_ia2_wrapper)
           ${REWRITTEN_HEADERS}
           --
           -fgnuc-version=6
-          -I ${REWRITTEN_HEADER_DIR}
```
ayrtonm, line 166: This seems to be causing the error with overlapping replacements.

PR #161, opened 2022-08-26
Define _GNU_SOURCE for tests

ayrtonm:
ia2.h requires defining _GNU_SOURCE since it uses the definition of
`dl_phdr_info`. Using `#define _GNU_SOURCE` is unreliable since link.h defines
dl_phdr_info and may be included by another header before ia2.h so we have to
use `-D_GNU_SOURCE`.

### DIFF
--- a/cmake/define-ia2-wrapper.cmake
+++ b/cmake/define-ia2-wrapper.cmake
@@ -179,6 +179,7 @@ function(define_ia2_wrapper)
     if(LIBIA2_INSECURE)
         target_compile_definitions(${WRAPPER_TARGET} PUBLIC LIBIA2_INSECURE=1)
     endif()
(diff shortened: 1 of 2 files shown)


cmake/define-ia2-wrapper.cmake
```diff
@@ -166,6 +166,7 @@ function(define_ia2_wrapper)
           -D_GNU_SOURCE
           -isystem ${C_SYSTEM_INCLUDE}
           -isystem ${C_SYSTEM_INCLUDE_FIXED}
+          -I ${REWRITTEN_HEADER_DIR}
```
rinon, line 169: We can't do this because libavutil includes `time.h` which must not shadow the system header of the same name.

ayrtonm: So this should be the root build dir?

Issue #160, opened 2022-08-26
Deduplicate code in libia2 headers


ayrtonm:
We've been defining some functions and variables in ia2.h/scrub_registers.h out of convenience and since we weren't exactly sure how the runtime needed to be initialized. Now that we have a better idea of how compartment initialization should work, we should move some functions from ia2.h into a proper source file to avoid issues with duplicate definitions. We should try to only define things once where possible, but in some cases dynamic linking may make this difficult. In those cases if it is significantly easier to have one definition per shared object we should make the names unique if possible (e.g. append the compartment number) to make debugging easier.

It might also make sense to move some of what remains in ia2.h to a "private" header. A lot of the small helper macros are used in the function pointer macros so they'll still need to be accessible by including ia2.h but it'll make it more clear what's in the public API.

Required for #155.

ayrtonm:
> We should try to only define things once where possible, but in some cases dynamic linking may make this difficult.

@rinon I think for things like the register scrubbing asm it might be hard to have one definition since you need the right PKRU to call something through the PLT. Things that run before main (e.g. stack allocation stuff) may be easier to deduplicate though.

ayrtonm:
#163 addressed the main issue with defining things in a header which is what was blocking #155. Let's keep this issue open since there are still a few other opportunities to deduplicate code in the libia2 headers. This is lower priority though.

PR #163, opened 2022-08-27
Split ia2.h into its own shared library

rinon:
ia2.h had code that didn't need to be duplicated in every compilation
unit, which was the case since functions were all declare static inline
in the header. This change starts to split that apart by moving
protect_pages into a shared library.

Addresses #160


rinon:
I haven't gone through and determined if there's more code we can split out into the library, but I wanted to put up a minimal change that at least gets the build system working.

include/ia2.h
```diff
@@ -26,6 +26,11 @@
 #define IA2_WRPKRU "wrpkru"
 #endif
 
+// The callback passed to dl_iterate_phdr in the constructor inserted by
+// INIT_COMPARTMENT to pkey_mprotect the pages corresponding to the
+// compartment's loaded segments.
+int protect_pages(struct dl_phdr_info *info, size_t size, void *data);
```
ayrtonm, line 32: Maybe we should add an `#ifndef _GNU_SOURCE/#error "This header requires using -D_GNU_SOURCE"` since we need link.h to declare `dl_phdr_info` and it's not always clear what the issue is when it's not defined.

libia2/include/pkey_init.h
```diff
@@ -0,0 +1,30 @@
+#ifndef PKEY_INIT_H
```
ayrtonm, line 1: Previously we generated this file to call the rust libia2, but we don't need this file (ia2.h already has the things it defines/declares).

include/ia2.h
```diff
@@ -26,6 +26,11 @@
 #define IA2_WRPKRU "wrpkru"
```
ayrtonm, line 26: Shouldn't we remove this file if we also have libia2/include/ia2.h?

ayrtonm:
LGTM, just a few minor issues.

include/ia2.h
rinon: Good idea! Done.

Issue #134, opened 2022-05-02
Add an allocator with MPK support


endbr64:
Our current plan is to leave `ngx_alloc()` and `ngx_calloc()` unwrapped to read the PKRU of the caller and allocate on the corresponding heap. We also still need to choose an allocator.

ayrtonm:
I switched our approach for replacing the allocator (#145) from symbol overriding to ld --wrap because it gives us more precise control over which shared libraries will use partition alloc and which use the default malloc implementation.

PR #164 adds support for protecting heap memory, but it only handles memory reserved for allocations and does not protect any [allocator metadata](https://chromium.googlesource.com/chromium/src/+/master/base/allocator/partition_allocator/PartitionAlloc.md#architecture). Looking into what metadata needs to be protected is probably higher priority than the missing items in the previous comment.

ayrtonm:
I fixed the issue [Stephen pointed out](https://github.com/immunant/IA2-Phase2/pull/164#discussion_r967559165) with shared allocations in #164. Ideally we'd use partition alloc for both shared and private allocations, but that requires a decent number of changes to PA and it's less important than the allocator metadata issue (#165). PRs #145 and #164 are ready to merge as soon as I add a few more tests and throw together an update script (or just copy my .bash_history) for updating the chromium sources.

fw-immunant:
Unassigned as I'm not working on this at present, and I think @endbr64 is closer as that's who did most of the PartitionAlloc work.

rinon:
Nginx with partition-alloc is working and landed

PR #164, opened 2022-08-29
Tag heap memory with protection keys

ayrtonm:
This works as expected, but I still need to write a few more tests and fix up their reference output.


external/chromium/src/base/allocator/partition_allocator/partition_bucket.cc
```diff
@@ -167,6 +170,14 @@ uintptr_t ReserveMemoryFromGigaCage(pool_handle pool,
 #endif
 
   PA_DCHECK(!(reserved_address % kSuperPageSize));
+
+  size_t pkey = ::ia2_get_pkey();
```
rinon, line 174: Is this what we want? Partition 0 should be shared, but what happens if the allocator needs to reserve more pages for this shared partition while running in a compartment? Don't we need to track the partition compartment as a field of that partition's metadata or something?

ayrtonm: You're right, this doesn't actually ever reserve memory for the shared partition. I don't think we need to store the partition's compartment in the metadata, just need make sure that the pkey used for allocations that trigger calls to this function gets propagated down to it.

ayrtonm: Actually here we have two options (1) use partition allocator for private and shared allocations, (2) use partition alloc for private allocations and glibc malloc for shared allocations. I originally intended to implement the former, but the latter is much easier to implement. I think the first option is ideal since it uses a single allocator, but I'm going to go with the latter for now to have something that we can use with nginx.

ayrtonm:
Ready to merge after adding a few more tests. Closes #134.

ayrtonm:
Changes merged as part of #176

PR #169, opened 2022-10-25
Refactor and clean up `protect_pages()` and fix alignment requirements

rinon:
Clean up the protect_pages callback for `dl_iterate_phdr`.

Also relaxes the requirement that `ia2_shared_*` must exist, which lets us fix the linker script to eliminate these sections if they aren't required.

### DIFF
--- a/libia2/ia2.c
+++ b/libia2/ia2.c
@@ -71,28 +71,46 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
   if (!lib)
     exit(-1);
 
-  struct AddressRange shared_ranges[NUM_SHARED_RANGES];
+  struct AddressRange shared_ranges[NUM_SHARED_RANGES] = {0};
+  struct AddressRange *cur_range = &shared_ranges[0];
   for (size_t i = 0; i < NUM_SHARED_RANGES; i++) {
     // Clear any potential old error conditions
     dlerror();
-    shared_ranges[i].start = (uint64_t)dlsym(lib, shared_sections[i][0]);
-    uint64_t aligned_start = shared_ranges[i].start & ~0xFFFUL;
-    if (aligned_start != shared_ranges[i].start) {
+
+    cur_range->start = (uint64_t)dlsym(lib, shared_sections[i][0]);
+    if (!cur_range->start) {
+      // We didn't find the start symbol for this shared section. Either the
+      // user didn't mark any shared global data or we didn't link with the
+      // correct linker script. We can't distinguish these cases here and the
+      // first shouldn't be an error, so let's continue.
+      continue;
+    }
+    uint64_t aligned_start = cur_range->start & ~0xFFFUL;
+    if (aligned_start != cur_range->start) {
       printf("Start of section %s is not page-aligned\n",
              shared_sections[i][0]);
       exit(-1);
     }
-    if (dlerror())
-      exit(-1);
 
-    shared_ranges[i].end = (uint64_t)dlsym(lib, shared_sections[i][1]);
-    uint64_t aligned_end = (shared_ranges[i].end + 0xFFFUL) & ~0xFFFUL;
-    if (aligned_end != shared_ranges[i].end) {
+    cur_range->end = (uint64_t)dlsym(lib, shared_sections[i][1]);
+    char *dl_err = dlerror();
+    if (dl_err) {
+      printf("Could not find end symbol of shared section %s: %s\n",
+             shared_sections[i][1], dl_err);
+      exit(-1);
+    }
+    if (!cur_range->end) {
+      printf("End symbol of shared section %s was unexpectedly NULL\n",
+             shared_sections[i][1]);
+      exit(-1);
+    }
+    uint64_t aligned_end = (cur_range->end + 0xFFFUL) & ~0xFFFUL;
+    if (aligned_end != cur_range->end) {
       printf("End of section %s is not page-aligned\n", shared_sections[i][1]);
       exit(-1);
     }
-    if (dlerror())
-      exit(-1);
+
+    cur_range++;
   }
 
   // TODO: We avoid dynamically allocating space for each of the `dlpi_phnum`
--- a/libia2/ia2.c
+++ b/libia2/ia2.c
@@ -1,15 +1,35 @@
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#include <link.h>
+
 #include "ia2.h"
 
-int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
-  // Use the address in `data` (PhdrSearchArgs) to determine which program
-  // headers belong to the compartment being initialized.
-  if (!data || !info) {
-    printf("Passed invalid args to dl_iterate_phdr callback\n");
-    exit(-1);
-  }
-  struct PhdrSearchArgs *search_args = (struct PhdrSearchArgs *)data;
-  Elf64_Addr address = (Elf64_Addr)search_args->address;
-  bool this_compartment = false;
+static const char *shared_sections[][2] = {
+    {"__start_ia2_shared_data", "__stop_ia2_shared_data"},
+    {"__start_ia2_shared_rodata", "__stop_ia2_shared_rodata"},
+};
+
+// The number of ELF sections that may be shared by protect_pages
+#define NUM_SHARED_RANGES (sizeof(shared_sections) / sizeof(shared_sections[0]))
+
+// The number of program headers to allocate space for in protect_pages. This is
+// only an estimate of the maximum value of dlpi_phnum below.
+#define MAX_NUM_PHDRS 20
+
+struct AddressRange {
+  uint64_t start;
+  uint64_t end;
+};
+
+struct SegmentInfo {
+  uint64_t addr;
+  size_t size;
+  int prot;
+};
+
+/// Check if \p address is in a LOAD segment in \p info
+static bool in_loaded_segment(struct dl_phdr_info *info, Elf64_Addr address) {
   for (size_t i = 0; i < info->dlpi_phnum; i++) {
     if (!info->dlpi_phdr) {
       exit(-1);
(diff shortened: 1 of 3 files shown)


endbr64 (approved):

libia2/ia2.c
```diff
@@ -71,28 +71,37 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
   for (size_t i = 0; i < NUM_SHARED_RANGES; i++) {
     // Clear any potential old error conditions
     dlerror();
-    shared_ranges[i].start = (uint64_t)dlsym(lib, shared_sections[i][0]);
-    uint64_t aligned_start = shared_ranges[i].start & ~0xFFFUL;
-    if (aligned_start != shared_ranges[i].start) {
+
+    cur_range->start = (uint64_t)dlsym(lib, shared_sections[i][0]);
+    if (!cur_range->start) {
+      // We didn't find that symbol, oh well.
```
fw-immunant, line 82: I think it would be better to note that we didn't find the symbol because the section didn't exist--sections not existing is fine, but if the section did exist, not being able to find its start symbol would be a real problem.

libia2/padding.ld
```diff
@@ -10,21 +10,19 @@
...
 } INSERT AFTER .rodata;
 
 SECTIONS {
-    ia2_shared_data : {
-        . = ALIGN(4096);
+    ia2_shared_data ALIGN(4096): {
         __start_ia2_shared_data = .;
         *(ia2_shared_data)
-        . = ALIGN(4096);
+        . = ALIGN(. != 0 ? 4096 : 1);
```
fw-immunant, line 25: Same here.

fw-immunant (changes requested):
Comments inline. Looks good otherwise once comments are addressed.

libia2/ia2.c
```diff
@@ -21,65 +41,81 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
     dlerror();
-    ignored_ranges[i].start = (uint64_t)dlsym(lib, ignored_sections[i][0]);
-    uint64_t aligned_start = ignored_ranges[i].start & ~0xFFFUL;
-    if (aligned_start != ignored_ranges[i].start) {
-      printf("Start of section %s is not page-aligned\n",
-             ignored_sections[i][0]);
-      exit(-1);
+
+    cur_range->start = (uint64_t)dlsym(lib, shared_sections[i][0]);
+    if (!cur_range->start) {
```
ayrtonm, line 81: We should check `dlerror` rather than comparing with `NULL`. Though I'm not sure how relevant getting `NULL` from `dlsym` is for start symbols of sections in `PT_LOAD` segments.

libia2/ia2.c
```diff
@@ -21,65 +41,81 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
   }
-  if (!this_compartment) {
+  return false;
+}
+
+/// Map ELF segment flags to mprotect access flags
+static int segment_flags_to_access_flags(Elf64_Word flags) {
+  return ((flags & PF_X) != 0 ? PROT_EXEC : 0) |
+         ((flags & PF_W) != 0 ? PROT_WRITE : 0) |
+         ((flags & PF_R) != 0 ? PROT_READ : 0);
```
ayrtonm, line 54: You dropped `| PROT_NONE`

libia2/padding.ld
```diff
@@ -10,21 +10,19 @@
    __start_$SECTION and before __stop_$SECTION.
 */
 SECTIONS {
-    ia2_shared_rodata : {
-        . = ALIGN(4096);
+    ia2_shared_rodata ALIGN(4096): {
         __start_ia2_shared_rodata = .;
         *(ia2_shared_rodata)
-        . = ALIGN(4096);
+        . = ALIGN(. != 0 ? 4096 : 1);
```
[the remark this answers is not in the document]
ayrtonm: I think it's just to ensure the start and stop symbols are at the same address so we don't load a page unnecessarily if ia2_shared_rodata is empty.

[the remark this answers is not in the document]
ayrtonm: I think lld doesn't support `A ? B : C` in some contexts, but the change itself seems fine to me.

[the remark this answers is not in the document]
rinon: I took this from the ld default linker script. I'll try without it, but I think we will end up with a page if we don't do this.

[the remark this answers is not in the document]
rinon: Yes, I tested this, it's required to avoid emitting an empty section. We don't need this section to exist if its size is 0, and this is what is required to do that...

libia2/ia2.c
rinon: While I agree in general, we don't want to accept NULL values for either of these symbols, right?

libia2/ia2.c
rinon: PROT_NONE is 0, we don't need to OR that in here.

libia2/ia2.c
rinon: done

libia2/padding.ld
```diff
@@ -10,21 +10,24 @@
...
-    ia2_shared_rodata : {
-        . = ALIGN(4096);
+    ia2_shared_rodata ALIGN(4096): {
         __start_ia2_shared_rodata = .;
         *(ia2_shared_rodata)
-        . = ALIGN(4096);
+        /*
+           This weird alignment expression is required so that the section will
+           get elided if empty. Apparently ld won't drop a section that has a
+           non-zero aligned end, even if its size ends up being zero?
```
fw-immunant, line 19: My point from earlier is that `. = ALIGN(4096);` will not change the value of `.` (and presumably therefore not change the end of the section) if `.` is zero. I don't want to be a huge pain in the ass here, but I really don't understand what effect this `ALIGN` expression is having here and why.

My test:
test.ld:
```
SECTIONS
{
  . = 4096;
  "1_test_assign_4096" = .;
  . = ALIGN(4096);
  "2_test_align_4096_on_4096" = .;
  . = 0;
  "3_test_assign_0" = .;
  . = ALIGN(4096);
  "4_test_align_4096_on_0" = .;
  . = 1;
  "5_test_assign_1" = .;
  . = ALIGN(4096);
  "6_test_align_4096_on_1" = .;
  . = 0;
  . = ALIGN(1);
  "7_test_align_1_on_0" = .;
  . = 0;
  . = ALIGN(. != 0 ? 4096 : 1);
  "8_test_ternary_on_0" = .;
  . = 1;
  . = ALIGN(. != 0 ? 4096 : 1);
  "9_test_ternary_on_1" = .;

  . = 1-000;
  .text : { *(.text) }
  . = 0x8000000;
  .data : { *(.data) }
  .bss : { *(.bss) }
}
```
trivial.c:
```
int main(int argc, char** argv) {
	return 0;
}
```

harness: `gcc trivial.c -T test.ld -o test && nm test | fgrep test`
output:
```
0000000000001000 T 1_test_assign_4096
0000000000001000 T 2_test_align_4096_on_4096
0000000000000000 T 3_test_assign_0
0000000000000000 T 4_test_align_4096_on_0
0000000000000001 T 5_test_assign_1
0000000000001000 T 6_test_align_4096_on_1
0000000000000000 T 7_test_align_1_on_0
0000000000000000 T 8_test_ternary_on_0
0000000000001000 T 9_test_ternary_on_1
```

So if the effect on `.` of both `. = ALIGN(4096)` and `. = ALIGN(. != 0 ? 4096 : 1)` is identical, then what's going on here?

rinon: The only way I got this to work was the combination of the first ALIGN on the section (not on . like your test) PLUS the weird ternary. I don't have a way to test handy but I can reduce this later.

libia2/ia2.c
ayrtonm: I think you're right that we don't want to accept `NULL` values, but I remember @endbr64 ran into issues with this. I think we were modifying the ld script and it made one of the symbols `NULL` so splitting this into two separate checks would be good to easily distinguish between the symbol not existing and the symbol having an invalid value.

ayrtonm: fwiw I'd like to avoid modify ld scripts in weird ways to avoid the root issue in that case, but splitting this into two checks would make debugging easier.

rinon: Done. I left the first one as a null check because we want to silently continue in either case, though.

libia2/padding.ld
rinon: I tested the following

test.ld:

```SECTIONS
{
  . = 4096;
  "1_test_assign_4096" = .;
  . = ALIGN(4096);
  "2_test_align_4096_on_4096" = .;
  . = 0;
  "3_test_assign_0" = .;
  . = ALIGN(4096);
  "4_test_align_4096_on_0" = .;
  . = 1;
  "5_test_assign_1" = .;
  . = ALIGN(4096);
  "6_test_align_4096_on_1" = .;
  . = 0;
  . = ALIGN(1);
  "7_test_align_1_on_0" = .;
  . = 0;
  . = ALIGN(. != 0 ? 4096 : 1);
  "8_test_ternary_on_0" = .;
  . = 1;
  . = ALIGN(. != 0 ? 4096 : 1);
  "9_test_ternary_on_1" = .;
  .test10 ALIGN(4096): {
    . = 0;
    . = ALIGN(4096);
    "10_test_section_align_4096" = .;
  }
  test11 ALIGN(4096): {
    . = 0;
    . = ALIGN(. != 0 ? 4096 : 1);
    "11_test_section_ternary_on_0" = .;
  }
}
```

Harness: `gcc trivial.c -Wl,-T test.ld -o test && readelf -S test | grep test`

What we're doing is avoiding emission of the _section_, not the symbols.

fw-immunant: That does convince me that somehow, *despite* calculating the same value, the ternary does somehow allow the section to be omitted while the regular ALIGN doesn't. Neither replacing the ternary with `ALIGN(1)` nor with `ALIGN(4096)` lets the `.test11` section be omitted.

Some cases I tested just now:
- `. = ALIGN(. != 0 ? 4096 : 1);` allows section omission
- `. = ALIGN(. == 0 ? 1 : 4096);` does not
- `. = ALIGN(0 != . ? 4096 : 1);` does not
- `. = ALIGN(. ? 4096 : 1);` does not

Something Very Strange is happening here, and I don't like it. It's not just use of ternary, but some magic property of using that exact syntax within an `ALIGN`. It must be a feature since this is used in the default linker script, but it's not one I've ever heard of.

That said, you've demonstrated the necessity of this particular syntax and I'm fine with merging it as long as we have a comment that sufficiently exposes this arcanum.

fw-immunant: Out of curiosity, I tested with lld.

lld wants a `  . = 0;` before `.test10`, but otherwise does accept this linker script... however, it does not seem to obey the same undocumented, syntax-specific rules for omitting the section. The only way I can get it to omit the `.test11` section is by removing all the assignments in it.

rinon: Spoooky. Well, I think we should still keep this for BFD, and we can investigate further.

fw-immunant:
```suggestion
           This weird alignment expression is required so that the section will
           get elided if empty. ld.bfd special-cases this syntax to mean this,
           which is not documented but appears in its default linker script; see
           binutils commit 9d12f64cdc6540b75938097d7f4ab460bb346528
```

fw-immunant (approved):
Approved as long as you do something along the lines of the comment suggestion I just made re: the `ALIGN` ternary.

libia2/padding.ld
rinon: Good sleuthing, I'll make that change when I get home, thanks.

PR #170, opened 2022-10-25
Share all read-only segments and relro section

rinon:
We don't need to keep read-only segments mapped in from disk secret, as
we assume binaries on disk are not secret. There is various metadata in
these segments that needs to be shared, so its simpler to just share all
of it. The same applies to the relro section of the RW segment -
relocations aren't secret, and relocations are the only modifications to
the relro sections that are made before the dynamic linker marks the
page(s) as read-only.


libia2/ia2.c
```diff
@@ -123,49 +143,44 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
-                 (shared_ranges[j].start >= start)) {
-        size_t len = stop - shared_ranges[j].end;
-        segment_info[extra_seg_info].addr = shared_ranges[j].end;
-        segment_info[extra_seg_info].size = len;
-        segment_info[extra_seg_info].prot = prot;
-
-        stop = shared_ranges[j].start;
-        extra_seg_info += 1;
+    Elf64_Addr seg_end = (start + phdr.p_memsz + 0xFFFUL) & ~0xFFFUL;
+    while (start < seg_end) {
```
endbr64, line 160: This is great, I like that it handles a segment being split into an arbitrary number of pieces.

endbr64 (approved):

libia2/ia2.c
rinon: The real question... is it correct? I _think_ so? But it's tricky. Looks ok to you?

libia2/ia2.c
```diff
@@ -123,49 +143,44 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
+      Elf64_Addr cur_end = seg_end;
+
+      // Adjust start and cur_end to bound the lowest region between start and
+      // seg_end that is not covered by a shared range
+      for (size_t j = 0; j < shared_range_count; j++) {
+        if (shared_ranges[j].start <= start && shared_ranges[j].end > start) {
+          start = shared_ranges[j].end;
+        } else if (shared_ranges[j].start >= start &&
+            shared_ranges[j].start <= cur_end) {
```
fw-immunant, line 169: I believe this should be comparing against `seg_end`, not `cur_end`. The comment also suggests as much.

fw-immunant: I also think that a `<` comparison is more appropriate than `<=` here, as the guarded assignment is a NOP in the equal case.

libia2/ia2.c
```diff
@@ -123,49 +143,44 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
+    Elf64_Addr seg_end = (start + phdr.p_memsz + 0xFFFUL) & ~0xFFFUL;
+    while (start < seg_end) {
+      Elf64_Addr cur_end = seg_end;
+
+      // Adjust start and cur_end to bound the lowest region between start and
+      // seg_end that is not covered by a shared range
+      for (size_t j = 0; j < shared_range_count; j++) {
+        if (shared_ranges[j].start <= start && shared_ranges[j].end > start) {
+          start = shared_ranges[j].end;
+        } else if (shared_ranges[j].start >= start &&
```
fw-immunant, line 168: Suppose `shared_ranges[j].start` is equal to `start` (to see what happens in the `=` case of this `>=`).

Then, either the condition on L167 holds (and we don't reach the `else if`), or `shared_ranges[j].end` is not greater than `start` (i.e., this is a zero-length range). The body of this `if` is a no-op in the zero-length range case, and it's clearer to not match it in the condition (which is looking for an overlapping, rather than touching, range to skip) than match it and perform a no-op. So I would advocate making this `>=` simply a `>`.

libia2/ia2.c
```diff
@@ -123,49 +143,44 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
-        stop = shared_ranges[j].start;
-        extra_seg_info += 1;
+    Elf64_Addr seg_end = (start + phdr.p_memsz + 0xFFFUL) & ~0xFFFUL;
+    while (start < seg_end) {
+      Elf64_Addr cur_end = seg_end;
+
+      // Adjust start and cur_end to bound the lowest region between start and
+      // seg_end that is not covered by a shared range
+      for (size_t j = 0; j < shared_range_count; j++) {
+        if (shared_ranges[j].start <= start && shared_ranges[j].end > start) {
```
fw-immunant, line 166: Comment this condition: ``// Look for a shared range that overlaps `start`
``

libia2/ia2.c
fw-immunant: Comment this condition: ``// Look for a shared range that starts within [`start`, `cur_end`)``

fw-immunant (changes requested):
Comments inline. Only correctness concern (blocking) is `cur_end` vs. `seg_end`.

libia2/ia2.c
```diff
@@ -123,49 +142,44 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
...
-        size_t len = stop - shared_ranges[j].end;
-        segment_info[extra_seg_info].addr = shared_ranges[j].end;
-        segment_info[extra_seg_info].size = len;
-        segment_info[extra_seg_info].prot = prot;
-
-        stop = shared_ranges[j].start;
-        extra_seg_info += 1;
+    Elf64_Addr seg_end = (start + phdr.p_memsz + 0xFFFUL) & ~0xFFFUL;
+    while (start < seg_end) {
+      Elf64_Addr cur_end = seg_end;
```
ayrtonm, line 161: Why are we resetting `cur_end` in each `while` loop iteration if we're trying to bound it?

ayrtonm: Ah nvm, I see what we're doing and this makes sense.

libia2/ia2.c
ayrtonm: I think `cur_end` here is correct since we may decrease it in the `for` loop and `seg_end` is fixed. The comment also makes sense since `cur_end` is reset to `seg_end` at the start of each `while` loop iteration, though I agree it's not super clear.

ayrtonm (approved):
The changes look correct to me.

libia2/ia2.c
rinon: cur_end is monotonically decreasing in each loop iteration, so the comparison is needed. LT instead of LTE makes sense, will fix. I'll add more comments.

libia2/ia2.c
rinon: Which >= are you referring to here? I'm making `shared_ranges[j].start < cur_end` in second half of the second conditional.

rinon: The `else` is unnecessary, I just put it for clarity because if the first conditional triggers the second cannot.

rinon: Ahhh I see what you mean. Hopefully I've clarified sufficiently now?

fw-immunant (approved):

libia2/ia2.c
fw-immunant: Yes, looks clear and correct now. :+1:

docs/usage.md
```diff
@@ -114,9 +114,17 @@ gcc main.c libbar.so libbar_shim.so libmain_shim.so -Wl,-z,now \
     -I $IA2_INCLUDE_DIR -fPIC -Wl,-rpath=. -Wl,-T/$REPO_ROOT/libia2/padding.ld
 ```
 
-Shared objects that are assigned one a trusted protection key must have certain
+All read-only and relro sections of shared objects will be shared with all
```
ayrtonm, line 117: super minor nit: s/shared objects/DSOs/ since "shared" means two different things in this sentence.

rinon: Done. I wanted to avoid DSO here because that gets confusing when the main binary is PIE, but hopefully its better now.

libia2/ia2.c
endbr64: I ran it by hand on a couple of examples on paper. While not as good as a unit test, I'm convinced.

rinon: Heh that's exactly what I did too. I feel sufficiently bad about not having a unit test here but I didn't want to get derailed adding a unit test framework...

PR #173, opened 2022-10-31
Inline compartment initialization

### LINKED ISSUE #172: Initial page protection is indirect and can be replayed
Initial page protection of writable segments is done by passing a callback to dl_iterate_phdr. This callback takes indirect params which control the target object and pkey. An attacker could call this at any point in the program and change the protection mapping of an arbitrary object to an arbitrary pkey. We should either emit a separate callback for each compartment that can only set its own object to the correct pkey, or enforce a global lock on changing pkeys after initialization.

rinon:
Compartment initialization is a sensitive operation and it must not be
possible to modify inputs to this operation and replay it with different
compartment IDs or ranges. This change inlines initialization statically
into the compartment so that we don't have to accept arbitrary addresses
and key IDs. As a nice side-effect, this means we no longer have to rely
on dlopen/dlsym to find the shared data section symbols.

Fixes #172

### DIFF
--- a/libia2/include/ia2.h
+++ b/libia2/include/ia2.h
@@ -20,35 +20,10 @@
 #define PASTE4(w, x, y, z) PASTE4_(w, x, y, z)
 
 #ifdef LIBIA2_INSECURE
-#define INIT_COMPARTMENT(n)
 #define IA2_WRPKRU
 #else
-#define INIT_COMPARTMENT(n) _INIT_COMPARTMENT(n)
 #define IA2_WRPKRU "wrpkru"
 #endif
-
-/// Protect pages in the given shared object
-///
-/// \param info dynamic linker information for the current object
-/// \param size size of \p info in bytes
-/// \param data pointer to a PhdrSearchArgs structure
-///
-/// The callback passed to dl_iterate_phdr in the constructor inserted by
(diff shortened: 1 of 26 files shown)


libia2/include/ia2.h
```diff
@@ -260,3 +219,10 @@ asm(".macro mov_mixed_pkru_eax pkey0, pkey1\n"
       ia2_stackptrs[i] = stack_start + STACK_SIZE - 8;                         \
     }                                                                          \
   }
+
+
+#ifdef IA2_INIT_COMPARTMENT
+#ifndef LIBIA2_INSECURE
+#include "../ia2.c"
```
ayrtonm, line 226: Still haven't gone through the PR in detail, but the point of moving stuff from ia2.h into a separate library was to avoid including everything it defined multiple times. This may be less of an issue in this case because of the `#ifdef IA2_INIT_COMPARTMENT` but I think we should try to avoid this if possible.

rinon: I think we need to so we're not taking in a pointer that control the pkey or address. If you have another way to avoid that, I'm definitely open to it.

rinon:
We don't need this with the new page probing design.

Issue #158, opened 2022-08-24
Remove header rewriter unnecessary options


ayrtonm:
tl;dr: #104 added flags to the rewriter to handle cases where headers are shared between compartments. #149 made `--omit-wrapper` unnecessary so we could remove it. `--shared-headers` also became unnecessary after the changes in #114.

[This list](https://github.com/immunant/IA2-Phase2/issues/104#issuecomment-1084876277) is outdated so here's an updated list of what the rewriter should do.

1. Generate a linker args file from function declarations
2. Generate call gate wrappers for direct calls from function declarations and add them to the shim source
3. Rewrite function pointer types as opaque structs
4. Generate call gate wrapper macros from function pointer types and add them to the output header
5. `#include` ia2.h and the output header (if any) in headers with our opaque structs

The rewriter does all five by default (as of #157). In some cases, functions in a header don't all belong to one compartment so `--function-allowlist=file` was added to allow skipping fn declarations selectively. In other cases, headers are only needed for the opaque pointer types so `--omit-wrappers` allowed skipping all fn declarations.

As of #149 `--omit-wrappers` isn't strictly necessary since function declarations only generate the files in points 1 and 2 and never trigger any changes in the headers (i.e. we don't add `.symver` to the headers anymore). It would be simpler to only keep the `--function-allowlist=file` option for when a subset of the function declarations should appear in the shim. For cases where all compartment transitions are through function pointers we can remove the `--omit-wrappers` option and either

1. just ignore the shim source and linker args that we generate or
2. make the `wrapperOutputFilename` argument optional and use that to determine whether to skip all function declarations or not

As for `--shared-headers`, I can't recall why we didn't get rid of it in #114, but it was unnecessary even then so we should remove it.

Issue #131, opened 2022-05-02
Remove type annotations in `IA2_DEFINE_WRAPPER`


ayrtonm:
In #121 the following frequently occurs
```
// foo.h
int foo(int x);
typedef int (*fn_ptr_t)(int);

// foo.c
fn_ptr_t global_fn_ptr = foo;
```

We have to rewrite the function pointer as follows (ignoring the `target_pkey` macro parameters)
```
// foo.c
IA2_DEFINE_WRAPPER(foo, _ZTSPFiiE);
fn_ptr_t global_fn_ptr = IA2_WRAPPER(foo);
```
Looking up mangled type names is time-consuming and `foo` appears in the headers so we should be able to remove it by using the header-rewriter. The idea is to generate something like
```
#define IA2_foo_mangled_type _ZTSPFiiE
```
in the output header to map a function's name to the mangled type of its function signature, removing the need to specify `_ZTSPFiiE`.

As mentioned in #130, about 60% of the type annotations added so far are for this macro (and `IA2_DECLARE_WRAPPER`). The idea above won't allow removing all type annotations since sometimes the function pointer is initialized to a function that's not declared in a header, but I think requiring some type annotations is acceptable. I don't have a good sense of how frequently each case occurs, but I'd guess it's about 50/50 so this might let us remove ~30% of the type annotations.

Issue #130, opened 2022-05-02
Remove type annotations in `IA2_CALL`


ayrtonm:
To ensure indirect calls go through compartment transitions, we manually modify callsites in the source code by adding the `IA2_CALL` macro. The macro expands to an asm statement defining the call gate wrapper for the callsite which must be specific to the function signature. We make `IA2_CALL` signature-specific by passing it the function pointer type's mangled name (e.g. `IA2_CALL(my_function, _ZTSPFvvE, caller_pkey)`). Finding the correct mangled type names is time-consuming and they hurt readability. In #121, 122 out of the 295 type annotations added so far use `IA2_CALL`. The rest are mostly `IA2_DEFINE_WRAPPER` and only about half of those can be removed in a reasonable way so we should start with just `IA2_CALL`.

To remove these type annotations we can use [generic selection](https://en.cppreference.com/w/c/language/generic) and fallback to our current approach if it's not available. For example, let's assume the headers for a compartment have two function pointer types `int(*)(int)` and `void (*)(void)`. `IA2_CALL` currently defines one wrapper at each callsite, but since the target functions are not fixed at build-time we really only need one wrapper for each function pointer type in the headers.

`IA2_CALL` could then look something like
```
/// Target is the opaque function pointer expression. Let's ignore caller_pkey for simplicity.
/// Variadic arguments are new and will be the target function's arguments
#define IA2_CALL(target, caller_pkey, ...)
    ({
    // Define all wrappers using existing macros.
    // Since we're doing this at every callsite we should put the wrappers in a new executable
    // section and use .linkonce to ensure there's only one copy
    asm(IA2_CALL__ZTSPFiiE);
    asm(IA2_CALL__ZTSPFvvE)
    // Declare all wrappers as regular function pointers
    int(*__ia2_ZTSPFiiE_wrapper)(int);
    void(*__ia2_ZTSPFvvE_wrapper)(void);
    // All wrappers will load the target function pointer from this variable.
    // This should be thread-local, but that's an issue with the current implementation anyway.
    extern void *target_ptr;
    target_ptr = target;
    // Call one of the wrappers based on the target pointer expression's type
    _Generic(target,
        struct IA2_fnptr__ZTSPFiiE:  __ia2_ZTSPFiiE_wrapper,
        struct IA2_fnptr__ZTSPFvvE: __ia2_ZTSPFvvE_wrapper)(__VA_ARGS__)
    })
```

[...]

Issue #51, opened 2021-11-16
Pointers to function typedefs are not rewritten at all


ayrtonm:
In this example
```
typedef int(callback_t)(int);
int apply(callback_t *cb, int x) {
    return cb(x);
}
```
the function pointer `callback_t *cb` isn't being wrapped like `typedef int(*fnptr_t)(int)` would be.

PR #174, opened 2022-11-23
docs: Add source rewriter design doc

ayrtonm:
For reference, the latest function pointer macros are defined in ia2.h [in this PR](https://github.com/immunant/IA2-Phase2/pull/121/files#diff-f732e46bfbfd1583f6055e79a50dc0c290bd90c0a89ada1a2b4101f5bcc198c5).

This makes #158, #131, #130, ~~#51,~~ #49, #43, ~~#24, #22,~~ #21 and #4 obsolete.

edit: #51, #24 and #22 are still relevant


docs/source_rewriter.md
```diff
@@ -0,0 +1,195 @@
...
+super low priority.
+
+### Is a compile commands JSON required?
+
+Up for debate so it's TBD
+
+### What if I want to make the source file changes myself? 
+
+We can add a flag to only generate the output `.c` with the call gate wrappers
+and leave the input `.c`s unmodified.
```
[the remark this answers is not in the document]
ayrtonm: Since we're going with replacing fn ptrs again, this is an option but it's low priority.

ayrtonm:
I omitted talking about rewriting function pointer types in the last commit to focus on the new things the tool will rewrite and make it easy to review. We will rewrite function pointer types though.

docs/source_rewriter.md
```diff
@@ -0,0 +1,210 @@
...
+The rewriter will take all source files as input along with which compartment
+each belongs to. It will generate one output `.c` and one `.h` for use with all
+compartments. It will also generate compartment-specific linker args and
+`objcopy --redefine-syms` files.
+
+### Define call gates for direct transitions
+
+The rewriter will work in two phases. In the first phase it'll parse all source
+files and keep track of functions defined and their compartment, functions
+called and from which compartment (TODO: Should we check functions called or
```
ayrtonm, line 58: If we create call gates for undefined functions in each translation unit, then we will almost certainly have more call gates than we actually need but we'll only need to rerun the rewriter when more functions are declared (e.g. a new #include is added).

If we create call gates for functions that are called, then we'll have the minimum number of call gates we need, but we'll need to rerun the rewriter each time a new function call to another compartment is added.

ayrtonm: Went with the first option in #182 for now.

docs/source_rewriter.md
```diff
@@ -0,0 +1,210 @@
...
+compartment (TODO: What if a user forgets an input .c? maybe assert that these
+functions are declared in /usr/ headers).
+
+In phase two we'll generate the call gates for each direct transition. Since we
+know all the caller and target compartment we can hard-code the pkey in each
+call gate wrapper. These wrappers are inserted with `ld --wrap` so their names
+must start with `__wrap_`. Since a function may be called from multiple
+compartments and call gates have a hardcoded caller pkey, we need one wrapper
+per caller compartment. To distinguish wrappers for different compartments we'll
+name the wrappers `__wrap_$FUNCTION_NAME_from_$CALLER_PKEY`. This will require
```
rinon, line 70: What happened to defining these symbols with hidden visibility? I.e. define `__wrap_foo` separately in compartment1 and compartment2, each with hidden visibility then linking that into the DSO?

docs/source_rewriter.md
```diff
@@ -0,0 +1,210 @@
...
+This wrapper will be identical to the wrappers for external linkage except that
+it will not have `.global __ia2_fn_name` to make the wrapper static. This will
+make it easier to support the case where static functions in different files
+with the same name have their address taken. TODO: We write the
+`IA2_DEFINE_STATIC_WRAPPER_*` functions to one output header so we'd need a way
+to distinguish different files.
+
+### Call gates for indirect callsites
+
+We also need call gates for indirect callsites. Each compartment might not
```
rinon, line 151: Do we rewrite every indirect callsite? Is there a way to opt out for hot callsites known to only call functions internal to its own compartment?

Issue #165, opened 2022-09-12
Harden the compartment-aware allocator


ayrtonm:
PR #164 added protected heaps to the compartment-aware allocator. This now works as expected for both private and shared allocations, but we left the shim and [allocator's metadata](https://chromium.googlesource.com/chromium/src/+/master/base/allocator/partition_allocator/PartitionAlloc.md#architecture) unprotected. I haven't looked very much into what needs to be done for PA's metadata, but an easy place to start would be with the shim's data.

The allocator shim defines an array of pointers to the objects managing the different partitions which is completely unprotected. We should do something relro-like where we initialize it then make the page read-only. Alternatively we could put the pointers on separate pages and tag them accordingly but that may be tricky for the shared `PartitionRoot`. I only mention the second option because we have a similar issue with the compartment stack pointers #67. There we have another array of pointers which will almost certainly have to be on separate pages (stack pointers can't easily be marked read-only), so it may make sense to combine the allocator shim and stack metadata at some point.

rinon:
Allocator metadata should be protected by partition-alloc as of #280

fw-immunant:
Now that we have the syscall filtering runtime, we can straightforwardly define a protocol to make pages immutable, which should solve the shim half of things. If multiple threads can be present, we need to do an ABA type sequence of first writing data, then making the page immutable, then verifying that the data was not modified concurrently. Otherwise it's fine to simply write the data and freeze it.

PR #176, opened 2023-01-11
Add PartitionAlloc, its shims, and tests

endbr64:
This PR combines and supersedes #145 and #164 as we discussed today.


scripts/partition_alloc/update.sh
```diff
@@ -0,0 +1,69 @@
...
+NEW_CHROMIUM=$REPO_ROOT/chromium
+NEW_SRC=$NEW_CHROMIUM/src
+DEPOT_TOOLS=$REPO_ROOT/depot_tools
+
+# Download depot_tools
+if [[ -d "$DEPOT_TOOLS" ]];
+then
+	  rm -rf $DEPOT_TOOLS
+fi
+git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
```
fw-immunant, line 17: I'd rather this used `$DEPOT_TOOLS` instead of coincidentally populating a matching dir.

partition-alloc/src/allocator_shim.cc
```diff
@@ -0,0 +1,221 @@
...
+    return root->Alloc(bytes, nullptr);
+  }
+}
+
+void ShimFree(void *ptr) {
+  size_t pkey = ::ia2_get_pkey();
+  ShimFreeWithPkey(ptr, pkey);
+}
+
+void ShimFreeWithPkey(void *ptr, size_t pkey) {
```
fw-immunant, line 176: Can this (or `ShimReallocWithPkey` or `ShimCallocWithPkey`) be called maliciously with another compartment's pkey?

fw-immunant (commented):
One nit on the script, one security-relevant question on the shim that I'd like answered before r+.

partition-alloc/src/allocator_shim.cc
endbr64: I don't see why not. Looking at `SHIM_ALWAYS_EXPORT`, I think this lib is meant to be compiled with `-fvisibility=hidden`. Even then, would that prevent someone from jumping to the address just like the exported methods do?

@ayrtonm?

ayrtonm: Yeah this can currently be called with another compartment's pkey and fixing it would be part of #165. We can prevent this kind of attack with the second idea in [this comment](https://github.com/immunant/IA2-Phase2/issues/165#issue-1370194839) where we put the partition roots on separate pages with each compartment's pkey.

endbr64:
Removed the test failing due to #112 to make CI pass.

PR #177, opened 2023-01-11
WIP Add source rewriter prototype

ayrtonm:
This source rewriter is a WIP, but I'd like to get feedback on the assumptions it makes and anyone who wants to review it can focus on `main()` in SourceRewriter.cpp. You can also look at the output generated for the `minimal-main`, `trusted_indirect-main` and `untrusted_indirect-main` targets.

Right now you pass in all of a compartment's .c's, it rewrites those in-place and creates one .c, one .h and a linker args file for it. The generated .c has all call gate wrappers for calling from that compartment to the others (*except wrappers for static functions, see SourceRewriter.cpp:602). The generated .h has all opaque function pointer types, the generated `IA2_CALL` and `IA2_FN` macros and declarations for all wrappers for function pointers. This assumes that types with the same name are the same across source files within a compartment to avoid generating one .h per input .c and simplify file management. It also assumes that static functions that have their address taken have unique names. It also assumes that generic selection is available to remove most of the mangled types from the rewritten .c's.

Also I want to highlight two TODOs:
- opaque function pointer typedefs should be structs
- the rewriter assumes there are two trusted compartments

For the first point, using typedefs made the rewriting easier but typedefed pointers are unsafe since they can be cast to/from `void *` which is an operation that should be manually checked by a user. Making them structs means we have to add a pass to change `NULL` to `{NULL}` or `(typeof(ptr)) { NULL }` for function pointers. I wrote a pass that does this, but it still doesn't handle some corner cases where the `typeof` is required. I'd also need to add the pass to rewrite comparisons with `==` back in.

The second point refers to SourceRewriter.cpp:561. The issue is that when we take the address of a function in another compartment (see the last commit) we need to create a wrapper for that pointer and the target pkey is fixed at build-time. The problem is that we don't have a way to specify what compartment each undefined function is defined in and I'm not sure how we want to handle this.


ayrtonm:
I'm starting to think the right fix to the issue in the last paragraph is to pass in every compartment's sources at once and specify what belongs to which compartment instead of running the rewriter once per compartment. Maybe a good middle ground between that and what I have now is to limit the rewriter to either
- two trusted compartments or
- a trusted and untrusted compartment
and have a flag to specify which mode we're in. That should let us rewrite the change in the last commit correctly.

Also compiling the generated .c with `-DTARGET_PKEY` won't work correctly for more than two trusted compartments, but if we were to pass in all sources at once we wouldn't need to define it.

ayrtonm:
The idea of limiting the rewriter to two modes in my last comment doesn't actually solve the problem for two trusted compartments, e.g.
```
// This is all in compartment 1
typedef int(*fn)(int);
void set_callback(fn cb);
int fn_in_compartment1(int x) { ... }
int fn_in_compartment2(int x);
int fn_in_untrusted_compartment(int x);

set_callback(fn_in_compartment1);
set_callback(fn_in_compartment2);
set_callback(fn_in_untrusted_compartment); // e.g. could be a function in libc
```
The rewriter can see that `fn_in_compartment1` is defined in compartment 1 so it can generate the correct call gate wrapper. For `fn_in_compartment2` and `fn_in_untrusted_compartment` it only sees that the function is undefined not what compartment they're defined in. So I think we will just have to pass in every source file at once.

ayrtonm:
This PR has lots of unnecessary cmake changes and the refactoring in GenCallAsm.cpp doesn't exactly match the two PRs I opened so I'll reopen this prototype on top of #181.

PR #178, opened 2023-01-17
Fix build with Clang 15

fw-immunant:
Updated my machine and ran into an API change.


ayrtonm:
The change looks fine to me, but can we add clang 15 to CI?

ayrtonm (approved):

fw-immunant:
I hesitated because we also don't have Clang 14 in CI...

fw-immunant:
I'll follow up with a PR to see what explodes adding Clang 14+15 to CI.

ayrtonm:
#179 probably isn't the most important thing to debug at the moment but thanks for opening the PR so we at least know what the state of it it.

PR #179, opened 2023-01-17
Test against Clang 14 and 15 in CI

fw-immunant:


fw-immunant:
```
lit: /home/runner/.local/lib/python3.8/site-packages/lit/llvm/config.py:449: note: using clang: /usr/lib/llvm-15/bin/clang
lit: /home/runner/.local/lib/python3.8/site-packages/lit/llvm/config.py:314: fatal: Could not turn '' into Itanium ABI triple
```
from the [clang-15 build](https://github.com/immunant/IA2-Phase2/actions/runs/3944119989/jobs/6749705125)...

fw-immunant:
I managed to reproduce this locally. The problem is that our generated `lit.site.cfg.py` contains `config.target_triple = ""`, which is obviously invalid and trips up newer `lit`. Our template file contains `config.target_triple = "@TARGET_TRIPLE@"`, so it looks like `TARGET_TRIPLE` is empty.

fw-immunant:
Seems this is due to https://reviews.llvm.org/D119918.

PR #180, opened 2023-01-18
Refactor GenCallAsm.cpp

ayrtonm:
The source rewriter prototype in PR #177 required refactoring GenCallAsm.cpp so I'm splitting out those changes to make review easier.

A `caller_pkey` argument is useful since the rewriter will eventually read all source files and know which compartment all functions. This means it'll be able to insert the caller and target pkeys itself instead of relying on the "compile with -DCALLER_PKEY" hack.


fw-immunant (approved):
Sane refactor, LGTM.

PR #181, opened 2023-01-18
Add explicit third variant to WrapperKind enum in GenCallAsm.cpp

ayrtonm:
The three calls to `emit_asm_wrapper` generate three different types of wrappers. It accepts a `WrapperKind` enum with two variants which we used in combination with the presecence of an explicit `as_macro` argument to determine which of the three types of wrappers to generate. This commit adds an explicit 3rd `WrapperKind` variant and fixes the logic that relies on `as_macro` to use just the `WrapperKind` instead. The only thing that `as_macro` affects now is the terminator used when writing each line of asm.

This is part of the refactoring from PR #177 and the last one for now.


endbr64 (approved):
I don't fully understand `emit_asm_wrapper()`, but I think I understand the changes you're making. `WrapperKind::Direct` with `as_macro` set becomes `WrapperKind::Pointer` which shares most logic with `WrapperKind::Indirect` as another type of indirect call.

PR #182, opened 2023-01-18
Add source rewriter

ayrtonm:
~This just cherry picks SourceRewriter.cpp and a few commits from #177, but doesn't implement what's in the updated doc (#174) yet.~

This adds the source rewriter described in #174. To summarize, the new rewriter is run once per test and requires passing in all source files at once. The tool checks for `-DPKEY=N` in a compile_commands.json to determine what compartment each `.c` belongs to.

It then generates a single output `.c` with the following types of call gates:
- `__wrap_foo` for direct calls to `foo`, these are inserted by the ld args file
  - transitions directly from the caller to the target pkeys
  - the rewriter generates one ld args file per compartment with pkey != 0
  - TODO: we're limited to calling these wrappers from only one compartment. We'll enable calling `foo` directly from multiple compartments by renaming the wrapper symbols using `objcopy` on the intermediate `.o`s in a follow-up
- pointers to `foo` are replaced by `__ia2_foo` using the `IA2_FN` macro
  - assume the caller has pkey 0 since they may be called from any compartment.
- indirect callsites are replaced by calls to `__ia2_indirect_callgate_ID_pkey_PKEY` using the `IA2_CALL` macro
  - assumes the target has pkey 0 since it may call pointers from any compartment
  - ID is an arbitrary integer corresponding to each function signature
  - PKEY refers to the caller's pkey

It generates an output `.h` with the opaque structs that replace function pointer types and the macros above. The header may also include `IA2_DEFINE_WRAPPER_foo` macros to define call gates for pointers to static functions in each source file that defines a static function. The rewriter also invokes these macros at the bottom of each source file where necessary.

Minor implementation details: We replace fn ptr types with opaque structs to ensure that new code is compiled with wrappers when necessary. These structs cannot be set to/compared with `NULL` so we must rewrite these things. The `FnPtrNull` pass handles setting pointers to `NULL` while comparing with `NULL` is done manually for now (see simple1/main.c:70).

Note that the `FnPtrNull` pass rewrites

`ptr = NULL;` as `ptr = (fn_type) { NULL };`
`fn_type ptr = NULL;` as `fn_type ptr = { NULL };`

but does not rewrite `NULL` function pointers embedded in a larger struct. In these cases the language infers that the `NULL` which originally corresponded to the function pointer is the value of the opaque struct's field so it does not require a rewrite. We may want to rewrite these cases for clarity/completeness, but the complexity of that may not be worth it for now.


header-rewriter/SourceRewriter.cpp
```diff
@@ -0,0 +1,632 @@
...
+
+struct FnPtrTy {
+  std::string fn_ptr;
+  std::string opaque_struct;
+
+  bool operator==(const FnPtrTy &other) const {
+    return (this->fn_ptr == other.fn_ptr) &&
+           (this->opaque_struct == other.opaque_struct);
+  };
+  bool operator<(const FnPtrTy &other) const {
```
endbr64, line 74: Is it possible for `this->fn_ptr < other.fn_ptr` and `this->opaque_struct > other.opaque_struct` to both be true? Then `==` and `<` will always return `false`.

header-rewriter/tests/simple1/simple1.c
```diff
@@ -38,13 +39,13 @@ void simple_destroy(struct Simple *s) { free(s); }
...
   for (;;) {
-    int value = (*s->scb.read_cb)(s->state);
+    int value = (s->scb.read_cb)(s->state);
     if (value == 0) {
       break;
     }
 
-    value = (*map_fn)(value);
-    (*s->scb.write_cb)(value);
+    value = (map_fn)(value);
```
ahomescu, line 47: Without the dereference, you don't need the parentheses.

```suggestion
    value = map_fn(value);
```

header-rewriter/tests/simple1/simple1.c
```diff
@@ -38,13 +39,13 @@ void simple_destroy(struct Simple *s) { free(s); }
 
 void simple_foreach_v1(struct Simple *s, int (*map_fn)(int)) {
   for (;;) {
-    int value = (*s->scb.read_cb)(s->state);
+    int value = (s->scb.read_cb)(s->state);
```
ahomescu, line 42: Parentheses are optional here.

header-rewriter/SourceRewriter.cpp
ayrtonm: Removed this struct.

header-rewriter/tests/simple1/simple1.c
ayrtonm: done

ayrtonm:
Still need to modify the new rewriter to avoid rewriting function pointer types and omit annotations in untrusted compartments

fw-immunant:
@ayrtonm and I discussed a design point in Slack that should be documented until addressed: this branch introduces new sharing (of `ia2_fn_ptr`): the previous equivalent was a `static void* target_ptr` that was declared at each indirect call site, which made it specific to the wrapped library, and incidentally mprotected by being rw data of the shared object. `ia2_fn_ptr` is presently shared between compartments, but it shouldn't be. Something along the lines of splitting it into `void *ia2_fn_ptr_N;` (where `N` is the compartment id), perhaps declared in `ia2.h`, and referencing the calling compartment's `ia2_fn_ptr_N` in `GenCallAsm.cpp`.

This also needs to be thread-specific, which will have to be dealt with when threading support lands.

header-rewriter/tests/simple1/simple1.c
ayrtonm: We're not rewriting fn ptrs in this file anymore (it has pkey 0) so it's not relevant.

header-rewriter/SourceRewriter.cpp
```diff
@@ -0,0 +1,632 @@
...
+static std::string get_expansion_filename(const clang::Decl *decl) {
+  assert(decl != nullptr);
+  const auto &sm = decl->getASTContext().getSourceManager();
+  return sm.getFilename(sm.getExpansionLoc(decl->getLocation())).str();
+}
+
+static bool ignore_file(const std::string &file_name) {
+  auto file = llvm::StringRef(file_name);
+  // TODO: Add a proper blacklist with libia2 headers in it by default
+  return file.startswith("/usr/") || file.endswith("ia2.h") ||
```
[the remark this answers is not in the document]
ayrtonm: Clang might provide some API for checking that, but I don't thinkit's exactly what we want. Basically this function checks if a file is in a library that doesn't really belong to a single compartment like libc or libia2.

I agree that `startswith("/usr/")` is not a good solution, but I don't know if filtering out by file or symbol name is better since that requires a larger blacklist and we might miss things due to variations in libc implementations.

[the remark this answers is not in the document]
ayrtonm: Also `comp_cmds` only includes what's in the `compile_commands.json` so it won't have `-isystem` if you don't explicitly have it in your build system. I think finding the built-in `-isystem` paths would be a bit more involved than just reading from the .json which is why I didn't just implement this.

ayrtonm:
To recap an offline discussion, there's an issue with automating the tests. Basically the cmake targets have the flags for the compartmentalized builds which may differ from the normal builds (e.g. `--Wl,@ld_args_file -include output_header.h`). CMake uses these targets to generate the `compile_commands.json` so it'll include those flags. We then use the `compile_commands.json` to figure out how to preprocess sources when doing the rewriting. The issue is that there is a circular dependency because `output_header.h` and `ld_args_file` are generated by the rewriter, so we can't use them when the rewriter is preprocessing the sources.

One solution might be to split targets into compartmentalized and regular builds, but this will introduce a lot more complexity in the file management that we need to do. A simpler solution is to create empty `output_header.h` and ld_args files for each target that will need them. Then have the rewriter replace these empty files with the real ones. This means the `compile_commands.json` will still have `--Wl,@ld_args -include output_header.h`, but the flags do nothing when these files are empty. That means the rewriter will see those flags but preprocess the sources as if they weren't there.

rinon:
> This means the compile_commands.json will still have --Wl,@ld_args -include output_header.h, but the flags do nothing when these files are empty. That means the rewriter will see those flags but preprocess the sources as if they weren't there.

We don't actually build with `-include output_header.h` until after the rewriter runs, right? So this is idempotent and can be repeated?

rinon:
Does the user have to pass all the source files to rewrite on the command line all at once? Do you have a plan for how we should do that for larger projects where that might not really be feasible? Argument file?

header-rewriter/GenCallAsm.cpp
```diff
@@ -229,9 +220,14 @@ static void append_arg_kinds(std::stringstream &ss,
 }
 
 static std::string sig_string(const CAbiSignature &sig,
-                              const std::string &name) {
+                              const std::string *name) {
```
rinon, line 223: std::optional?

header-rewriter/SourceRewriter.cpp
```diff
@@ -0,0 +1,920 @@
...
+    }
+
+    assert(result.SourceManager != nullptr);
+    assert(result.Context != nullptr);
+    assert(old_decl != nullptr);
+
+    const auto &sm = *result.SourceManager;
+    auto &ctxt = *result.Context;
+
+    if (get_file_pkey(sm) == 0) {
```
rinon, line 181: Is zero not a valid pkey here?

rinon (approved):
I haven't gone over the new source rewriter in very close detail, but it looks like it will be reasonable to follow and maintain, great work!

cmake/define-test.cmake
```diff
@@ -1,108 +1,261 @@
...
-        target_compile_definitions(${LIBNAME} PUBLIC LIBIA2_INSECURE=1)
+        add_dependencies(${target} libia2)
+
+        if(LIBIA2_INSECURE)
+            target_compile_definitions(${target} PUBLIC LIBIA2_INSECURE=1)
+        endif()
+        if (DEFINED SHARED_LIB_PKEY)
+            target_compile_definitions(${target} PRIVATE PKEY=${SHARED_LIB_PKEY})
+        endif()
+        target_compile_definitions(${target} PRIVATE _GNU_SOURCE)
```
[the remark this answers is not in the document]
fw-immunant: I believe we need it for `dl_iterate_phdr`, see 7742edaf1e091310fecc02e3b82c355b5732f59e.

cmake/define-test.cmake
```diff
@@ -1,108 +1,261 @@
...
+        set(ORIGINAL_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/include)
+        set(REWRITTEN_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/include)
     endif()
+    
+    set(ORIGINAL_SRCS ${SHARED_LIB_SRCS})
+    list(TRANSFORM ORIGINAL_SRCS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
+    set(COPIED_SRCS ${SHARED_LIB_SRCS})
+    list(TRANSFORM COPIED_SRCS PREPEND ${CMAKE_CURRENT_BINARY_DIR}/)
+
+    add_custom_target(copy_files_${LIBNAME}
```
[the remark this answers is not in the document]
ayrtonm: `files(COPY` would copy the files when running cmake, but not at build time. The `${TEST}_call_gate_generation` targets runs the rewriter on the source files in the build directory so we need to copy the files there before each rewriter run. That's why the `call_gate_generation` targets depend on `${TEST}_copy_files`. We can check if/make the rewriter idempotent in a follow-up.

cmake/define-test.cmake
```diff
@@ -1,108 +1,261 @@
...
+    set(oneValueArgs LIBNAME PKEY)
+    set(multiValueArgs SRCS INCLUDE_DIR)
     cmake_parse_arguments(SHARED_LIB "${options}" "${oneValueArgs}"
                           "${multiValueArgs}" ${ARGN})
+
+    # Set library and wrapped library target names
     get_filename_component(TEST_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME)
     if(DEFINED SHARED_LIB_LIBNAME)
         set(LIBNAME ${SHARED_LIB_LIBNAME})
+        set(WRAPPED_LIBNAME ${SHARED_LIB_LIBNAME}_wrapped)
```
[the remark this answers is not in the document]
ayrtonm: done

cmake/define-test.cmake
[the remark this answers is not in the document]
ayrtonm: yep, it's for dl_iterate_phdr

header-rewriter/SourceRewriter.cpp
ayrtonm: Rewriter passes that modify source files (currently all except FnDecl) should change translation units with pkey 0. I'll add a comment to these

cmake/define-test.cmake
```diff
@@ -1,108 +1,261 @@
-# Defines a shared library target.
-#
-# Disables lazy binding (see issue #44). Unconditionally uses -fPIC.
+include(CMakePrintHelpers)
```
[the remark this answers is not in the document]
ayrtonm: removed

cmake/define-ia2-wrapper.cmake
```diff
@@ -1,199 +1,57 @@
...
-  OUTPUT_VARIABLE C_SYSTEM_INCLUDE_FIXED
-  OUTPUT_STRIP_TRAILING_WHITESPACE)
-
-# Define an IA2 shim library for direct call wrappers.
-#
-# This runs the rewriter on the headers exported by the library that will be
-# wrapped. Public headers are those that can be explicitly `#include`d by source
-# files and private headers are those that are used internally (i.e. require
-# context to be parsed. see issue #4).
+include(CMakePrintHelpers)
```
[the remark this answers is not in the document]
ayrtonm: removed

header-rewriter/GenCallAsm.cpp
ayrtonm: done

ayrtonm:
Tests are currently failing because the the `copy_files_${TEST}_main` and `copy_files_${TEST}_lib` targets for each test (that uses a single include directory) are both trying to copy the same headers to the build directory then rewriting them. This creates a race condition which is why tests passed, but were flaky when run locally.

I'll fix this by just having the `copy_files_${TEST}_main` targets copy over the headers if the include directory is not specified.

fw-immunant:
I don't understand `Keep static functions that have their address taken`. Shouldn't taking the address constitute a use for the purposes of optimization? Which test/function was having issues that this fixed?

ayrtonm:
> I don't understand Keep static functions that have their address taken. Shouldn't taking the address constitute a use for the purposes of optimization? Which test/function was having issues that this fixed?

Sorry the commit msg was shorthand for the actual issue. The problem is that the call gate wrapper for the static function has its address taken (so it's not marked as dead code), but the original static function is only referenced in the call gate's asm statement. I think everything in the asm statement is a black box to the compiler which is why this isn't recognized as a use.

The tests modified in that commit are the only ones currently affected and the fix I was going to propose is to have the source rewriter add those `__attribute__((used))` annotations where necessary.

ayrtonm:
I considered `-fno-[tree-]dce/-femit-all-decls/-Wl,--no-gc-sections` (for gcc/clang/ld) as an alternative, but I still had functions eliminated at compile-time with gcc and `-O3 -fno-dce` so I didn't look into that option very much.

ayrtonm:
It may be worth looking at assembler directives to see if we can add something that has the same effect as the `used` attribute to the generated call gates instead

Issue #167, opened 2022-10-10
Do not leave `mmap`ed memory shared by default


endbr64:
The default behavior when the memory is mapped leaves it accessible to all compartments. We should consider protecting it by hooking `mmap()` or inserting `mprotect()` calls.

fw-immunant:
Is there a race between `mmap()` and `mprotect()` in a multithreaded context? We may need to map as read-only before `pkey_mprotect` and then changing the mapping itself to RW.

rinon:
I think we could map read-only, then pkey_mprotect the pages if we hook mmap, but this has to be done in a way that is transparent to the dynamic linker.

PR #183, opened 2023-02-01
Implement threading support.

fw-immunant:
There is an outstanding hardening TODO which shouldn't block this as it's tracked by #167:
- We `mmap` a thunk when starting threads, but do not `pkey_mprotect` it to prevent tampering from other compartments. This isn't trivial to fix (e.g. simply calling `pkey_mprotect` explicitly) because we want to protect the memory with the current pkey (which is inherited from the thread that calls `pthread_create`), not a fixed one. A similar concern applies to any internal calls using `mmap` directly for memory allocation that a compartmentalized library might make.

This has some interaction with #182:
- the header-rewriter/source-rewriter changes will need merging.
- `ia2_fn_ptr(_N)` interaction (see [comment](https://github.com/immunant/IA2-Phase2/pull/182#issuecomment-1411273158)).


libia2/include/ia2.h
```diff
@@ -22,10 +22,12 @@
 
 #ifdef LIBIA2_INSECURE
 #define INIT_COMPARTMENT(n)
-#define IA2_WRPKRU
+#define IA2_WRPKRU ""
+#define IA2_RDPKRU ""
 #else
 #define INIT_COMPARTMENT(n) _INIT_COMPARTMENT(n)
 #define IA2_WRPKRU "wrpkru"
+#define IA2_RDPKRU "rdpkru"
```
ayrtonm, line 30: #182 is moving away from having the wrappers depend on ia2.h [(SourceRewriter.cpp:631)](https://github.com/immunant/IA2-Phase2/pull/182/files#diff-ad4877eea1240e8a16c89351f2759f73775a01b7666d3744cad36319e32b7db9R631). If rdpkru is just for this test I'd lean towards having the macro defined in one of the test's headers.

libia2/include/ia2.h
```diff
@@ -264,13 +254,86 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
       *n_to_alloc = 0;                                                         \
     }                                                                          \
   }                                                                            \
-  char *ia2_stackptrs[n + 1] IA2_SHARED_DATA;                                  \
-  __attribute__((constructor)) static void init_stacks() {                     \
-    ensure_pkeys_allocated(&ia2_n_pkeys_to_alloc);                             \
+                                                                               \
+  /* one page of stack pointers for each compartment */                        \
+  char *ia2_stackptrs[PAGE_SIZE * (n + 1)] __attribute__((aligned(PAGE_SIZE))) \
+  IA2_SHARED_DATA;                                                             \
```
ayrtonm, line 260: Move `ia2_stackptrs` out of `ia2_shared_data` if we're going to protect it?

libia2/include/ia2.h
ayrtonm: nvm I just saw that this is also used in `allocate_compartment_stack`

libia2/include/ia2.h
```diff
@@ -264,13 +254,86 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+    char *stack_ptr_value = stack + STACK_SIZE - 8;                            \
+                                                                               \
+    /* Save the current pkru. */                                               \
+    uint32_t old_pkru = 0;                                                     \
+    __asm__ volatile(IA2_RDPKRU : "=a"(old_pkru) : "a"(0), "d"(0), "c"(0));    \
+    /* Switch to the compartment's pkru to access its stack pointer. */        \
+    uint32_t new_pkru = ~((3 << (2 * i)) | 3);                                 \
+    __asm__ volatile(IA2_WRPKRU : : "a"(new_pkru), "d"(0), "c"(0));            \
+    /* Forbid overwriting an existing stack. */                                \
+    if (ia2_stackptrs[stack_idx] != NULL) {                                    \
```
ayrtonm, line 310: Don't we need to explicitly initialize `ia2_stackptrs`?

libia2/include/ia2.h
```diff
@@ -264,13 +254,86 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+      }                                                                        \
+    }                                                                          \
+  }                                                                            \
+                                                                               \
+  /* Allocate and protect the stack for this thread's \p i'th compartment. */  \
+  /* Stores the stack in the appropriate slot of ia2_stackptrs. */             \
+                                                                               \
+  /* I'm worried about interference from other threads if any sensitive */     \
+  /* locals of in this function get spilled to the stack. We probably need */  \
+  /* to write this in assembly. */                                             \
```
ayrtonm, line 280: I agree with this comment. Also if `old_pkru` can be any compartment this function might trigger an mpk violation after the wrpkru if it access the stack. We could avoid that by switching to a mixed pkru, but I think it's better to avoid that when possible.

libia2/include/ia2.h
fw-immunant: My reasoning was that `ia2_stackptrs` is, in aggregate, shared, so there's no one compartment for which we want to mprotect it... but we do need to `pkey_mprotect` each page of it based on the compartment whose stack pointers it will hold.

Because we do `pkey_mprotect` each page of it, though, I guess it's also fine to not be `IA2_SHARED_DATA`--I tested it and it does work .

libia2/include/ia2.h
```diff
@@ -264,13 +254,86 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+                                                                               \
+  extern void ia2_init_threads(void);                                          \
+                                                                               \
+  __attribute__((constructor)) static void ia2_init(void) {                    \
+    /* set up global resources */                                              \
+    ia2_init_threads();                                                        \
+    ensure_pkeys_allocated(&ia2_n_pkeys_to_alloc);                             \
+    protect_stackptrs();                                                       \
+    /* initialize stacks for the main thread */                                \
+    init_stacks();                                                             \
```
ayrtonm, line 338: This call wouldn't need the wrpkru back to `old_pkru`, but maybe we can fix that in a follow-up

libia2/include/ia2.h
fw-immunant: We don't, because it has static storage duration, but we could for clarity: https://stackoverflow.com/questions/16015656/are-global-variables-always-initialized-to-zero-in-c

libia2/include/ia2.h
fw-immunant: This isn't exactly true: `init_stacks` allocates a stack for each compartment, in numerical order, so we can't just stop at the current compartment if it isn't the last one.

libia2/threads.c
```diff
@@ -0,0 +1,85 @@
...
+}
+
+__thread size_t ia2_compacted_tid;
+
+size_t ia2_get_compacted_tid(void) { return ia2_compacted_tid; }
+
+_Noreturn extern void __real_pthread_exit(void *retval);
+
+_Noreturn void ia2_release_own_compacted_tid(void *retval) {
+  thread_exists_bitmap[ia2_compacted_tid] = 0;
```
ayrtonm, line 69: Don't we need `pthread_mutex_lock` here?

libia2/include/ia2.h
ayrtonm: yeah it's not a correctness thing, either should work. I was thinking that it might be confusing to debug later if we see something in ia2_shared_data getting pkey_mprotected

libia2/include/ia2.h
ayrtonm: Oh right we're calling wrpkru in a loop

ayrtonm:
The libia2 changes mostly LGTM. I'll review the GenCallAsm changes later today and think about the compacted tid issue a bit more

libia2/threads.c
fw-immunant: I think you're right, we do. I had thought we might be safe because there's only one concurrent writer to this index, since obtaining ownership of that index with `ia2_alloc_compated_tid` is locked, but that relies on `ia2_compacted_tid` always being correct, and still means we have synchronization around readers and (setting) writers but not around clearing writers... there's no reason not to lock here.

fw-immunant:
We discussed some security/design concerns around `thread_exists_bitmap` and `ia2_compacted_tid` in Slack.

`ia2_compacted_tid`:

`ia2_compacted_tid` is used to determine the location to store/load compartments' stack pointers during compartment transition. This means that if it gets modified during a thread's lifetime, a compartment transition in that thread could result in the instances of one compartment in different threads sharing a stack, and subsequently result in corruption of the stack contents. If one compartment can modify ia2_compacted_tid such that this corruption happens to another compartment, that constitutes compartment escape.

As such, to ensure that no compartment can modify the tid that will be used to compute the stack pointer storage location for another compartment, we either need to have per-compartment copies of `ia2_compacted_tid`, or we need to make it read-only after thread initialization.

The latter could be done by `pkey_mprotecting` a thread's copy of `ia2_compacted_tid` itself to read-only for all compartments, but this might require carving out a whole page to store it, to avoid pkey_mprotecting other unrelated TLS data. I'm also not sure if it's possible to do this in a non-racy way--I'm not yet sure what the solution to compartment-isolating TLS data in general looks like.

`thread_exists_bitmap`:

The idea behind having `thread_exists_bitmap` as separate from simply checking `ia2_stackptrs` is basically hoping to have a compartment-independent source of truth about which threads exist, which we can use to compute a compartment-independent `ia2_compacted_tid` at thread startup. The discussion above seems to indicate that we might be able to have `ia2_compacted_tid` itself be compartment-independent if we make it read-only after thread startup. But I'm not sure how to make `thread_exists_bitmap` safe. Any compartment needs to be able to start or finish a thread, so both cases of accesses to `thread_exists_bitmap` seem fraught. `ia2_stackptrs`, on the other hand, is already sharded by compartment, but since any compartment could zero its own stack pointers, there's no compartment that can be considered authoritative for purposes of checking if it has a stack pointer saved for a given tid. A compartment running in two threads could zero one of its stack pointers from the other thread, then start a third thread, which would confuse itself with the first thread even when transitioning to a second compartment.

So it may be that the only viable design is sharding both `thread_exists_bitmap` and `ia2_compacted_tid`, and having every compartment compute them separately. This would mean accessing two thread-locals during every compartment transition, rather than getting away with only doing it once as we do now.

header-rewriter/GenCallAsm.cpp
```diff
@@ -216,6 +216,72 @@ static void emit_mixed_wrpkru(AsmWriter &aw, const std::string &pkey0,
...
+// new stack pointer.
+// \p sp_addr_reg and \p compartment_offset_reg are register names (sans %
+// prefix) that will be clobbered.
+static void emit_switch_stacks(AsmWriter &aw, const std::string &old_pkey,
+                               const std::string &new_pkey,
+                               const std::string &sp_addr_reg,
+                               const std::string &compartment_offset_reg) {
+  add_comment_line(
+      aw, llvm::formatv(
+              "Compute location to save old stack pointer (in {0}, using {0})",
```
ayrtonm, line 260: is the second `{0}` supposed to be `{1}`?

header-rewriter/GenCallAsm.cpp
```diff
@@ -216,6 +216,72 @@ static void emit_mixed_wrpkru(AsmWriter &aw, const std::string &pkey0,
...
+  add_asm_line(
+      aw, llvm::formatv("sub %{0}, %{1}", compartment_offset_reg, sp_addr_reg));
+  // Compute new compartment offset in compartment_offset_reg
+  emit_load_pkey(aw, new_pkey, compartment_offset_reg);
+  add_asm_line(aw, llvm::formatv("shl $12, %{0}", compartment_offset_reg));
+  // Add new compartment offset to thread offset
+  add_asm_line(
+      aw, llvm::formatv("add %{0}, %{1}", compartment_offset_reg, sp_addr_reg));
+  add_comment_line(aw, "Read the new stack pointer from memory");
+  add_asm_line(aw, llvm::formatv("movq (%{0}), %rsp", sp_addr_reg));
```
ahomescu, line 282: Minor nit: you might avoid having to read and then write `rsp` separately if you do something like `rsp += (new_offset - old_offset)`.

header-rewriter/GenCallAsm.cpp
```diff
@@ -216,6 +216,72 @@ static void emit_mixed_wrpkru(AsmWriter &aw, const std::string &pkey0,
...
+// \p sp_addr_reg and \p compartment_offset_reg are register names (sans %
+// prefix); after this asm runs, sp_addr_reg will hold the address where the
+// compartment's stack pointer is stored, and compartment_offset_reg will hold
+// a compartment-specific offset.
+static void emit_load_sp_addr(AsmWriter &aw, const std::string &pkey,
+                              const std::string &sp_addr_reg,
+                              const std::string &compartment_offset_reg) {
+  // Compute compartment offset in compartment_offset_reg
+  emit_load_pkey(aw, pkey, compartment_offset_reg);
+  add_asm_line(aw, llvm::formatv("shl $12, %{0}", compartment_offset_reg));
```
ahomescu, line 234: Guessing `$12` here is the page size? I wonder if we want to move the constant from the output assembly text into the C++ code, and maybe even compute it at run time (or maybe as an IA2 build constant).

ahomescu (commented):
Did a quick drive-by review, I had a little bit of spare time.

header-rewriter/GenCallAsm.cpp
ayrtonm: This runtime computation of the page offset will be switched to a codegen-time constant when #182 gets merged since that PR changes the caller/target pkeys in emit_asm_wrapper to known integers instead of macros. So I think it's probably fine as-is for now.

header-rewriter/GenCallAsm.cpp
fw-immunant: I don't think we can do that here, because the new and old rsp values have both different bases (different stacks) and different offsets (different location in the stack). We actually never examine the value of rsp at all in this code, just store the old one and load the new one. As such, there's no inherent relation between the old and new values of rsp. We *do* do such a delta optimization on the *location* that the old rsp is saved in memory and where the new rsp is loaded from; that's what's happening around L271.

libia2/threads.c
fw-immunant: Fixed.

rinon:
Looking through this now, but I'm gonna be _that_ guy, because I think it's important. Do you have any documentation of the overall architecture of this threading support? If not, could you write something? Doesn't have to be detailed or exhaustive, just a few paragraphs and maybe a picture if relevant. Just enough for someone to get an idea of the high-level idea, without all the implementations details getting in the way.

libia2/include/ia2.h
rinon: Does that solve the problem though? Are there any ways that registers can be spilled underneath us in process memory?

ayrtonm: The problem of spilling onto the stack or are you thinking about things like thread-locals? I don't see what the issue is if we don't write to the stack in this asm function

fw-immunant: As far as I can tell that will solve the problem. We generally rely on registers being safe from outside interference in IA2, e.g. in stack switching and register scrubbing.

fw-immunant:
> Looking through this now, but I'm gonna be _that_ guy, because I think it's important. Do you have any documentation of the overall architecture of this threading support? If not, could you write something? Doesn't have to be detailed or exhaustive, just a few paragraphs and maybe a picture if relevant. Just enough for someone to get an idea of the high-level idea, without all the implementations details getting in the way.

See the added [`docs/threading.md`](https://github.com/immunant/IA2-Phase2/blob/fw/threading/docs/threading.md).

header-rewriter/GenCallAsm.cpp
```diff
@@ -216,6 +216,72 @@ static void emit_mixed_wrpkru(AsmWriter &aw, const std::string &pkey0,
...
+  // Compute thread offset in sp_addr_reg, adding compartment offset too
+
+  add_asm_line(aw, llvm::formatv("mov ia2_compacted_tid@GOTTPOFF(%rip), %{0}",
+                                 sp_addr_reg));
+  add_asm_line(aw, llvm::formatv("mov %fs:(%{0}), %{0}", sp_addr_reg));
+
+  // add_asm_line(aw, llvm::formatv("mov thread_id, %{0}", sp_addr_reg));
```
rinon, line 241: Leftover?

header-rewriter/GenCallAsm.cpp
```diff
@@ -216,6 +216,72 @@ static void emit_mixed_wrpkru(AsmWriter &aw, const std::string &pkey0,
...
+
+  add_asm_line(aw, llvm::formatv("lea (%{0}, %{1},8), %{1}",
```
rinon, line 243: This line could use a comment, not obvious what this does at first glance.

rinon (commented):
I haven't looked at the test, but I assume that's fine.

libia2/include/ia2.h
```diff
@@ -264,13 +254,94 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+    }                                                                          \
+    /* Write the stack pointer. */                                             \
+    ia2_stackptrs[stack_idx] = stack_ptr_value;                                \
+    /* Restore the old pkru. */                                                \
+    __asm__(IA2_WRPKRU : : "a"(old_pkru), "d"(0), "c"(0));                     \
+  }                                                                            \
+                                                                               \
+  size_t ia2_alloc_compacted_tid(void);                                        \
+                                                                               \
+  void init_stacks(void) {                                                     \
```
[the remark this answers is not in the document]
fw-immunant: The definition of `init_stacks` needs to statically depend (via the macro argument `n`) on the number of compartments that exist in the compartmentalized program, so it needs to be defined inside the `INIT_RUNTIME` (actually `INIT_RUNTIME_COMMON`) macro in the `ia2.h` header. It ends up in the main compartment's library rather than linked in via libia2.

libia2/include/ia2.h
fw-immunant: This is now all obsolete as we use thread-local `ia2_stackptr_N` variables instead of an `ia2_stackptrs` array.

libia2/include/ia2.h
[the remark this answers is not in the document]
rinon: Got it, alright.

libia2/include/ia2.h
```diff
@@ -264,13 +254,94 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+    if (ia2_stackptrs[stack_idx] != NULL) {                                    \
+      printf(                                                                  \
+          "compartment %zd in thread %zd tried to allocate existing stack\n",  \
+          i, tid);                                                             \
+      exit(1);                                                                 \
+    }                                                                          \
+    /* Write the stack pointer. */                                             \
+    ia2_stackptrs[stack_idx] = stack_ptr_value;                                \
+    /* Restore the old pkru. */                                                \
+    __asm__(IA2_WRPKRU : : "a"(old_pkru), "d"(0), "c"(0));                     \
```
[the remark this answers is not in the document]
fw-immunant: We now use `__asm__ volatile` now that this entire bit is rewritten in asm.

fw-immunant:
I've rewritten the portion of `ALLOCATE_COMPARTMENT_STACK` that alters the PKRU and writes the stack pointer in asm to prevent its locals from spilling and being tampered with by other compartments in preexisting threads. This was a significant security concern because this macro contains code that does a `wrpkru` with a pkey that is not simply computed immediately prior to the `wkrpru`, as it must switch to each pkey before returning to whatever the pkey was before switching.

libia2/include/ia2.h
```diff
@@ -268,13 +367,54 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
-  char *ia2_stackptrs[n + 1] IA2_SHARED_DATA;                                  \
-  __attribute__((constructor)) static void init_stacks() {                     \
-    ensure_pkeys_allocated(&ia2_n_pkeys_to_alloc);                             \
-    for (int i = 0; i < n + 1; i++) {                                          \
-      char *stack_start = allocate_stack(i);                                   \
-      protect_stack(i, stack_start);                                           \
-      /* Each stack frame start + 8 is initially 16-byte aligned. */           \
-      ia2_stackptrs[i] = stack_start + STACK_SIZE - 8;                         \
+  /* Forbid overwriting an existing stack. */                                  \
+  void ia2_reinit_stack_err(int i) {                                           \
```
rinon, line 371: noreturn?

libia2/include/ia2.h
```diff
@@ -226,28 +228,125 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+                     "mov_pkru_eax " #i "\n" IA2_WRPKRU                        \
+                     "\n"                                                      \
+                     "mov ia2_stackptr_" #i                                    \
+                     "@GOTTPOFF(%%rip), %%r11\n"                               \
+                     "cmpq $0x0,%%fs:(%%r11)\n"                                \
+                     "je .Lfresh_init" #i                                      \
+                     "\n"                                                      \
+                     "mov $" #i                                                \
+                     ", %%rdi\n"                                               \
+                     "call ia2_reinit_stack_err\n"                             \
```
rinon, line 264: I would add a `ud2` after this call just to make sure it never returns, as returning would be a problem.

libia2/include/ia2.h
```diff
@@ -226,28 +228,125 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+                     ":\n"                                                     \
+                     "# store the stack addr in the stack pointer\n"           \
+                     "mov %%rbp, %%fs:(%%r11)\n"                               \
+                     "# restore old pkru\n"                                    \
+                     "mov %%r12d,%%eax\n" IA2_WRPKRU                           \
+                     "\n"                                                      \
+                     "pop %%rbp\n"                                             \
+                     :                                                         \
+                     :                                                         \
+                     : "rdi", "rax", "rbx", "rcx", "rdx", "r11", "r12");       \
```
rinon, line 275: Can you just clobber rbp and not add the pushes?

rinon: Also do we have to put all the caller-saved regs in the clobber list because we make function calls? I don't remember extended asm well enough off the top of my head...

libia2/include/ia2.h
```diff
@@ -226,28 +228,125 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+  }
+
+#define INIT_RUNTIME_COMMON(n)                                                 \
+  /* The 0th compartment is unprivileged and does not protect its memory, */   \
+  /* so declare its stack pointer in the shared object that sets up the */     \
+  /* runtime instead of using the INIT_COMPARTMENT macro for it. */            \
+  __thread void *ia2_stackptr_0;                                               \
```
rinon, line 282: I would put this down with the definition of `init_stacks` because that's where it is used. Do you need it declared higher?

libia2/include/ia2.h
```diff
@@ -226,28 +228,125 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+  void init_tls_8(void);                                                       \
+  void init_tls_7(void);                                                       \
+  void init_tls_6(void);                                                       \
+  void init_tls_5(void);                                                       \
+  void init_tls_4(void);                                                       \
+  void init_tls_3(void);                                                       \
+  void init_tls_2(void);                                                       \
+  void init_tls_1(void);                                                       \
+  /* Ensure that TLS is protected in a new thread. */                          \
+  void protect_tls(void) {                                                     \
```
rinon, line 318: This is a strong, public symbol. Won't it conflict between shared libs? This could use better documentation of why you did the goto thing here, why not just call `protect_tls_for_compartment(n)`?

libia2/threads.c
```diff
@@ -0,0 +1,52 @@
...
+int __real_pthread_create(pthread_t *restrict thread,
+                          const pthread_attr_t *restrict attr, void *(*fn)(),
+                          void *data);
+
+int __wrap_pthread_create(pthread_t *restrict thread,
+                          const pthread_attr_t *restrict attr, void *(*fn)(),
+                          void *data) {
+  /* Allocate a thunk for the thread to call `ia2_thread_begin` before the
+  provided thread body function. */
+  void *mmap_res = mmap(NULL, sizeof(struct ia2_thread_thunk),
```
rinon, line 37: I'm guessing you don't want to use malloc here, but noting why would be nice.

libia2/threads.c
```diff
@@ -0,0 +1,52 @@
...
+  }
+  struct ia2_thread_thunk *thread_thunk = (struct ia2_thread_thunk *)mmap_res;
+  thread_thunk->fn = fn;
+  thread_thunk->data = data;
+  return __real_pthread_create(thread, attr, ia2_thread_begin, thread_thunk);
+}
+
+_Noreturn void __real_pthread_exit(void *retval);
+
+_Noreturn void __wrap_pthread_exit(void *retval) {
```
rinon, line 50: Why are we wrapping this?

libia2/include/ia2.h
```diff
@@ -226,28 +228,125 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
...
+  {                                                                            \
+    extern __thread void *ia2_stackptr_##i;                                    \
+                                                                               \
+    /* We must change the pkru to write the stack pointer because each */      \
+    /* stack pointer is part of the compartment whose stack it points to. */   \
                                                                                \
-  /* Allocate a fixed-size stack. */                                           \
-  static char *allocate_stack(int i) {                                         \
+    __asm__ volatile("mov $" #i                                                \
+                     ",%%rdi\n"                                                \
```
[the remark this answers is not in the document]
fw-immunant: clang-format utterly mangles inline asm in a macro like this. In fact, it somehow even decided to use 81 columns for the whole thing. I'll go back and turn off clang-format for this whole bit.

libia2/include/ia2.h
fw-immunant: Clobbering rbp is somewhat contentious: see [some comments here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104805). We could use another register instead, e.g. r10.

I believe you're right, we should mark caller-save regs as clobbered. That's kind of a pain, but we can also relatively easily just pass a single variable that we ask gcc to assign a register into our asm and move the call to `allocate_stack` out. I'll do that.

rinon: hmmm yeah I'd be in favor of both pulling out the call and using a different reg

libia2/include/ia2.h
fw-immunant: No need to have it declared higher; I'll move it as suggested.

libia2/threads.c
fw-immunant: We don't need to anymore. I can remove it.

libia2/include/ia2.h
fw-immunant: It will not conflict between shared libs because this definition is in the `INIT_RUNTIME_COMMON` macro; `INIT_RUNTIME` should only appear in the main compartment, so there is only one definition of the function.

The way we have our code laid out and split between libia2 itself and the main compartment is really not ideal. `ia2.h` contains:
- IA2 interface declarations, like a normal header (e.g. for internal use by libia2 to refer to symbols present in the main compartment or in libia2 itself)
- definitions of macros that have the side-effect of defining per-compartment symbols such as `init_pkey_##n##_ctor`
- definitions of macros that have the side-effect of defining per-program symbols built into the main compartment, such as `protect_tls` and `ia2_stackptr_0`

It gets confusing which definitions are where for what reasons.

I'll document the `goto` hack. We do an equivalent thing with `switch` in `init_stacks`, and I couldn't decide which is cleaner... I have both for variety and hoped to get some feedback on which is preferable. The reason either hack is necessary is that we want to unroll the pseudo-loop of `for(int i=0; i<n; i++) { macro(i) }` for a compile-time integer `n`, in order to get straight-line code that only does the required work for the number of compartments that the program actually uses. For example, from `two_keys_minimal`:
```asm

00000000000028f0 <init_stacks>:
    28f0:	41 54                	push   %r12
    28f2:	bf 02 00 00 00       	mov    $0x2,%edi
    28f7:	53                   	push   %rbx
    28f8:	48 83 ec 08          	sub    $0x8,%rsp
    28fc:	e8 af fd ff ff       	call   26b0 <allocate_stack>
    2901:	49 89 c2             	mov    %rax,%r10
    2904:	31 c0                	xor    %eax,%eax
    2906:	89 c3                	mov    %eax,%ebx
    2908:	89 c1                	mov    %eax,%ecx
    290a:	89 c2                	mov    %eax,%edx
    290c:	41 89 c4             	mov    %eax,%r12d
    290f:	b8 cc ff ff ff       	mov    $0xffffffcc,%eax
    2914:	4c 8b 1d dd 26 00 00 	mov    0x26dd(%rip),%r11        # 4ff8 <ia2_stackptr_2@Base>
    291b:	64 49 83 3b 00       	cmpq   $0x0,%fs:(%r11)
    2920:	74 0e                	je     2930 <init_stacks+0x40>
    2922:	48 c7 c7 02 00 00 00 	mov    $0x2,%rdi
    2929:	e8 92 ff ff ff       	call   28c0 <ia2_reinit_stack_err>
    292e:	0f 0b                	ud2
    2930:	64 4d 89 13          	mov    %r10,%fs:(%r11)
    2934:	44 89 e0             	mov    %r12d,%eax
    2937:	bf 01 00 00 00       	mov    $0x1,%edi
    293c:	e8 6f fd ff ff       	call   26b0 <allocate_stack>
    2941:	49 89 c2             	mov    %rax,%r10
    2944:	31 c0                	xor    %eax,%eax
    2946:	89 c3                	mov    %eax,%ebx
    2948:	89 c1                	mov    %eax,%ecx
    294a:	89 c2                	mov    %eax,%edx
    294c:	41 89 c4             	mov    %eax,%r12d
    294f:	b8 f0 ff ff ff       	mov    $0xfffffff0,%eax
    2954:	4c 8b 1d 75 26 00 00 	mov    0x2675(%rip),%r11        # 4fd0 <ia2_stackptr_1@@Base+0x4fd0>
    295b:	64 49 83 3b 00       	cmpq   $0x0,%fs:(%r11)
    2960:	74 0e                	je     2970 <init_stacks+0x80>
    2962:	48 c7 c7 01 00 00 00 	mov    $0x1,%rdi
    2969:	e8 52 ff ff ff       	call   28c0 <ia2_reinit_stack_err>
    296e:	0f 0b                	ud2
    2970:	64 4d 89 13          	mov    %r10,%fs:(%r11)
    2974:	44 89 e0             	mov    %r12d,%eax
    2977:	31 ff                	xor    %edi,%edi
    2979:	e8 32 fd ff ff       	call   26b0 <allocate_stack>
    297e:	48 89 c3             	mov    %rax,%rbx
    2981:	64 48 8b 04 25 00 00 00 00 	mov    %fs:0x0,%rax
    298a:	48 03 05 2f 26 00 00 	add    0x262f(%rip),%rax        # 4fc0 <ia2_stackptr_0@@Base+0x4fb8>
    2991:	48 89 18             	mov    %rbx,(%rax)
    2994:	48 83 c4 08          	add    $0x8,%rsp
    2998:	5b                   	pop    %rbx
    2999:	41 5c                	pop    %r12
    299b:	c3                   	ret
```

fw-immunant: We can, at least, namespace this as `ia2_protect_tls`.

libia2/include/ia2.h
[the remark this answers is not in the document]
fw-immunant: Fixed.

libia2/include/ia2.h
fw-immunant: Done.

libia2/include/ia2.h
fw-immunant: Done.

libia2/include/ia2.h
fw-immunant: Done.

libia2/include/ia2.h
fw-immunant: I documented the static dispatch hack. I'd still like to settle on one of `goto label##n`/`switch(literal)`, and I'll namespace the symbol.

libia2/threads.c
fw-immunant: Done.

libia2/threads.c
fw-immunant: Done.

PR #193, opened 2023-03-14
Add a recursive call test

rinon:
Adds a test that makes mutually recursive calls between two compartments.

### DIFF
--- a/header-rewriter/tests/recursion/CMakeLists.txt
+++ b/header-rewriter/tests/recursion/CMakeLists.txt
@@ -1,0 +1,36 @@
+if(LIBIA2_INSECURE)
+    set(STDOUT_REF ${CMAKE_CURRENT_SOURCE_DIR}/Output/recursion.insecure.out)
+else()
+    set(STDOUT_REF ${CMAKE_CURRENT_SOURCE_DIR}/Output/recursion.out)
+endif()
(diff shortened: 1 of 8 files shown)


fw-immunant (approved):
LGTM.

header-rewriter/tests/recursion/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+    HEADERS recursion_main.h
+    INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/include/main/
+    CALLER_PKEY 2
+)
+
+# Build the recursion lib
+define_shared_lib(
+    SRCS dso.c
+    LINK_LIBS recursion-main-wrapper
+    LINK_OPTS "-Wl,-T${libia2_BINARY_DIR}/padding.ld"
```
fw-immunant, line 29: This doesn't need to block merging, but I've forgotten: what exactly requires the padding linker script? Some of our tests define shared libs with a `LINK_OPTS` entry for it (mostly ones that initialize two compartments?) and others don't; cf. `ro_sharing` (does) and `tls_protected` (doesn't).

rinon: afaik it's only needed to align shared globals to isolated pages.

Issue #198, opened 2023-04-04
Add rewriter pass for function pointers that are compared or implicitly cast to bool


ayrtonm:
Since the rewriter changes function pointers to structs with a single pointer field, comparisons and implicit casts to bool are not allowed. This occurs once in the simple1 test and in #182 we worked around it by replacing the struct with the struct's field in the compartmentalized build.

This should be done automatically by the source rewriter for at least the following cases
```
#include <stdio.h>
#include <stdbool.h>
#define IA2_ADDR(fn) fn.ptr
#define IA2_WRAPPER_ADDR(fn) IA2_FN(fn).ptr

typedef int (*bin_op)(int, int);

int add(int x, int y) {
    return x + y;
}

int main() {
    bin_op fn = add;
    bin_op fn2 = NULL;
    if (fn) {
    // if (IA2_ADDR(fn)) {
    }
    if (!fn) {
    // if (!IA2_ADDR(fn)) {
    }
    bool set = fn;
    //bool set = IA2_ADDR(fn);
    if (fn == fn2) {
    //if (IA2_ADDR(fn) == IA2_ADDR(fn2)) {
    }
    if (fn == add) {
    //if (IA2_ADDR(fn) == IA2_WRAPPER_ADDR(add)) {
    }
    if (fn == NULL) {
    //if (IA2_ADDR(fn) == NULL) {
    }
    if (fn != fn2) {
    //if (IA2_ADDR(fn) != IA2_ADDR(fn2)) {
    }
    if (fn != add) {
    //if (IA2_ADDR(fn) != IA2_WRAPPER_ADDR(add)) {
    }
    if (fn != NULL) {
    //if (IA2_ADDR(fn) != NULL) {
    }
}
```

ayrtonm:
Assigned as high priority since this came up frequently in nginx. I worked around it by making the changes manually then applying a patch each time I ran `make clean` which is far from ideal.

ayrtonm:
```
    bool set = fn;
    //bool set = IA2_ADDR(fn);
```
is probably rare so it should be low priority. Also we should support rewriting
```
for (init; fn_ptr; expr)
// for (init; IA2_ADDR(fn_ptr); expr)
```
since that did appear in `ngx_rtmp_eval.c`.

ayrtonm:
[...]

PR #200, opened 2023-04-06
Add FnPtrEq pass for pointers cast to bool or compared for equality

ayrtonm:
This rewriter pass is an initial attempt at issue #198. It only handles the first two cases in the code in the issue. I added a new test for it and removed [this hack](https://github.com/immunant/IA2-Phase2/commit/40240e7b2d9ea1603cc628af69dec916c521e21e) that was used to get the simple1 test to build.

Also using FileCheck when rewritting a file requires using `sed` because otherwise the FileCheck annotations will always match themselves in the output. This needs to be fixed for the simple1 test, but we should try to factor this out somehow because it's easy to forget and to avoid duplication.


ayrtonm:
Passing `--match-full-lines` to `FileCheck` avoids the issue where the FileCheck annotations match themselves and is a much cleaner solution than using `sed` to remove annotations.

ayrtonm:
Added a few new test cases based on the most commonly occurring cases in the particular nginx/rtmp build config I'm using. AFAICT there are only 4 places I had to rewrite in the rtmp module that don't match any of the test cases so I'll handle these cases and get this merged then do the rest later.

PR #206, opened 2023-04-07
tests/threads: re-enable test

fw-immunant:
This test just needed its lit invocations updated. Could someone double-check that I've done it right?


ayrtonm:
LGTM

PR #211, opened 2023-04-07
Compartmentalize Nginx and its rtmp module

ayrtonm:
Build instructions:
```
cd external/nginx
./reconfigure
make
```

to run with permissive mode, add `#include <permissive_mode.h>` to src/core/nginx.c (or any one source file in this executable) and rerun `make`. This should produce multiple `mpk_log_$PID` files.


ayrtonm:
I fixed the ld invocations to use the call gate shared lib, padding scripts, the padded libc and ld args files and was able to build nginx and the module. I also built it with partition alloc but that segfaults because we need a wrapper for posix memalign. Running without the allocator gives the same behavior as the previous PR where the parent process faults after main but the child processes keep running (not surprising since it's not confused to do much yet).

Next step is to cherry pick the assert pkru commit to see if we have any missing wrappers and under what conditions/configs we'll need the signal handler.

ayrtonm:
We merged the assert pkru PR (#212) and it hasn't caught any unexpected caller PKRUs so it seems the build system changes are fine and we aren't missing callgates. I also rebased the sighandler stuff (#227) locally on this PR, defined the two signal handlers that nginx needs and added calls to `sigaltstack` for the main thread (from the permissive mode ctor) and from `ia2_thread_begin` for the other threads. With the config I'm testing (based off the webcam demo in the [rtmp README](https://github.com/arut/nginx-rtmp-module)) this should be enough to run nginx in permissive mode, but I'm running into a `SIGSEGV` with `SI_KERNEL` (forced by linux) which kills the worker process and its parent. Aside from this, permissive mode seems to be working correctly and I fixed a couple of memory access errors for which `dladdr` was able to provide the variable names. I still get 807 memory access errors w/o variable names, but based off the addresses it's mostly the same set of variables. I also tried opening and reading from the `.symtab`s to get names that `dladdr` doesn't provide but that didn't help much so it might be worth adding support for reading dwarf info eventually.

`strace -ff` creates 4 log files (one per process) with the calls to `sigaltstack` as expected and a `SIGSEGV` followed by `SIGTRAP` for each cross-compartment violation we handle. The first log shows a clean exit which is expected, the second (master process) gets a `SIGCHLD` then terminates, the third (worker process) terminates as shown below and the fourth handles some compartment violations then execve's ffmpeg as expected so it's irrelevant after that.

```
$ strace -ff -o logs ./build/nginx
...
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_PKUERR, si_addr=0x55b754aeb618, si_pkey=1} ---
rt_sigreturn({mask=[]})                 = 94245888112144
--- SIGTRAP {si_signo=SIGTRAP, si_code=TRAP_TRACE, si_addr=0x55b754a1cc88} ---
rt_sigreturn({mask=[]})                 = 94245888112144
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_PKUERR, si_addr=0x55b754b7bbfc, si_pkey=1} ---
rt_sigreturn({mask=[]})                 = 94245888703484
--- SIGTRAP {si_signo=SIGTRAP, si_code=TRAP_TRACE, si_addr=0x55b754a18293} ---
rt_sigreturn({mask=[]})                 = 94245888703484
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_PKUERR, si_addr=0x7fc41b810460, si_pkey=2} ---
--- SIGSEGV {si_signo=SIGSEGV, si_code=SI_KERNEL, si_addr=NULL} ---
+++ killed by SIGSEGV (core dumped) +++
```

Since the `SI_KERNEL` happen after the `SEGV_PKUERR` but before its `rt_sigreturn` it seems that we're getting an issue in the permissive mode signal handler as explained in [this comment](https://elixir.bootlin.com/linux/latest/source/kernel/signal.c#L1691). I also tried running w/o the permissive mode logging thread in case opening/writing to one of the logging files was causing issues and to make permissive mode minimally invasive. From skimming linux's signal handling code I think the issue is caused by one of the accesses to the signal handler stack frame, but I need to try to get more info from the sighandler to figure out exactly which access is causing this.

ayrtonm:
TODO: there's some variadic functions in the ld wrap file so we need to have the rewriter omit these. I made the change locally but need to double check that the other rewriter passes don't need changes.

rinon:
I've narrowed the `SI_KERNEL` segfault down to the signal handler frame being allocated on a stack that it doesn't currently have permission to access. Not sure _why_ the stack is wrong, but that's what seems to be happening.

rinon:
The crash observed currently is caused by the following scenario:

Compartment 1 = nginx, 2 = rtmp module

C2 `ngx_rtmp_cmd_connect_init` calls `ngx_log_error_core` which is variadic so doesn't get a wrapper emitted (yet).
`ngx_log_error_core` is now running in C2, but expects to be C1. It calls an indirect function (`log->handler`) which transitions from C1 -> C0 then calls the function, but this saves the C2 stack to the C1 stack save slot. When the handler returns, we transition back to C1 but restore the C2 stack and can't access it, causing a double fault.

If I remove the indirect call to `log->handler` in `ngx_log_error_core` everything runs under the permissive mode signal handler.

rinon:
This should still land on top of #227, but I think it's working now (with the permissive mode handler).

I've re-written the `reconfigure` script and build system changes some to put _all_ modifications under the build directory. Sources are first copied to `build/nginx/` before rewriting so that we keep a clean source tree. It was too much of a pain to clean the checked in sources between builds.

TLDR new build instructions:

    cd external/nginx
    ./reconfigure
    (cd build && make)

endbr64 (approved):
1) It looks like we're keeping the `.patch` files around to fix the rewriter output. Is the intent to continue rewriter work on `main`?
1) Likewise, the signal stack stuff in `libia2/threads.c` is commented out.
1) When I compartmentalize the Perl module on top of this, should I try to run RTMP and Perl together, or just do Perl as a separate config?

rinon:
> 1. It looks like we're keeping the `.patch` files around to fix the rewriter output. Is the intent to continue rewriter work on `main`?

Yep, exactly.

>     2. Likewise, the signal stack stuff in `libia2/threads.c` is commented out.

Yeah, #249 

>     3. When I compartmentalize the Perl module on top of this, should I try to run RTMP and Perl together, or just do Perl as a separate config?

Good question. What's your opinion on that? I don't know much about the perl module itself. If we can put it in compartment 3 and RTMP in compartment 2, that could be an interesting demo. If that adds a lot of complexity then we can make them mutually exclusive.

endbr64:
I'm trying to build it with `-DPKEY=3` at the moment.

PR #212, opened 2023-04-10
Assert that caller PKRU matches expected value

### LINKED ISSUE #147: Assert PKRU matches expected value in wrappers
To aid debugging we should read the PKRU's value with `rdpkru` and assert that its value matches what we expect at the start of each wrapper. The best place to do this would be right after [pushing callee-saved registers (`preserved_registers`)](https://github.com/immunant/IA2-Phase2/blob/main/header-rewriter/GenCallAsm.cpp#L380) on the caller's stack. Since all call gates have statically-known caller/target PKRU values, this isn't strictly necessary so we could gate it behind a feature flag (must be mutually exclusive with `LIBIA2_INSECURE` since that omits the `wrpkru`s).

This would essentially automate the debugging process that @endbr64 and I went through with nginx. It would also make it more obvious that what seemed like a stack manipulation issue was actually an unwrapped cross-compartment call. The call gates are in assembly though (there's no reliable way to use C since we switch stacks), so this is straightforward but requires some care to avoid introducing new bugs.

We could also assert that the PKRU matches the expected value after calling the wrapped function, but that would be more of a slight security enhancement rather than to aid debugging.

ayrtonm:
This should make debugging nginx easier when we get it building again in #211. We might want to gate this with a macro other than `LIBIA2_INSECURE` at some point since this is not necessary if a program is compartmentalized correctly. Also tests using mpk are expected to fail until #205 gets merged. Closes #147.


ayrtonm:
TODO: stringify the macro parameter in `ASSERT_PKRU`

ayrtonm:
Canceled checks since self-hosted runner is down. I fixed the issues mentioned in the last comment so tests should pass, but I still need to deduplicate stuff before merging this.

ayrtonm:
The latest set of commits adds a `LIBIA2_DEBUG` flag to enable asserting the caller's pkru in all callgates. I also added this to CI but only for the pkru-enabled host since the flag is mutually exclusive with `LIBIA2_INSECURE`. If we really need to support both flags, we could keep track of the compartment in a variable and check that in insecure mode but that's more trouble than it's worth since `LIBIA2_INSECURE` is mostly intended to allow testing our instrumentation on non-pkru-enabled hosts.

PR #218, opened 2023-04-18
Add support for rewriting basic macros

ayrtonm:
This adds support for rewriting object-like macros by rewriting them at the expansion site. Rewriting at the expansion site is required because different annotations may need to be added depending on the context it's used in e.g.
```
#define SOME_PTR fn_ptr

SOME_PTR(); // rewritten as `IA2_CALL(SOME_PTR, ID)();
if (SOME_PTR == another_ptr) // rewritten as if (IA2_ADDR(SOME_PTR) == IA2_ADDR(another_ptr))
```

Function-like macros are currently not rewritten because they're typically more complex than the simple aliases which object-like macros are typically used for. In these cases the behavior is the same as before where the site that needs manual changes is logged to stdout.


ayrtonm:
Tests were previously failing because LIT was looking for the wrong ld args file, but I should add a more complex macro to the new test so we can test that it won't get rewritten.

ayrtonm:
Tests are failing for the right reason now.

Issue #219, opened 2023-04-18
Support nested rewriter changes


ayrtonm:
Some rewriter passes (like [IA2_ADDR](https://github.com/immunant/IA2-Phase2/blob/main/header-rewriter/SourceRewriter.cpp#L610) and [IA2_CALL](https://github.com/immunant/IA2-Phase2/blob/main/header-rewriter/SourceRewriter.cpp#L404)) replace entire expressions with `"IA2_MACRO(" + old_expression + ")"`, rather than inserting `"IA2_MACRO("` and `")"` before/after the expression. Since the new expression uses the text from the old expression any other changes to the new expression may be ignored. For example in
```
typedef int(*op)(int);
op identity(op fn) { return fn; }
int dec(int x);

if (identity(dec)) { }
```

the last line should be rewritten as `if (IA2_ADDR(identity(IA2_FN(dec))))` because
 - identity returns an opaque struct which needs `IA2_ADDR` to be compared
 - identity takes an opaque struct so we need `IA2_FN` to cast `dec`

However it's incorrectly rewritten as `if (IA2_ADDR(identity(dec)))`, because the pass adding `IA2_ADDR` uses the original text. This triggers a compiler error but the `IA2_FN` can be added manually.

I originally used the old expression in a few passes to avoid some technical rewriter issues with how/where the closing parenthesis was being inserted, but we should check which passes this issue may apply to and try to fix them.

fw-immunant:
Note that replacements are [considered](https://clang.llvm.org/doxygen/classclang_1_1tooling_1_1Replacements.html#details) to conflict if they start/end at the same location (so their order would matter):
> Two replacements are considered conflicts if they overlap or have the same offset (i.e. order-dependent).

This doesn't matter if there are guaranteed to be tokens separating the outer replacement from the inner one (e.g. "<ins>IA2_ADDR( </ins> identity( <ins>IA2_FN( </ins> dec <ins> )</ins> )<ins> )</ins>"), but does if we want to wrap one expression in multiple passes.

PR #220, opened 2023-04-18
Add support for nested rewriter changes

ayrtonm:
This just adds a failing test case from #219, but I'm not actively working on this


ayrtonm:
Not sure this is relevant anymore but I'll leave the issue open to revisit later

PR #221, opened 2023-04-24
Remove PKEY argument from IA2_CALL macro

ayrtonm:
Since all translation units must be compiled with `-DPKEY=N` we can use this macro to avoid specifying the pkey argument in the IA2_CALL macro. This commit also updates tests with `IA2_CALL` and enables the rewriter checks in the read_config and trusted_indirect tests.


ayrtonm:
Canceled checks since self-hosted runner is down.

Issue #208, opened 2023-04-07
Enable repeated rewriter runs


ayrtonm:
We should make sure that the rewriter can be run repeatedly on a codebase without making changes each time. This would be useful to make sure the developer workflow with compartmentalized code isn't drastically different, but there are lots of open questions and issues.

# Rewriter changes
- when any of the rewriter passes change code will they match the new code in the next run? do different rewriter passes interact in this way?
-  do we update generated build artifacts in-place or overwrite them? can we minimize the diff if we overwrite check-in build artifacts
- need to check for [existing `used` attributes on static functions](https://github.com/immunant/IA2-Phase2/blob/main/header-rewriter/SourceRewriter.cpp#L494)
- make sure that adding a new function pointer type does not use an existing [function pointer ID](https://github.com/immunant/IA2-Phase2/blob/main/header-rewriter/SourceRewriter.cpp#L364-L370). Alternatively (and ideally), we'd get rid of function pointer IDs even if we keep the mangled types they correspond to
# Testing
This will be slightly cumbersome to test with cmake because the sources are compiled differently before the first rewriter run and afterwards (e.g. the rewriter generates a header which we need to `-include` to compile the code after the first run). We could either have separate targets for pre/post-rewriter sources or create an empty file to `-include` the generated header on the first run. The former seems more sane, but it would require two compile_commands.json because each source file would have two entries which libtooling does not really support well. Given the hacks required to get cmake to [generate a compile commands](https://github.com/immunant/IA2-Phase2/blob/main/cmake/define-test.cmake#L201-L205) we can use, I'm not sure how feasible it is to generate separate compile commands in cmake.

ayrtonm:
Aside from ensuring that rewriter changes aren't duplicated, I also noticed that a second rewriter run doesn't pick up function pointer types that have already been turned into opaque structs. This means that the newly generated header will have missing types.

ayrtonm:
Here are the parts of the rewriter that either rewrite or generate code. We should consider if they'll need changes and check them off as we validate that they work as expected with repeated runs.
- FnPtrTypes pass
    - [ ] function pointer types -> opaque structs
        - opaque structs are named based on mangled types not the IDs so repeated runs should be fine
- FnPtrNull
    - [ ] NULL function pointer expressions -> zeroed opaque structs
- FnPtrCall
    - [ ] generates IDs as it finds indirect callsite function signatures
        - we need to check what IDs were assigned in previous runs before we assign new ones
[...]

PR #222, opened 2023-04-24
Remove mangled type ID from IA2_CALL macro

ayrtonm:
Marked as WIP since I'm not sure if we want to merge this yet and because this doesn't completely remove the mangled type IDs. Ideally we'd remove them completely and leave only the mangled types in symbol names (i.e. won't be visible to the user) to simplify #208.


ayrtonm:
We decided that compiler builtins are ok.
TODO: clean this and merge it

PR #223, opened 2023-04-24
Add permissive mode header

ayrtonm:
Permissive mode installs a signal handler that switches the PKRU for a single instruction on segfaults then switches it back after redoing the memory access. To use it just `#include` the new header. See the header documentation for details.


ayrtonm:
TODO: clean this up and merge with async-signal-unsafe functions and disclaimer that this is for debugging/experimental

PR #227, opened 2023-05-09
Add basic signal handler support

ayrtonm:


libia2/threads.c
```diff
@@ -20,6 +21,7 @@ void *ia2_thread_begin(void *arg) {
 
   protect_tls();
   init_stacks();
+  sigaltstack(&ia2_signal_stack[STACK_SIZE - 8], NULL);
```
ayrtonm, line 24: Note this is wrong. `sigaltstack` does not take the pointer directly.

libia2/threads.c
```diff
@@ -11,15 +12,23 @@ struct ia2_thread_thunk {
...
+  /*     .ss_sp = ia2_signal_stack, .ss_flags = 0, .ss_size = STACK_SIZE}; */
+
   /* Free the thunk. */
   munmap(arg, sizeof(struct ia2_thread_thunk));
 
   protect_tls();
   init_stacks();
+  /* TODO: Set up alternate stack when we have per-thread shared compartment
+   * data. */
+  /*  sigaltstack(&alt_stack, NULL); */
```
fw-immunant, line 31: This signal stack stuff is discussed [here](https://github.com/immunant/IA2-Phase2/issues/249#issuecomment-1652696362).

For now we don't have to do anything with it.

rewriter/SourceRewriter.cpp
```diff
@@ -494,6 +494,13 @@ class FnPtrExpr : public RefactoringCallback {
     assert(fn_decl != nullptr);
 
     Function fn_name = fn_decl->getName().str();
+
+    // Unlike the macro expansion check below, this check should go before
+    // modifying the maps in this pass because we don't want to treat the signal
+    // handlers as functions that have their address-taken.
```
fw-immunant, line 500: Why not?

libia2/include/ia2.h
```diff
@@ -44,6 +44,41 @@ static uint32_t ia2_get_pkru() {
...
+#ifdef LIBIA2_INSECURE
+#define _IA2_DEFINE_SIGNAL_HANDLER(function, pkey)    \
+    __asm__(".global ia2_sighandler_" #function "\n"  \
+            "ia2_sighandler_" #function ":\n"         \
+            "jmp " #function "\n")
+#else
+#define _IA2_DEFINE_SIGNAL_HANDLER(function, pkey)    \
+    __asm__(".global ia2_sighandler_" #function "\n"  \
+            "ia2_sighandler_" #function ":\n"         \
+            "movq %rcx, %r10\n"                       \
```
fw-immunant, line 58: Not a blocker, but reading this reminded me that we don't have `endbr64` instructions here (or likely a lot of places) for CFI...

fw-immunant (approved):
A few comments inline, but no blockers.

libia2/include/ia2.h
fw-immunant: Actually I have no idea if signal handler calls are considered indirect branches for the purposes of CFI.

rinon: I don't _think_ so? But we'll find out, that can all be handled at once.

PR #229, opened 2023-05-26
Fix init TLS in new threads

ayrtonm:
Calling `init_tls_N` when starting new threads may trigger compartment violations because the calls for each compartment do not have call gates so all `init_tls_N` calls happen in the same compartment as the `pthread_create` call. This was not a problem in the existing threads test when calling `pthread_create` from the main compartment because the library was in the default compartment so it's memory was not `pkey_mprotect`ed. I'm still not sure why the `pthread_create` from the library was not an issue though, but I believe it may related to the main compartment being an executable rather than a .so (see [this comment](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/dl-sym.c;h=b1ed1f2006cd3bd23f83da2ea3f2449d1c8b0fea;hb=HEAD#l99)) since the issue seems to happen in the call to `dlsym` in `protect_tls_pages`.

This PR fixes this by adding call gates for the calls to `init_tls_N` in `ia2_thread_begin`. Note that this keeps the calls to `init_tls_N` in `INIT_RUNTIME` w/o call gates since the runtime hasn't been fully initialized by then. Also this could use some cleanup -- I think we want to avoid changing the rewriter for this and instead wrap `init_tls_N` the same way we wrap `pthread_create` and `main` and replace the calls in `INIT_RUNTIME` with `__real_init_tls_N`.


ayrtonm:
@fw-immunant  manually running this in gdb shows the same backtrace I initially ran into.
```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
main-module main pkru=fffffff0
library pkru ffffffcc
main-module main pkru=fffffff0
[New Thread 0x7ffff65ff000 (LWP 9770)]
library pkru ffffffcc
[New Thread 0x7ffff5dfe000 (LWP 9771)]
[Thread 0x7ffff65ff000 (LWP 9770) exited]

Thread 3 "protected_threa" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff5dfe000 (LWP 9771)]
0x00007ffff7fd3636 in ?? () from /lib64/ld-linux-x86-64.so.2
(gdb) bt
#0  0x00007ffff7fd3636 in ?? () from /lib64/ld-linux-x86-64.so.2
#1  0x00007ffff7fd42a8 in ?? () from /lib64/ld-linux-x86-64.so.2
#2  0x00007ffff7769400 in ?? () from /home/ayrton/ia2/IA2-Phase2/build/rewriter/tests/protected_threads/libc.so.6
#3  0x00007ffff769ad88 in ?? () from /home/ayrton/ia2/IA2-Phase2/build/rewriter/tests/protected_threads/libc.so.6
#4  0x00007ffff7fcb4e1 in _dl_catch_exception () from /lib64/ld-linux-x86-64.so.2
#5  0x00007ffff7fcb603 in ?? () from /lib64/ld-linux-x86-64.so.2
#6  0x00007ffff769a75f in ?? () from /home/ayrton/ia2/IA2-Phase2/build/rewriter/tests/protected_threads/libc.so.6
#7  0x00007ffff769ae15 in dlsym () from /home/ayrton/ia2/IA2-Phase2/build/rewriter/tests/protected_threads/libc.so.6
#8  0x0000555555557195 in protect_tls_pages ()
#9  0x00007ffff7768dfd in dl_iterate_phdr () from /home/ayrton/ia2/IA2-Phase2/build/rewriter/tests/protected_threads/libc.so.6
#10 0x00007ffff7e87428 in init_tls_2 () at /home/ayrton/ia2/IA2-Phase2/build/rewriter/tests/protected_threads/library.c:9
#11 0x00005555555567b9 in protect_tls () at /home/ayrton/ia2/IA2-Phase2/build/rewriter/tests/protected_threads/main.c:15
#12 0x0000555555557b59 in ia2_thread_begin ()
#13 0x00007ffff769ebb5 in ?? () from /home/ayrton/ia2/IA2-Phase2/build/rewriter/tests/protected_threads/libc.so.6
#14 0x00007ffff7720d90 in ?? () from /home/ayrton/ia2/IA2-Phase2/build/rewriter/tests/protected_threads/libc.so.6
```

rinon:
Can (should) we land this? @fw-immunant it looks like Ayrton wanted you to take a look at this?

fw-immunant:
> Can (should) we land this? @fw-immunant it looks like Ayrton wanted you to take a look at this?

We should land this in some form. I was vaguely aware of it but didn't do a full read through until now. I'll submit comments as a review.

libia2/include/ia2.h
```diff
@@ -72,8 +72,10 @@ asm(".macro mov_pkru_eax pkey\n"
 
 #ifdef LIBIA2_INSECURE
 #define protect_tls_for_compartment(n)
+#define late_protect_tls_for_compartment(n)
```
fw-immunant, line 75: I think the `late_` name leaves something to be desired--this is called later than `protect_tls_for_compartment(n)` but specifically it's called on new thread creation, not just any time later. `late_` doesn't specify an invariant or explain specifically what is done differently (i.e., the name doesn't convey that the difference is that `late_init_tls_N` is wrapped in call-gates).

libia2/threads.c
```diff
@@ -18,8 +18,8 @@ void *ia2_thread_begin(void *arg) {
   /* Free the thunk. */
   munmap(arg, sizeof(struct ia2_thread_thunk));
 
-  protect_tls();
   init_stacks();
+  late_protect_tls();
```
fw-immunant, line 22: We should document the ordering requirements here.

Specifically, `late_protect_tls` comes after `init_stacks` because it wants to switch to those stacks as part of call-gates.

However, we have the problem that the previous ordering was also necessary: `init_stacks` writes initial stack pointers to the per-compartment `ia2_stackptr_N` TLS variables. If these haven't been `pkey_mprotected`, then any rogue compartment can overwrite them between when we set up stacks and when we actually try to use them.

`init_stacks` and the `ALLOCATE_COMPARTMENT_STACK` macro that it uses are specifically written to not require stacks to be set up yet when they run, and the macro changes pkeys and allocate per-compartment stacks without needing a call-stack.

Looking at `late_protect_tls`, it simply calls the right set of `late_init_tls_N()` functions, which does `dl_iterate_phdr(protect_tls_pages, &args);`. `protect_tls_pages` does `dlsym`, some logic, and then `pkey_mprotect`. The only part of this that should actually be compartment-dependent is `dlsym`, and that's what the backtrace shows. But what I understand here is that we need to call `dlsym` from the compartment for libdl, not the compartment whose TLS pages we're protecting?

Is that what we're doing by wrapping all the `late_init_tls_N` functions in call-gates? Why not just wrap `late_protect_tls`?

fw-immunant (changes requested):
Some comments inline. I think the ordering issues around stack pointers, stacks, and TLS are a problem.

libia2/threads.c
fw-immunant: I talked about this with @rinon. Basically, [`late_`]`protect_tls` needs to run each `late_protect_tls_for_compartment` with the appropriate pkey so that `dl_iterate_phdr` can access the stack. That stack switching has to happen before we set up `ia2_stackptr_N` (because they live in TLS), so we need to allocate and switch to a stack for each compartment in order to be able to run `late_protect_tls_for_compartment` for it. This sounds like a dependency loop, but it isn't because we don't actually need `ia2_stackptr_N` to be set up until we real call-gate, and this sequence can just be hard-coded control flow without call gates. We can even save the stack we used to call `late_protect_tls_for_compartment` in `ia2_stackptr_N` after `late_protect_tls_for_compartment` ends.

I'll see if I can implement this atop this PR.

ayrtonm: > But what I understand here is that we need to call dlsym from the compartment for libdl

Did we decide we need to put libdl in its own compartment? Would this be optional or required?

> I'll see if I can implement this atop this PR.
@fw-immunant if you're making changes on top of this, feel free to push to this branch/PR if it'd make review easier

fw-immunant: I don't think we necessarily have to put libdl in its own compartment. I think we can entirely avoid libdl in `protect_tls_pages` if we either pass in the address of `ia2_stackptr_0` or hard-code `&ia2_stackptr_0` as the page to avoid in protect_tls_pages. I've done the former but I'm not completely sure about the security of that approach--is there a way to prevent this being a useful gadget? See `PhdrSearchArgsTls`--cc @rinon

rinon: Can't you just use `ia2_stackptr_0` as a PC-relative reference inside of `protect_tls_pages`? That should be safe, as you can only ever refer to that label, not an arbitrary address.

fw-immunant: We can do that. I was curious whether the approach we were taking in #248 would also help us with TLS setup (and allow us to not hard-code `ia2_stackptr_0` in `protect_tls_pages`, possibly helping with shared TLS later on) but I'm not sure that it would. I also think we can probably also eliminate the `pkey` field from `struct PhdrSearchArgs` as of that PR.

fw-immunant (approved):
I fixed the things I brought up in my review; now someone else should review my recent changes. In particular I wonder if there's a better option than blind hope for keeping r12 unspilled across our calls to `init_tls_N` (or a better option than keeping pkru in r12d to restore it after setting up stacks and TLS for all compartments--for the main thread we could reset to compartment 0 but threads need to return to the compartment that they started in).

libia2/include/ia2.h
```diff
@@ -77,38 +77,42 @@ static uint32_t ia2_get_pkru() {
...
-#define _IA2_DEFINE_SIGNAL_HANDLER(function, pkey)    \
-    __asm__(".global ia2_sighandler_" #function "\n"  \
-            "ia2_sighandler_" #function ":\n"         \
-            "jmp " #function "\n")
+#define _IA2_DEFINE_SIGNAL_HANDLER(function, pkey)                             \
+  __asm__(".global ia2_sighandler_" #function                                  \
+          "\n"                                                                 \
+          "ia2_sighandler_" #function                                          \
+          ":\n"                                                                \
+          "jmp " #function "\n")
```
rinon, line 85: Can we undo the formatting change here please? It's hard to see what changed.

libia2/include/ia2.h
```diff
@@ -77,38 +77,42 @@ static uint32_t ia2_get_pkru() {
...
+#define IA2_DEFINE_SIGACTION(function, pkey)                                   \
+  void ia2_sighandler_##function(int, siginfo_t *, void *);                    \
+  _IA2_DEFINE_SIGNAL_HANDLER(function, pkey)
 
-#define IA2_DEFINE_SIGHANDLER(function, pkey)  \
-    void ia2_sighandler_##function(int);       \
-    _IA2_DEFINE_SIGNAL_HANDLER(function, pkey)
+#define IA2_DEFINE_SIGHANDLER(function, pkey)                                  \
+  void ia2_sighandler_##function(int);                                         \
+  _IA2_DEFINE_SIGNAL_HANDLER(function, pkey)
```
rinon, line 112: Unnecessary formatting changes?

libia2/include/ia2.h
```diff
@@ -210,23 +208,41 @@ static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
         "xor %%ecx,%%ecx\n"                                                    \
         "# eax = old pkru; also zeroes edx, which is required for wrpkru\n"    \
         IA2_RDPKRU "\n"                                                        \
-        "# save pkru in r12d\n"                                                \
+        "# save pkru in r12d. XXX: if a callee here spills r12, it could\n"    \
+        "# be corrupted by another thread subsequently corrupt the pkru\n"     \
+        "# when we restore it from r12d\n"                                     \
```
rinon, line 213: Can we just assert the initial pkey value is compartment 0 and restore back to that value when finished? I don't like the idea of restoring from something that will likely be spilled.

rewriter/SourceRewriter.cpp
```diff
@@ -726,8 +726,10 @@ class FnDecl : public RefactoringCallback {
     assert(result.SourceManager != nullptr);
     auto &sm = *result.SourceManager;
 
+    Function fn_name = fn_node->getNameAsString();
+
     // Ignore declarations in libc and libia2 headers
-    if (ignore_file(get_filename(fn_node->getLocation(), sm))) {
+    if (ignore_file(get_filename(fn_node->getLocation(), sm)) && !fn_name.starts_with("late_init_tls_")) {
```
rinon, line 732: Is this leftover? I don't see this symbol being defined in this PR

libia2/include/ia2.h
rinon: Or put that in a separate change on the bottom of the stack, if you really want to do it. But I prefer the old formatting.

libia2/ia2.c
```diff
@@ -64,7 +64,7 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
     exit(-1);
   }
 
-  struct PhdrSearchArgs *search_args = (struct PhdrSearchArgs *)data;
+  struct PhdrSearchArgsTls *search_args = (struct PhdrSearchArgsTls *)data;
```
rinon: Is this change even relevant any longer? I don't see any of this commit in the final diff.

rinon (approved):
Small comments mostly. This seems alright to me.

libia2/include/ia2.h
fw-immunant: This is the formatting clang-format gives us, but it is pretty awful. For other macros containing asm I've been wrapping them in a `/* clang-format off */` block, so I'll do that here.

libia2/include/ia2.h
fw-immunant: This is also up to clang-format and arguably more consistent with the rest of the file. I can move the formatting to the bottom of the stack or a separate PR.

rewriter/SourceRewriter.cpp
fw-immunant: It is, I'll remove this and squash some other stuff to simplify the history.

libia2/ia2.c
fw-immunant: Will rebase away.

libia2/include/ia2.h
fw-immunant: We can't--in new threads, we need to restore the pkey that the thread started with, which is whatever the creating compartment had, not necessarily zero. I don't think this is ultimately the right way to do this either (given that we don't control register allocation of all C code that we want to call here), but I'm not sure what is.

fw-immunant: Filed as #257.

libia2/include/ia2.h
```diff
@@ -142,11 +142,28 @@ asm(".macro mov_pkru_eax pkey\n"
...
+    };                                                                         \
+    dl_iterate_phdr(protect_pages, &args);                                     \
+  }                                                                            \
+  void init_tls_##n(void) {                                                    \
+    struct PhdrSearchArgs args = {                                             \
+        .pkey = n,                                                             \
+        .address = &init_pkey_##n##_ctor,                                      \
+    };                                                                         \
+    dl_iterate_phdr(protect_tls_pages, &args);                                 \
+  }
```
rinon, line 166: I missed that this got restored in a rebase, this is now in `ia2_compartment_init.inc`

fw-immunant: Good call, thanks for catching it.

PR #232, opened 2023-06-28
rewriter: tests: run read_config in its test

fw-immunant:
I'm curious if this works in CI.

### DIFF
--- a/rewriter/tests/read_config/CMakeLists.txt
+++ b/rewriter/tests/read_config/CMakeLists.txt
@@ -1,3 +1,10 @@
+if(LIBIA2_INSECURE)
+    set(STDOUT_REF ${CMAKE_CURRENT_SOURCE_DIR}/Output/read_config.insecure.out)
+else()
+    set(STDOUT_REF ${CMAKE_CURRENT_SOURCE_DIR}/Output/read_config.out)
+endif()
+configure_file(${STDOUT_REF} ${CMAKE_CURRENT_BINARY_DIR}/read_config.out)
+
 define_shared_lib(
     SRCS plugin.c
     PKEY 2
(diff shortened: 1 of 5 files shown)


ayrtonm:
I believe this isn't expected to work because of #112 (destructors inserted by the compiler), but I don't remember the exact details off-hand.

fw-immunant:
The backtrace I get from read_config is:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff757efc2 in ?? ()
   from rewriter/tests/read_config/libc.so.6
(gdb) bt
#0  0x00007ffff757efc2 in ?? ()
   from rewriter/tests/read_config/libc.so.6
#1  0x00007ffff7de46b4 in get_opt (name=0x37a400204000 "name")
    at rewriter/tests/read_config/plugin.c:74
#2  0x00007ffff7dec448 in __wrap_get_opt ()
   from rewriter/tests/read_config/libread_config_call_gates.so
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb)
```

This is not the crash at exit that we expect based on #112. Looking at the context that for some reason the backtrace does not provide (it turns out this first call to `get_opt` gets a PKU violation), it looks like we're crashing at line 93 in main.c:

```c
      // This passes a pointer to the heap to the other compartment which is
      // fine since we're pretending the heap is intentionally shared for now
      opt = get_opt(tok);
```

Is this comment outdated now that we have PartitionAlloc? cc @endbr64

fw-immunant:
It turns out we do need to share these heap-allocated strings, and after doing so, we can ignore the issues caused by #112 in `read_config` the same way we do in `recursion`. This is ready to merge after review.

endbr64 (approved):

PR #239, opened 2023-07-21
Fix permissive mode to restore PKRU

### LINKED ISSUE #238: Permissive mode signal handler does not restore PKRU state after trapping
The permissive mode signal handler resets the PKRU register to 0 (all permitted) after a SIGSEGV PKRU fault, then is supposed to restore the PKRU register to the prior value after the faulting instruction executes. The restore isn't actually happening (at least not on my Ryzen 7).

The issue is that the kernel does not restore the PKRU from the extended fpstate unless the xsave header bit 9 is set to indicate that the PKRU state was saved in the xsave buffer. This bit is NOT set when the PKRU register is set to zero (its initial state) as an optimization to speed up the xsave operation. We need to explicitly set this bit of the feature mask if we want to change the value from zero to non-zero in the signal handler.

See the following kernel code where the PKRU state is restored from the signal context buffer:
https://github.com/torvalds/linux/blob/bfa3037d828050896ae52f6467b6ca2489ae6fb1/arch/x86/kernel/fpu/xstate.c#L1279-L1291

rinon:
Permissive mode had a few issues, most importantly:
- PKRU offset cpuid call was using the wrong register
- PKRU was not modified properly if it was 0 before the signal handler

The kernel only restores the PKRU value from the XSAVE area after the
signal handler if the PKRU bit in the XSAVE header is set (See Intel
manual Chapter 13.4 and linux kernel xstate.c). We want to change the
PKRU value after the permissive mode handler runs, so we need to make
sure this header bit is unconditionally set. If the PKRU value was 0,
the kernel will not set the bit before the signal handler runs.

Fixes #238

### DIFF
--- a/libia2/include/ia2.h
+++ b/libia2/include/ia2.h
@@ -36,6 +36,13 @@
 
 #endif
 
+/// Helper to get the PKRU register value
+static uint32_t ia2_get_pkru() {
+  uint32_t pkru;
+  __asm__ volatile("rdpkru" : "=a"(pkru) : "a"(0), "d"(0), "c"(0));
+  return pkru;
+}
+
 /// Protect pages in the given shared object
 ///
 /// \param info dynamic linker information for the current object
--- a/libia2/include/permissive_mode.h
+++ b/libia2/include/permissive_mode.h
@@ -7,6 +7,19 @@
 #include <stdio.h>
 
 /*
+ * PKRU is defined to be bit 9 in the XSAVE state component bitmap (Intel SDM
+ * Section 13.1).
+ */
+#define XFEATURE_PKRU 9
+#define XFEATURE_MASK_PKRU		(1 << XFEATURE_PKRU)
+
+/*
+ * Offset of the XSAVE header in bytes from the XSAVE area base (Intel SDM
+ * Section 13.4.2)
+ */
+#define XSAVE_HEADER_OFFSET 512
+
+/*
  * This header implements an signal handler to allow running compartmentalized
  * programs in a "permissive mode" which logs cross-compartment memory accesses
  * instead of terminating the process. To use it, just #include this header from
@@ -135,6 +148,12 @@ void permissive_mode_handler(int sig, siginfo_t *info, void *ctxt) {
   int offset = pkru_offset();
   char *fpregs = (char *)uctxt->uc_mcontext.fpregs;
   uint32_t *pkru = (uint32_t *)(&fpregs[offset]);
+
+  // Set bit 9 (PKRU) in the xfeatures so that the PKRU register always gets
+  // restored from the XSAVE state after the signal handler. See
+  // copy_uabi_to_xstate() in the kernel xstate handling.
+  uint64_t *xsave_header = (uint64_t *)(&fpregs[XSAVE_HEADER_OFFSET]);
+  *xsave_header |= XFEATURE_MASK_PKRU;
 
   const uint64_t trap_bit = 0x100;
   static __thread uint32_t old_pkru = 0;
--- a/libia2/include/permissive_mode.h
+++ b/libia2/include/permissive_mode.h
@@ -1,10 +1,15 @@
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE // for SIG* macro
 #endif
+#include <assert.h>
+#include <dlfcn.h>
 #include <pthread.h>
 #include <signal.h>
+#include <stdbool.h>
 #include <stddef.h>
+#include <stdint.h>
 #include <stdio.h>
+#include <string.h>
 
 /*
  * PKRU is defined to be bit 9 in the XSAVE state component bitmap (Intel SDM
--- a/libia2/include/permissive_mode.h
+++ b/libia2/include/permissive_mode.h
@@ -196,6 +196,8 @@ static char log_name[24] IA2_SHARED_DATA = {0};
 
 // The main function in the logging thread
 void *log_mpk_violations(void *arg) {
+  /* Explicitly enter compartment 1, because this function isn't wrapped. */
+  __asm__("wrpkru\n" : : "a"(0xFFFFFFF0), "d"(0), "c"(0));
   snprintf(log_name, sizeof(log_name), "mpk_log_%d", getpid());
   FILE *log = fopen(log_name, "w");
   assert(log);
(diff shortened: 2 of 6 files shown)


libia2/include/ia2.h
```diff
@@ -36,6 +36,12 @@
 
 #endif
 
+static uint32_t ia2_get_pkey() {
+  uint32_t pkru;
+  __asm__("rdpkru" : "=a"(pkru) : "a"(0), "d"(0), "c"(0));
```
fw-immunant, line 41: I believe this has to be `__asm__` `volatile` because the pkru value comes out of "nowhere" as far as the optimizer is concerned--`rdpkru` is impure in that it doesn't compute an output from the listed inputs to the inline asm statement.

libia2/include/ia2.h
```diff
@@ -36,6 +36,12 @@
 
 #endif
 
+static uint32_t ia2_get_pkey() {
```
fw-immunant, line 39: I would rather this were named `ia2_get_pkru`. The terminology convention we're following here is that there are 16 pkeys controlled by the one pkru register, and its value determines which set of pkeys the process has.

fw-immunant: We have several definitions of this as `rdpkru` in tests that should be changed to use this definition as a follow-up.

libia2/include/permissive_mode.h
```diff
@@ -1,10 +1,27 @@
...
+
+/*
+ * PKRU is defined to be bit 9 in the XSAVE state component bitmap by the Intel
+ * manual.
+ */
+#define XFEATURE_PKRU 9
+#define XFEATURE_MASK_PKRU		(1 << XFEATURE_PKRU)
+
+/* Offset of the XSAVE header in bytes from the XSAVE area base */
+#define XSAVE_HEADER_OFFSET 512
```
fw-immunant, line 24: Where does this constant come from? The Intel [Software Developers Manual](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-1-manual.pdf) says the offset used in has to be looked up via cpuid (§13.4.3):
> As noted in Section 13.2, CPUID.(EAX=0DH,ECX=9):EBX enumerates the offset (in bytes, from the base of the
XSAVE area) of the section of the extended region of the XSAVE area used for PKRU state (when the standard
format of the extended region is used).

The alternative to the standard format is the compacted format, which requires adding up sizes for the saved components, and which I'm guessing isn't enabled/relevant here.

fw-immunant (commented):
A few nits, the important two are volatile and documenting the xsave offset stuff.

libia2/test/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_executable(permissive_mode_test permissive_mode.c)
+target_link_libraries(permissive_mode_test libia2)
```
fw-immunant, line 2: Test failures on ubuntu say we need libdl here too.

Is there a reason not to use our `define_test` infrastructure? It includes things like making sure our padded libc is used and propagating the build-system `LIBIA2_INSECURE` into the C preprocessor `LIBIA2_INSECURE`.

libia2/test/permissive_mode.c
```diff
@@ -0,0 +1,26 @@
...
+
+    buffer = mmap(NULL, 4096, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANON, -1, 0);
+
+    pkey_mprotect(buffer, 4096, PROT_READ | PROT_WRITE, 1);
+    buffer[0] = 'a';
+
+    pkey_mprotect(buffer, 4096, PROT_READ | PROT_WRITE, 2);
+    buffer[0] = 'b';
+
+    assert(ia2_get_pkey() == 0xFFFFFFF0);
```
fw-immunant, line 23: These assertions and `pkey_mprotect`s will fail in insecure mode. Probably all of this can just be wrapped in `#ifndef LIBIA2_INSECURE`.

libia2/include/ia2.h
fw-immunant: To support insecure mode, this should be `IA2_RDPKRU` rather than `"rdplru"` and initialize `pkru` to zero.

rinon: We use an output, so we don't need volatile in this particular case. The compiler won't be able to discard this assembly statement because it needs the `pkru` output variable. The only time we _need_ volatile is if the assembly has side-effects not declared as outputs. So a wrpkru probably needs volatile, but reading doesn't.

libia2/include/permissive_mode.h
rinon: This is not the PKRU offset (which we look up from cpuid). This is the offset of the XSAVE header which is always 512 bytes in.

> The XSAVE header of an XSAVE area comprises the 64 bytes starting at offset 512 from the area’s base address:
 Bytes 7:0 of the XSAVE header is a state-component bitmap (see Section 13.1) called XSTATE_BV. It
identifies the state components in the XSAVE area.

The compacted format is never used for the signal handler context here, fortunately. We could and maybe should check the relevant bit of the header to ensure this but :man_shrugging:.

libia2/include/ia2.h
fw-immunant: The concern isn't the compiler removing it entirely, but reordering it with respect to other operations because it considers it to be "pure". For example, it could hoist it above somewhere that we write the PKRU, even if the write is considered side-effecting, because the compiler doesn't know that the write and read are linked through the PKRU.

libia2/include/permissive_mode.h
fw-immunant: Ah, I very much had the wrong context in mind here. This is fine then, though it would be nice to point to the docs where this comes from (Intel manual §13.4.2).

libia2/test/CMakeLists.txt
rinon: I didn't want to actually run the rewriter or wrap anything. TBH I don't know how that macro works anymore and I didn't feel like figuring it out.

libia2/test/permissive_mode.c
rinon: The test makes no sense to run in insecure mode, so I'll just disable the whole test.

fw-immunant (approved):
Looks good now. I'll follow-up to make the rdpkru wrapper work for insecure mode and use it in other tests.

PR #248, opened 2023-07-25
 Switch to compartment and probe global data pages during init

### LINKED ISSUE #172: Initial page protection is indirect and can be replayed

rinon:
Secure global data pkey_mprotect calls by ensuring that we can access
the pages that we are switching to our private compartment. We do this
by switching to the compartment target, verifying that the expected
target compartment and current pkey matches, then attempting to read
from each page in the protection range before protecting the pages as
private to the target compartment.

Fixes #172

### DIFF
--- a/libia2/ia2.c
+++ b/libia2/ia2.c
@@ -1,9 +1,14 @@
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
 #endif
+#include <inttypes.h>
 #include <link.h>
+#include <stdio.h>
 
 #include "ia2.h"
+
+// TODO: Do we want to use sysconf(3) here?
+#define PAGE_SIZE 4096
 
 #ifdef LIBIA2_INSECURE
 size_t ia2_get_pkey() { return 0; }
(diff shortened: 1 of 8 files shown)


libia2/ia2.c
```diff
@@ -341,8 +353,14 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
       }
 
       if (cur_end > start) {
+        for (size_t i = 0; i < cur_end - start; i += PAGE_SIZE) {
+          volatile char* cur = (volatile char*)start + i;
+          (void)*cur;
+        }
```
fw-immunant, line 359: This should be commented to explain that it probes each page so that we only go on to `pkey_mprotect` regions that we can successfully access all of.

I had some misgivings about whether `volatile` accesses could be reordered with system calls but as I understand it they can't, though I don't actually know how the compiler knows which functions *are* system calls, and I do know that `volatile` accesses can be reordered w/r/t non-dependent non-volatile accesses (and presumably "pure" function calls that are not system calls).

libia2/include/ia2.h
```diff
@@ -141,7 +141,21 @@ asm(".macro mov_pkru_eax pkey\n"
...
+                     "xorl %%edx, %%edx\n"                                     \
+                     "mov_pkru_eax " #n "\n" IA2_WRPKRU "\n"                   \
+                     :                                                         \
+                     :                                                         \
+                     : "rax", "rcx", "rdx");                                   \
     dl_iterate_phdr(protect_pages, &args);                                     \
+    __asm__ volatile(/* Set PKRU to the default value */                       \
+                     "xorl %%ecx, %%ecx\n"                                     \
+                     "xorl %%edx, %%edx\n"                                     \
+                     "xorl %%eax, %%eax\n" IA2_WRPKRU "\n"                     \
```
fw-immunant, line 155: I'm not sure "default value" is the right way to describe this--we're zeroing PKRU, which isn't the system's default.

I would say "Set PKRU to fully untrusted".

fw-immunant (approved):
Looks good in general; a couple comment nits inline.

PR #250, opened 2023-07-26
Move defines to libia2 cmake module

rinon:
Binaries that use libia2 should pick up defines from the library module, rather than having to re-declare the same defines every time we use the library. Moves the LIBIA2_INSECURE and LIBIA2_DEBUG definitions to libia2 itself.


fw-immunant (approved):
LGTM, these do seem more correct.

PR #251, opened 2023-07-27
ia2.h: simplify with REPEATB(n, fn, basefn) macro

fw-immunant:
This cleans things up a lot. We need a different base case because we often treat compartment 0 specially.


rinon:
I think a doc string on the macro just to explain what is happening might be useful, but LGTM, thanks.

rinon (approved):

PR #252, opened 2023-07-28
Add Perl module in compartment 3

endbr64:
This runs both RTMP and Perl modules in Nginx. Test Perl by running `curl localhost:8080/perl`.

TODO: 

- [ ] figure out why the rewriter exits with an error
- [ ] ensure this is actually working

### DIFF
--- a/external/nginx/auto/make
+++ b/external/nginx/auto/make
@@ -284,12 +284,16 @@ binary:	$NGX_OBJS${ngx_dirsep}nginx$ngx_binext
 	\$(CC) -shared -fPIC -Wl,-z,now $NGX_OBJS/../wrapper.c \
 		-I $REPO_ROOT/libia2/include -o \$(IA2_CALLGATES)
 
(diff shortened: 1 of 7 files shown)


ayrtonm:
IIRC our build process currently doesn't support more than two compartments. The issue is that call gate wrappers for direct calls are named `__wrap_$FN_NAME`, but they are specific to a pair of pkeys (caller/target). In the case where the main binary provides a function `foo` which both modules need to call, we'll need a call gate wrapper for each compartment for `foo`. Basically something like `__wrap_foo_from_2` and `__wrap_foo_from_3` (the target pkey is implicit since `foo` is defined in compartment 1).

I think we can make this work with something like `objcopy --redefine-sym foo=foo_from_1` as explained in the 'Define call gates for direct transitions' section of #174, but we'll have to reread that to make sure it works. I believe there is not an analogous problem with the names of indirect call gate wrappers, but it'd be worth double checking.

ayrtonm:
See #276 for list of blockers

rinon:
Closing this, as I don't think we have plans to land it. We can always re-open if it becomes relevant.

PR #253, opened 2023-07-28
 Run destructors with correct compartment pkey

### LINKED ISSUE #112: Run compartment destructors with compartment pkey

rinon:
Destructors for each library need to run with the relevant compartment's
pkey and stack in order to have access to the required memory. This
change adds a setup step that redirects the FINI and FINI_ARRAY dynamic
entries to our own shim that is wrapped and runs the relevant
constructors. Although this does not support wrapping destructors in an
unmodified external library (yet), that should also be possible with
this design.

Fixes #112


libia2/include/ia2_init.c
```diff
@@ -0,0 +1,147 @@
...
+
+struct FinalizerInfo {
+  uint64_t fini_offset;
+  uint64_t fini_array_offset;
+  uint64_t fini_array_size;
+};
+
+/* Filled in by runtime initialization with contents of  .fini_array and
+ * .fini */
+static volatile struct FinalizerInfo finalizers
```
rinon, line 55: Needs to be made read-only

rinon:
Finally passes CI! This is ready for review now.

libia2/main.c
```diff
@@ -27,6 +27,15 @@ __attribute__((naked)) int __wrap_main(int argc, char **argv) {
       "call __real_main\n"
       // Restore the old stack pointer before returning.
       "mov main_sp(%%rip), %%rsp\n"
+      // Save return value
+      "mov %%rax,%%rbx\n"
```
ayrtonm, line 31: `__wrap_main` needs to preserve %rbx. Can't we use one of 'rdi, rsi, rdx, rcx, r8, r9, r10, r11' instead?

rewriter/tests/permissive_mode/permissive_mode.c
```diff
@@ -8,6 +11,7 @@ INIT_RUNTIME(1);
 
 int main(int argc, char** argv) {
     char* buffer = NULL;
+#ifndef LIBIA2_INSECURE
```
ayrtonm, line 14: Weren't we removing insecure mode in another PR?

libia2/include/ia2_compartment_init.inc
```diff
@@ -60,3 +65,135 @@ void COMPARTMENT_IDENT(init_tls)(void) {
...
+ * .fini */
+static volatile struct FinalizerInfo finalizers
+    __attribute__((__section__(".data.rel.ro.1"))) = {0};
+
+extern void COMPARTMENT_IDENT(__wrap_ia2_compartment_destructor)(void);
+
+__attribute__((no_sanitize("undefined"))) __attribute__((used)) void
+COMPARTMENT_IDENT(ia2_compartment_destructor)(void) {
+  static bool ran_destructors = false;
+  if (!&__ehdr_start || ran_destructors)
```
ayrtonm, line 88: Is `!&__ehdr_start` valid or a malformed ELF (i.e. should we assert that it's true)?

libia2/include/ia2_compartment_init.inc
```diff
@@ -60,3 +65,135 @@ void COMPARTMENT_IDENT(init_tls)(void) {
   };
   dl_iterate_phdr(protect_tls_pages, &args);
 }
+
+/* Magic symbol from the linker that marks the start of the ELF headers */
+extern Elf64_Ehdr __ehdr_start __attribute__((weak));
```
ayrtonm, line 70: What compartments can write to `__ehdr_start`? I'm assuming it can't be relro since it's defined by the linker right?

libia2/include/ia2_compartment_init.inc
```diff
@@ -60,3 +65,135 @@ void COMPARTMENT_IDENT(init_tls)(void) {
...
+    fprintf(stderr, "Failed to set relro section writable: %d\n", res);
+    exit(1);
+  }
+  for (int i = 0; i < dynamic_phdr->p_memsz / sizeof(Elf64_Dyn); i++) {
+    if (dynamic[i].d_tag == DT_FINI || dynamic[i].d_tag == DT_FINI_ARRAY ||
+        dynamic[i].d_tag == DT_FINI_ARRAYSZ) {
+      uintptr_t dynamic_entry_addr = (uintptr_t)&dynamic[i].d_un.d_val;
+      bool in_finalizers_mapping =
+          dynamic_entry_addr >= (uintptr_t)finalizers_region &&
+          dynamic_entry_addr + sizeof(Elf64_Xword) <=
```
ayrtonm, line 146: This should be `<` since finalizers_region + region_size is on the next page

rinon: I don't think so, `dynamic_entry_addr + sizeof(Elf64_Xword)` is the address of the next byte after the entry.

ayrtonm: nit: Might be more obvious that it's the next byte if we use `&dynamic[i + 1]` like in the assert below.

ayrtonm:
Just finished going through `ia2_setup_destructors`. There are some minor issues when accessing the union in `Elf64_Dyn`, but both variants have the same size so it's fine. The rest of the runtime and rewriter changes look good to me so I just need to go over the CMake changes.

cmake/define-test.cmake
```diff
@@ -187,12 +188,18 @@ function(define_test)
...
     add_custom_command(
-        OUTPUT ${COPIED_SRCS} ${COPIED_HEADERS}
+        OUTPUT ${COPIED_SRCS}
         COMMAND cp ${ORIGINAL_SRCS} ${CMAKE_CURRENT_BINARY_DIR}/
-        COMMAND mkdir -p ${REWRITTEN_INCLUDE_DIR}
-        COMMAND cp ${ORIGINAL_HEADERS} ${REWRITTEN_INCLUDE_DIR}/
-        DEPENDS ${ORIGINAL_SRCS} ${ORIGINAL_HEADERS}
+        DEPENDS ${ORIGINAL_SRCS}
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+    if (NOT ORIGINAL_HEADERS STREQUAL "")
```
ayrtonm, line 195: Aren't you setting `ORIGINAL_TARGETS` instead of `ORIGINAL_HEADERS` in the permissive mode test?

rinon: When calling `define_ia2_wrapper()`, yeah? It's just that this test doesn't have any headers and only a single compartment.

rewriter/SourceRewriter.cpp
```diff
@@ -130,6 +131,13 @@ static bool ignore_file(const Filename &filename) {
   return !filename.starts_with(OutputDirectory) && !filename.starts_with(RootDirectory);
 }
 
+static bool ignore_function(const Filename &filename, const Function &function) {
```
ayrtonm, line 134: Shouldn't be necessary with `--root-directory`.

ayrtonm (approved):
There's a minor nit in the rewriter changes but otherwise the new CMake stuff looks good to me.

rewriter/SourceRewriter.cpp
rinon: I want to specifically _include_ `ia2_compartment_destructor()`, which is outside the source directory, while still ignoring the rest of ia2.h and ia2_internal.h.

PR #254, opened 2023-07-29
Add ia2_init.c for per-compartment initialization

rinon:
Our INIT_COMPARTMENT macro was getting rather large and unwieldy, in addition to being undebuggable because line info was missing for generated functions. Instead we can use a C file with the compartment id defined.

### DIFF
--- a/libia2/include/ia2.h
+++ b/libia2/include/ia2.h
@@ -41,12 +41,7 @@ instead of `fn`. */
 /* Handy as the base-case for repeating from N to 1, excluding 1. */
 #define nop_macro(x)
 
-#define INIT_COMPARTMENT_COMMON(n)                                             \
-  __thread void *ia2_stackptr_##n __attribute__((used));
-
 #ifdef LIBIA2_INSECURE
-
-#define INIT_COMPARTMENT(n) INIT_COMPARTMENT_COMMON(n)
 
 #if !defined(IA2_WRPKRU)
 #define IA2_WRPKRU ""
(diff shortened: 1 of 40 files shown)


ayrtonm (approved):
LGTM

libia2/include/ia2_init.c
```diff
@@ -0,0 +1,39 @@
...
+#define COMPARTMENT_IDENT(name) CONCAT(name##_, IA2_COMPARTMENT)
+
+__thread void *COMPARTMENT_IDENT(ia2_stackptr) __attribute__((used));
+
+#ifndef LIBIA2_INSECURE
+
+// Initializes a compartment with protection key `n` when the ELF invoking this
+// macro is loaded. This must only be called once for each key. The compartment
+// includes all segments in the ELF except the `ia2_shared_data` section, if one
+// exists.
```
fw-immunant, line 18:
```suggestion
// Initializes a compartment with protection key `n`. Called automatically as a
// constructor when the ELF defining this symbol is loaded. This must only be
// called once for each key. The compartment includes all segments in the ELF
// except the `ia2_shared_data` section, if one exists.
```

libia2/include/ia2.h
```diff
@@ -116,8 +107,8 @@ static uint32_t ia2_get_pkru() {
 /// \param data pointer to a PhdrSearchArgs structure
 ///
 /// The callback passed to dl_iterate_phdr in the constructor inserted by
-/// INIT_COMPARTMENT to pkey_mprotect the pages corresponding to the
-/// compartment's loaded segments.
+/// ia2_init.c to pkey_mprotect the pages corresponding to the compartment's
```
fw-immunant, line 110: Slight bikeshed but I think `ia2_init_compartment.inc` would be a better name for the file--`.c` files should be standalone-compileable, and it doesn't initialize ia2 as a whole but only one compartment.

fw-immunant (approved):
No blockers, nits on the filename and a comment.

Issue #260, opened 2023-08-02
Remove insecure (non-PKRU) mode


rinon:
Maintaining a non-PKRU build configuration is becoming burdensome, and we have a CI runner on a PKRU-capable host. Since the insecure mode was just for CI builds before we had this runner, we should remove it to avoid the extra maintenance burden.

PR #262, opened 2023-08-02
Remove insecure mode

rinon:
We don't need the insecure mode any longer as we have a CI runner that is running on a PKRU-capable CPU. Maintaining this mode is extra effort that isn't worthwhile.

Fixes: #260


libia2/include/ia2.h
```diff
@@ -149,19 +126,9 @@ asm(".macro mov_pkru_eax pkey\n"
...
-static int insecure_pkey_mprotect(void *ptr, size_t len, int prot, int pkey) {
-  return 0;
-}
-#define INIT_RUNTIME(n)                                                        \
-  int ia2_n_pkeys_to_alloc = 0;                                                \
-  INIT_RUNTIME_COMMON(n)
-#else
 #define INIT_RUNTIME(n)                                                        \
   int ia2_n_pkeys_to_alloc = n;                                                \
   INIT_RUNTIME_COMMON(n)
```
ayrtonm, line 131: nit: `INIT_RUNTIME_COMMON` can be inlined here since it was just the runtime init that was common to both INSECURE and the default.

ayrtonm (approved):
LGTM, just one minor nit in a macro in ia2.h

rinon:
Good point, folded the INIT_RUNTIME macro together now.

libia2/threads.c
```diff
@@ -33,7 +33,7 @@ void *ia2_thread_begin(void *arg) {
   __asm__ volatile(
       /* clang-format off */
       "xor %%ecx,%%ecx\n"
-      IA2_RDPKRU "\n"
+      "rdpkru\n"
       /* clang-format on */
       : "=a"(pkru)::"ecx", "edx");
   /* returns NULL in insecure mode, because we can't read a pkru there. */
```
fw-immunant, line 39: Now outdated.

rinon: Ah, yes, fixed in 5cdb743

PR #266, opened 2023-08-10
Refactor ia2.h into public and private headers

### LINKED ISSUE #160: Deduplicate code in libia2 headers

rinon:
Adds ia2_internal.h for internal implementation definitions that users don't need to see. Refactors everything internal into this header and combines macro definitions.

Fixes #160

### DIFF
--- a/libia2/ia2.c
+++ b/libia2/ia2.c
@@ -7,8 +7,11 @@
 
 #include "ia2.h"
 
-// TODO: Do we want to use sysconf(3) here?
-#define PAGE_SIZE 4096
+__attribute__((__used__)) uint32_t ia2_get_pkru() {
+  uint32_t pkru = 0;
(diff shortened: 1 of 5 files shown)


ayrtonm (approved):
LGTM

Issue #233, opened 2023-06-28
Syscall prioritization


fw-immunant:
There are hundreds of x86_64 syscalls, categorized in #231. We want to start by winnowing these down to a starting subset that we can allowlist and implement specialized policies for as needed. Once we're secure on programs that use only the allowlisted subset, we can expand to support more programs as needed.

I've surveyed a few programs, and the below are dedup'd+sorted lists of syscalls executed.

Collected with:
```sh
strace -f $PROGRAM 2>&1 | grep -o -P '^[^ =(]+(?=\()' | sort -u
```

<details><summary>ls</summary>

```
access
arch_prctl
brk
close
execve
exit_group
getdents64
getrandom
ioctl
mmap
mprotect
munmap
newfstatat
openat
prctl
pread64
prlimit64
read
rseq
set_robust_list
set_tid_address
write
```
</details>

<details><summary>sh -c exit</summary>

```
access
arch_prctl
brk
close
execve
exit_group
futex
getdents64
getegid
geteuid
getgid
getpgrp
getpid
getppid
getrandom
getuid
ioctl
mmap
mprotect
munmap
newfstatat
openat
pread64
prlimit64
read
rseq
rt_sigaction
rt_sigprocmask
set_robust_list
set_tid_address
sysinfo
uname
```
</details>

<details><summary>w</summary>

```
access
alarm
arch_prctl
brk
close
connect
execve
exit_group
fcntl
futex
getdents64
getrandom
ioctl
lseek
mmap
mprotect
munmap
newfstatat
openat
prctl
pread64
prlimit64
read
rseq
rt_sigaction
set_robust_list
set_tid_address
socket
write
```
</details>

<details><summary>echo 4+4 | python3 -i</summary>

```
access
arch_prctl
brk
chown
close
execve
exit_group
fcntl
futex
getcwd
getdents64
getegid
geteuid
getgid
getpid
getrandom
gettid
getuid
ioctl
lseek
mmap
mprotect
munmap
newfstatat
openat
pread64
prlimit64
read
readlink
rename
rseq
rt_sigaction
setfsgid
setfsuid
set_robust_list
set_tid_address
sysinfo
write
```
</details>

Across our tests, we have the following syscall counts:

     22 access
     22 arch_prctl
     22 brk
      1 clone3
     22 close
     23 execve
     20 exit_group
     22 futex
     22 getrandom
     22 mmap
     22 mprotect
     22 munmap
     22 newfstatat
     22 openat
[...]

PR #267, opened 2023-08-10
Syscall filtering PoC

fw-immunant:
This adds some demos and infrastructure used by both those demos and eventually the final runtime. We still need to weld together the pieces in the manner described best in our slides, and the memory-map tracking is still WIP. I would plausibly prefer rewriting part of the memory-map tracking in Rust (`memory_map.c`) to have access to fancier data structures and more assurances.


fw-immunant:
There are demos of four runtime components here:
- `read-pkru` demonstrates reading an inferior's current pkru value with ptrace, which is needed for memory-map tracking.
- `seccomp-filter` demonstrates the basic structure of the seccomp syscall policy we need; to make this into the MVP for IA2 itself we'll just have to add a clause for each of the 24ish syscalls from #233.
- `landlock` demonstrates the path exclusion we need for `/proc/self/mem`; try running `DENY_PATH=/proc/self/mem ./landlock cat /proc/self/mem` and compare to the same with a different `DENY_PATH` or programs that interact with a specified file in different ways.
- `track-memory-map` demonstrates memory-map tracking; try running `../rewriter/tests/mmap_loop/mmap_loop_main_wrapped &` followed by `sudo ./track-memory-map $(pgrep mmap_loop)`.

fw-immunant:
`memory_map.c` currently relies on linear scans of a vector of memory mappings; I would much rather this used a tree so we get sustainable (logarithmic) scaling. We could open-code a simple binary tree in C, but I would be happier using Rust's BTreeMap, something along [these lines](https://docs.rs/nonoverlapping_interval_tree/).

rewriter/tests/mmap_loop/main.c
```diff
@@ -0,0 +1,53 @@
...
+    This program tests that mmap and heap allocations are handled properly.
+*/
+
+INIT_RUNTIME(2);
+#define IA2_COMPARTMENT 1
+#include <ia2_compartment_init.inc>
+
+int main() {
+  // wait to be traced
+  sleep(2);
```
fw-immunant, line 24: This is only necessary for now because our memory-map tracking demo ptraces an existing process instead of doing fork+exec. The actual runtime will fork so the tracee is its child.

runtime/seccomp_filter.c
```diff
@@ -0,0 +1,135 @@
...
+
+long syscall(long no, ...);
+
+int configure_seccomp(void) {
+  struct sock_filter filter[] = {
+      BPF_STMT(BPF_LD | BPF_W | BPF_ABS, (offsetof(struct seccomp_data, nr))),
+      BPF_SYSCALL_POLICY(write, ALLOW),
+      // this would compare syscall number to write() and allow if it matches
+      /*BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_write, 0, 1),
+      BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_ALLOW),*/
```
rinon, line 44: I don't think these comment are necessary, the macros is pretty straightforward.

runtime/seccomp_filter.c
```diff
@@ -0,0 +1,135 @@
...
+  // able to gain privs via exec() of setuid binaries as they would inherit our
+  // seccomp filters.
+  prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0);
+
+  printf("calling seccomp()\n");
+  // we must make two separate calls to seccomp() here because we want to create
+  // a user notification fd to pass to the supervisor, but we also want to pass
+  // FLAG_TSYNC, and these two cannot be combined in one call because they
+  // impose conflicting interpretations on the syscall return value.
+  int sc_unotify_fd = syscall(SYS_seccomp, SECCOMP_SET_MODE_FILTER,
```
rinon, line 76: Might be worth adding a comment after this saying what we would do with this fd?

runtime/landlock.h
```diff
@@ -0,0 +1,99 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <sys/prctl.h>
+#include <sys/stat.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include "strv.h"
+
+#ifndef landlock_create_ruleset
```
rinon, line 21: If this symbol is defined as a function rather than a macro, will this work? What is the point of these `ifndef`s?

runtime/track_memory_map_demo.c
```diff
@@ -0,0 +1,405 @@
...
+  ptrace(PTRACE_SYSCALL, pid, 0, 0);
+  waitpid(pid, NULL, 0);
+  fprintf(stderr, "continued\n");
+
+  if (ptrace(PTRACE_GETREGS, pid, 0, &regs) < 0) {
+    perror("could not PTRACE_GETREGS");
+    return;
+  }
+  /* return -EPERM */
+#include <errno.h>
```
rinon, line 283: nit: move all includes to top

runtime/memory_map.c
rinon, line 1: I'm skipping close review of this file as I like the idea of doing this in Rust instead. It LGTM for now.

rinon (approved):
Mostly just minor comments.

Could you add a bit of doc comments (e.g. what you put in the PR comments) for each of the POC main files? Just so someone can tell what a thing should do when they open it.

I like the idea of doing the memory map tracker in Rust, let's do it.

runtime/get_inferior_pkru.c
```diff
@@ -0,0 +1,42 @@
...
+
+  /* Pre-4.14 kernels have a bug (fixed by commit 0852b374173b
+     "x86/fpu: Add FPU state copying quirk to handle XRSTOR failure on
+     Intel Skylake CPUs") that sometimes causes the mxcsr location in
+     xstateregs not to be copied by PTRACE_GETREGSET.  Make sure that
+     the location is at least initialized with a defined value.  */
+  memset(xstateregs, 0, sizeof(xstateregs));
+  iov.iov_base = xstateregs;
+  iov.iov_len = sizeof(xstateregs);
+  if (ptrace(PTRACE_GETREGSET, pid, (unsigned int)NT_X86_XSTATE, (long)&iov) <
```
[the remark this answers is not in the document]
fw-immunant: Yeah, I was following a pattern from gdb but I don't think it actually makes sense. Will fix.

runtime/mem_region.h
rinon, line 1: Empty file?

runtime/memory-map/src/disjoint_interval_tree.rs
rinon, line 1: This isn't your code, it should be clearly marked as such and either pulled in via cargo or vendored in external.

rinon (commented):
Have a few comments from last time that still seem relevant. New stuff looks ok, but I'll take another look tomorrow.

runtime/memory-map/src/disjoint_interval_tree.rs
fw-immunant: Yeah, we need to vendor it or otherwise set something up where we can patch it because the implementation of `range` needs to be modified to fix a significant bug. It might be worth just rewriting the entire module ourselves so we don't have to bother, as there isn't a lot of code there.

fw-immunant: For now I've put a fork up at https://github.com/fw-immunant/non-overlapping-interval-tree/, made PRs upstream, and added the fork to our `Cargo.toml`.

rewriter/tests/mmap_loop/main.c
```diff
@@ -0,0 +1,53 @@
...
+  char *buf = malloc(4096);
+  buf[0] = 'r';
+  assert(buf[0] == 'r');
+
+  char *lib_buf = lib_malloc_buf(4096);
+  /*printf("%p was r?\n", &lib_buf[0]);
+  bool x = CHECK_VIOLATION(lib_buf[0] = 'r');
+  printf("%s\n", x ? "true" : "false");*/
+
+  {
```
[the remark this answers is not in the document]
fw-immunant: It doesn't really need to be; it looped at one point to give time for a ptracer to attach, but now we start the inferior ourselves.

rewriter/tests/mmap_loop/main.c
```diff
@@ -0,0 +1,53 @@
...
+  sleep(2);
+
+  char *buf = malloc(4096);
+  buf[0] = 'r';
+  assert(buf[0] == 'r');
+
+  char *lib_buf = lib_malloc_buf(4096);
+  /*printf("%p was r?\n", &lib_buf[0]);
+  bool x = CHECK_VIOLATION(lib_buf[0] = 'r');
+  printf("%s\n", x ? "true" : "false");*/
```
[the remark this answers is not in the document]
fw-immunant: This test was a staging ground for various one-off checks while developing the syscall tracing; now, it serves as a test of something that should fail with our syscall policy (compartment 1 unmapping compartment 2's buffer) but which works fine without it.

runtime/mem_region.h
fw-immunant: Removed.

runtime/memory-map/src/disjoint_interval_tree.rs
fw-immunant: We now use the latest version from crates.io.

fw-immunant:
I said I wasn't going to clean up this history but I ended up doing it just to make sure the code was all in good shape. As such, merging via rebase rather than squash. Some of the intermediate commits are worth noticing individually as they're changes to our memory-management policy needed to run real programs, which we may want to revisit later.

PR #268, opened 2023-08-11
Add --output-directory rewriter option to avoid rewriting in-place

### LINKED ISSUE #208: Enable repeated rewriter runs

ayrtonm:
This adds the --output-directory option to the rewriter to avoid rewriting
source and header files in-place. It also modifies the test setup to avoid
copying over source files in CMake and instead use the rewriter. By copying over
sources with the rewriter instead of rewriting in-place we can rerun the
rewriter without having to parse the output of previous rewriter runs. Also
treating the rewriter output as "intermediate .c files" allows us to integrate
with existing build systems more cleanly since the compile commands don't need
to `-include` the output header which is generated by the rewriter itself.

The --root-directory option is not strictly necessary for the --output-directory
flag, but simplifies how we distinguish files that should be copied from those
that shouldn't. Also while in theory we could make these flags optional and
default to rewriting in-place for now they are required to avoid having to
maintain and test multiple rewriter configurations.

Closes #208.


rewriter/SourceRewriter.cpp
```diff
@@ -40,27 +42,53 @@ typedef std::string OpaqueStruct;
...
-static llvm::cl::opt<bool>
-    BackupSources("backup-sources", llvm::cl::Optional,
-                  llvm::cl::cat(SourceRewriterCategory),
-                  llvm::cl::desc("Back up source files before rewriting"));
-
 // Map each translation unit's filename to its pkey.
 static std::map<Filename, Pkey> file_pkeys;
 
+static std::optional<std::string> try_map_input_file(const std::string &path) {
+  bool in_root_dir = path.starts_with(RootDirectory);
```
rinon, line 64: This requires that both paths are absolute, right? This is surprising to me, I would expect some sort of canonicalization first. Perhaps `FileManager::makeAbsolutePath()` or something with realpath of the `FileEntry`?

rewriter/SourceRewriter.cpp
```diff
@@ -252,8 +278,16 @@ class FnPtrTypes : public RefactoringCallback {
...
-    Replacement r{sm, expansion_range, new_decl};
-    auto err = file_replacements[file_name].add(r);
+
+    unsigned offset =
+        sm.getFileOffset(sm.getExpansionLoc(expansion_range.getBegin()));
+    auto char_range =
+        clang::CharSourceRange::getTokenRange(expansion_range.getAsRange());
+    auto orig_decl =
+        clang::Lexer::getSourceText(char_range, sm, ctxt.getLangOpts());
+    unsigned int len = orig_decl.size();
```
rinon, line 288: I think you're better off constructing the old replacement as a temporary, then calling its accessors to pull out canonical offset and length to create a new replacement with the new file path, old offset and length. Can even extract that into a helper function. There's some tricky logic in `Replacement::setFromSourceRange()` that I don't think we want to duplicate.

rewriter/SourceRewriter.cpp
```diff
@@ -798,6 +846,52 @@ int main(int argc, const char **argv) {
...
 
+  for (auto s : options_parser.getSourcePathList()) {
+    assert(s.starts_with(RootDirectory));
+    std::string relative_path = s.substr(RootDirectory.length());
+    std::string new_path = OutputDirectory;
+    new_path.append(relative_path);
+
+    std::string parent_dir = new_path;
+    auto n = parent_dir.rfind('/');
+    parent_dir = parent_dir.substr(0, n);
```
rinon, line 857: `llvm::parent_path()`

rewriter/SourceRewriter.cpp
```diff
@@ -798,6 +846,52 @@ int main(int argc, const char **argv) {
                        options_parser.getSourcePathList());
   CompilationDatabase &comp_db = options_parser.getCompilations();
 
+  for (auto s : options_parser.getSourcePathList()) {
+    assert(s.starts_with(RootDirectory));
+    std::string relative_path = s.substr(RootDirectory.length());
+    std::string new_path = OutputDirectory;
+    new_path.append(relative_path);
```
rinon, line 853: `llvm::replace_path_prefix()`?

rewriter/SourceRewriter.cpp
```diff
@@ -798,6 +846,52 @@ int main(int argc, const char **argv) {
...
+
+    std::string parent_dir = new_path;
+    auto n = parent_dir.rfind('/');
+    parent_dir = parent_dir.substr(0, n);
+    if (!opendir(parent_dir.c_str())) {
+      auto cmd = "mkdir -p "s + parent_dir;
+      system(cmd.c_str());
```
rinon, line 860: `system()` always scares me because we're passing an input param to a shell. Would it be too hard to recursively check parents and call `mkdir(3)`?

rewriter/SourceRewriter.cpp
```diff
@@ -798,6 +846,52 @@ int main(int argc, const char **argv) {
...
+    std::ofstream dst(new_path, std::ios::binary);
+    dst << src.rdbuf();
+  }
+
+  std::set<Filename> copied_headers;
+  auto asts = std::vector<std::unique_ptr<clang::ASTUnit>>();
+  tool.buildASTs(asts);
+  for (auto &ast : asts) {
+    auto &sm = ast->getSourceManager();
+    for (auto it = sm.fileinfo_begin(); it != sm.fileinfo_end(); it++) {
```
rinon, line 874: Won't this iterate over all sources, not just headers? Do we need the copy above?

cmake/define-ia2-wrapper.cmake
```diff
@@ -28,30 +28,31 @@ function(define_ia2_wrapper)
...
 
     add_custom_command(
         OUTPUT ${CALL_GATE_SRC} ${CALL_GATE_HDR} ${LD_ARGS_FILES}
+               ${ALL_SRCS}
         COMMAND ia2-rewriter
             --output-prefix=${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}_call_gates
-            ${ALL_SRCS}
+            --root-directory=${CMAKE_CURRENT_SOURCE_DIR}
+            --output-directory=${CMAKE_CURRENT_BINARY_DIR}
+            -p=${CMAKE_BINARY_DIR}
```
rinon, line 45: What does this do? Is there a long option version?

cmake/define-test.cmake
```diff
@@ -54,37 +54,22 @@ function(define_shared_lib)
...
-    add_library(${LIBNAME} SHARED ${COPIED_SRCS})
+    add_library(${LIBNAME} SHARED ${ORIGINAL_SRCS})
     set(CMAKE_EXPORT_COMPILE_COMMANDS OFF)
     add_library(${WRAPPED_LIBNAME} SHARED ${COPIED_SRCS})
     add_dependencies(${WRAPPED_LIBNAME} ${TEST_NAME}_call_gate_generation)
     set_target_properties(${LIBNAME} PROPERTIES EXCLUDE_FROM_ALL 1)
 
     target_compile_definitions(${LIBNAME} PRIVATE PRE_REWRITER=1)
+    if (${SHARED_LIB_PKEY} GREATER 0)
+        target_include_directories(${LIBNAME} BEFORE PRIVATE ${ORIGINAL_INCLUDE_DIR})
```
rinon, line 67: This include dir is the same in both branches.

cmake/define-test.cmake
```diff
@@ -54,37 +54,22 @@ function(define_shared_lib)
...
+        target_include_directories(${WRAPPED_LIBNAME} BEFORE PRIVATE ${REWRITTEN_INCLUDE_DIR})
```
rinon, line 68: Shouldn't this be under the foreach?

cmake/define-ia2-wrapper.cmake
ayrtonm: It sets the build path which we need so the tool finds the compile commands JSON (otherwise it'll look in parent directories of the source file's directory and won't find the JSON in the build dir). I'll document this.

cmake/define-test.cmake
ayrtonm: Oh right, I'll pull it out of the if/else.

cmake/define-test.cmake
ayrtonm: No, we only want to use the rewritten headers for the wrapped target.

cmake/define-test.cmake
ayrtonm: and do the same in `define_test`

rewriter/SourceRewriter.cpp
ayrtonm: I think we already required absolute paths when using `bear` but I'd have to double check.

rewriter/SourceRewriter.cpp
```diff
@@ -40,27 +42,53 @@ typedef std::string OpaqueStruct;
 static llvm::cl::OptionCategory
     SourceRewriterCategory("Source rewriter options");
 
+static llvm::cl::opt<std::string>
+    RootDirectory("root-directory", llvm::cl::Required,
+                  llvm::cl::cat(SourceRewriterCategory),
+                  llvm::cl::desc("<root directory for input files>"));
```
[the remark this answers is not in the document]
ayrtonm: Maybe, if we can set `-working-directory` and still find the compile commands JSON then it should be fine.

rewriter/SourceRewriter.cpp
ayrtonm: I think that should work and would definitely be preferable to this.

rewriter/SourceRewriter.cpp
ayrtonm: No, that should be fine and was what I initially started doing but I just wanted to get something that worked first. I can switch it out.

rewriter/SourceRewriter.cpp
ayrtonm: Right I believe we can remove the loop above.

cmake/define-ia2-wrapper.cmake
```diff
@@ -28,30 +28,32 @@ function(define_ia2_wrapper)
...
             --output-prefix=${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}_call_gates
-            ${ALL_SRCS}
+            --root-directory=${CMAKE_CURRENT_SOURCE_DIR}
+            --output-directory=${CMAKE_CURRENT_BINARY_DIR}
+            # Set the build path so the rewriter can find the compile_commands JSON
+            -p=${CMAKE_BINARY_DIR}
+            ${ORIGINAL_SRCS}
         WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-        DEPENDS ia2-rewriter ${ALL_SRCS}
+        DEPENDS ia2-rewriter
```
rinon, line 49: Don't we want a dependency edge from source files to the rewriter?

rinon (approved):

rewriter/SourceRewriter.cpp
ayrtonm: Unfortunately we had to canonicalize paths in this PR since using `fileinfo_iterator` was giving some relative paths for the main binary in the nginx demo. `FileManager::makeAbsolutePath` prepends the working directory of the tool but what we want is to prepend the directory of the compile command. I couldn't find an llvm function that did this so I implemented this myself when iterating over the files in the AST. Then I just saved the results of the mapping from relative paths to full paths.

rewriter/SourceRewriter.cpp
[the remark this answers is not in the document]
ayrtonm: I don't think `-working-directory` is a default option for clang-tools, only clang itself.

ayrtonm:
The nginx demo builds and works, but there are a [few steps missing](https://github.com/immunant/IA2-Phase2/pull/268/commits/085bdfb01a57383802e223ecbad3df392e21ad10) from the configure script. I'd be ok with merging this after adding those changes and moving the rewriter step from the config script to a Makefile rule in a separate PR.

PR #269, opened 2023-08-14
Support protecting writable ELF segments of dependencies

rinon:
Adds the IA2_COMPARTMENT_LIBRARIES define for a compartment to declare any dependency system libraries that should be considered part of the compartment.

Includes build system support to fix dlopen from compartmentalized libraries and video player demo integration for a couple system libraries


libia2/ia2.c
```diff
@@ -123,6 +124,49 @@ static bool in_loaded_segment(struct dl_phdr_info *info, Elf64_Addr address) {
...
+  }
+  size_t extra_libraries_len = strlen(extra_libraries);
+  size_t extra_libraries_pos = 0;
+  size_t library_name_pos = 0;
+  while (extra_libraries_pos < extra_libraries_len) {
+    // Compare this extra library with the current library
+    for (library_name_pos = 0; extra_libraries_pos < extra_libraries_len &&
+                               library_name_pos < library_name_len;
+         extra_libraries_pos++, library_name_pos++) {
+      if (extra_libraries[extra_libraries_pos] !=
```
ayrtonm, line 144: Shouldn't this use strcmp or am I missing something?

rewriter/SourceRewriter.cpp
```diff
@@ -50,6 +50,11 @@ static llvm::cl::opt<bool>
                   llvm::cl::cat(SourceRewriterCategory),
                   llvm::cl::desc("Back up source files before rewriting"));
 
+static llvm::cl::list<std::string>
+    IgnorePrefixes("ignore-prefix",
```
ayrtonm, line 54: Might not be necessary since only things under `--root-directory` are rewritten.

cmake/define-ia2-wrapper.cmake
```diff
@@ -1,13 +1,36 @@
+# We need to get the system headers specifically from clang
```
ayrtonm, line 1: Why does this need to be specifically from clang?

cmake/define-ia2-wrapper.cmake
```diff
@@ -40,6 +63,9 @@ function(define_ia2_wrapper)
         OUTPUT ${CALL_GATE_SRC} ${CALL_GATE_HDR} ${LD_ARGS_FILES}
         COMMAND ia2-rewriter
             --output-prefix=${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}_call_gates
+            --extra-arg=-isystem "--extra-arg=${CLANG_HEADERS_INCLUDE}"
+            --extra-arg=-isystem "--extra-arg=${CLANG_HEADERS_INCLUDE_FIXED}"
```
ayrtonm, line 67: I think adding `-isystem` as a `target_compile_option` is preferable to this, though it depends on why we have to use clang headers in the first place.

libia2/ia2.c
rinon: This is doing per-char comparison, but strncmp would work instead, and is probably cleaner. I'll update this.

rewriter/SourceRewriter.cpp
rinon: It's not necessary, and in fact is unused now. We could probably just drop this.

cmake/define-ia2-wrapper.cmake
rinon: This is actually from the video player PR (#246), so should be addressed there. This change is built on top of that.

I think the reason is that we are parsing with clang, and on some systems clang can't parse the gcc compiler-rt headers? @ahomescu might know for sure.

cmake/define-ia2-wrapper.cmake
rinon: We don't need it for the actual compiler, just the parsing with clang part.

libia2/ia2.c
rinon: Actually on second thought, this is simpler. I'll comment it better. We're actually comparing substrings to substrings, so using a `strncmp` would require `strchr` too.

ayrtonm: Oh right, I misread the arg type as `char **` and thought you were comparing pointers at first. I went through it again and it looks correct.

ayrtonm:
It might be good to add a check after the `dl_iterate_phdr` to ensure that all shared objects in `IA2_COMPARTMENT_LIBRARIES` were actually protected but that can be a follow-up. Otherwise LGTM

PR #273, opened 2023-08-19
rewriter: Add clang attribute to skip wrapping functions

rinon:
Some functions, e.g. utility functions that don't touch global state,
should not be wrapped and should always run in the caller compartment.
If we wrap these functions they may act as simple accessors for another
compartment's private data. For example, consider an vector_pop()
function that pops an element from a vector. If this function is
declared in compartment A and used from compartment B, it allows
compartment B to pop any element from compartment A's private vectors.

This change adds the `ia2_skip_wrap` annotation attribute that skips
wrapping of any functions, and function pointers it is applied to (and
calls to these function pointers). It should be applied with the
following set of pragmas:

    #pragma clang attribute push (__attribute__((annotate("ia2_skip_wrap"))), apply_to = hasType(functionType))

    // function declarations...

    #pragma clang attribute pop

### DIFF
--- a/rewriter/SourceRewriter.cpp
+++ b/rewriter/SourceRewriter.cpp
@@ -17,10 +17,13 @@
 #include "llvm/Support/FormatVariadic.h"
 #include "llvm/Support/raw_ostream.h"
 #include <algorithm>
+#include <clang/AST/Decl.h>
 #include <clang/AST/PrettyPrinter.h>
+#include <clang/Basic/SourceLocation.h>
 #include <dirent.h>
 #include <fstream>
 #include <iostream>
+#include <llvm/ADT/Optional.h>
 #include <map>
 #include <optional>
 #include <string>
@@ -33,6 +36,8 @@
 using namespace clang::ast_matchers;
 using namespace clang::tooling;
 using namespace std::string_literals;
+
+static constexpr llvm::StringLiteral SKIP_WRAP_ATTR("ia2_skip_wrap");
 
 typedef std::string Function;
 typedef std::string Filename;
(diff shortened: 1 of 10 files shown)


rewriter/SourceRewriter.cpp
```diff
@@ -509,8 +528,13 @@ class FnPtrExpr : public RefactoringCallback {
...
                            ignoringImplicit(equalsBoundNode("fnPtrExpr"))))));
 
+    auto param_expr = hasParent(implicitCastExpr(hasParent(callExpr(
+        forEachArgumentWithParam(equalsBoundNode("fnPtrExpr"),
+                                 parmVarDecl().bind("fnPtrParamDecl"))))));
+
     StatementMatcher fn_ptr_expr =
-        expr(fn_expr, unless(call_expr), unless(in_bin_op));
+        expr(fn_expr, unless(call_expr), unless(in_bin_op),
+             anyOf(param_expr, anything()));
```
ayrtonm, line 537: Why `anyOf(anything())`?

rinon: Because `expr()` is implicitly `allOf` and we want to match things that aren't call arguments. Is there a better way to do this? I basically just wanted to bind the param decl to a name with the `param_expr` matcher. Probably deserves a comment at least though.

ayrtonm: I vaguely remember running into a similar issue with the callsite AST matcher but I'll take a closer look this afternoon.

ayrtonm: It's not quite the same thing I ran into before, but I think using `optionally` might make your intent more clear.
```suggestion
             optionally(param_expr));
```

ayrtonm:
LGTM aside from the slight clarification I mentioned about the AST matcher.

rewriter/SourceRewriter.cpp
rinon: I missed that operator, that's perfect, thanks.

PR #274, opened 2023-08-19
Improve nginx sharing and reduce PKRU violations

rinon:


external/nginx/src/core/ngx_cycle.c
```diff
@@ -5,6 +5,7 @@
  */
 
 
+#include "ia2.h"
```
ayrtonm, line 8: nit: `<ia2.h>`?

rinon:
@ayrtonm friendly ping, could you take a look at this?

ayrtonm (approved):
LGTM. I think we could simplify how the variables you moved from the stack to the heap are cleaned up with something like
```
void free2(void *p) {
    if (!p || !*p) { return; }
    free(*(void **)p);
}
int foo(void) {
    // previously: int x = ...;
    __attribute__((cleanup(free2))) int *x = ...;
}
```
but I think it's a bit difficult in this particular case since we're using pools so we need a way to specify the pool used in each call to palloc.

rinon:
I'd prefer to not add `cleanup` attributes unless nginx already uses them, but that's a good point in general.

PR #278, opened 2023-08-30
Rename IA2_PREREWRITER to IA2_ENABLE and fix -DIA2_ENABLE=0

### LINKED ISSUE #277: Fix flag for non-compartmentalized builds
The `-DIA2_PREREWRITER=1` flag makes the most common IA2 macros no-ops which we need because the nginx demo requires a build in the config step to create some `.c`s generated by the build system. While this works for what we need (i.e. running make sucessfully), the binaries generated by this build step are not actually functioning non-compartmentalized builds. This is because there are other parts of the runtime and definitions which are not removed by defining this macro. We should fix this macro so that defining it produces working non-compartmentalized programs.

ayrtonm:
Fixes #277 and the nginx demo reconfigure step. We should also be able to build working non-compartmentalized binaries with cmake now, but I didn't enable them in the `check-ia2` target to keep our testing setup simple.


libia2/include/ia2_compartment_init.inc
```diff
@@ -25,6 +25,11 @@
 #define STR(x) #x
 
 __thread void *COMPARTMENT_IDENT(ia2_stackptr) __attribute__((used));
+#if IA2_DISABLE
+void COMPARTMENT_IDENT(ia2_setup_destructors)(void) { }
```
ayrtonm, line 29: TODO: Add comment on why these need to be defined when compartments are disabled.

rinon:
I think I'd prefer flipping this to IA2_ENABLE and have the default be disabled. Is there a good reason not to do that?

ayrtonm:
I made the default enabled in case the flag is accidentally omitted, but I don't have a strong opinion on this.

rinon:
We should have a post-link binary validator at some point anyway, so I'm more concerned about not being able to build without build system modifications after inserting annotations.

PR #280, opened 2023-09-01
partition-alloc: Add compartment support to upstream

rinon:
Adds up to 16 new pools for per-pkey compartment pools to the upstream AddressPoolManager. These compartment pools are pkey protected when created and all allocations from them are protected with their pkey. This support is built on the upstream thread isolation code, but replaces the single thread isolation pool with 16 different per-compartment pools.


partition-alloc/src/allocator_shim_default_dispatch_to_partition_alloc.cc
```diff
@@ -0,0 +1,679 @@
...
+                partition_alloc::PartitionOptions::BackupRefPtr::kDisabled,
+            .thread_isolation = partition_alloc::ThreadIsolationOption(pkey, pkey),
+        });
+
+    return new_root;
+  }
+};
+
+LeakySingleton<partition_alloc::PartitionRoot, MainPartitionConstructor>
+    g_root[partition_alloc::internal::kNumCompartments] PA_CONSTINIT = {};
```
ayrtonm, line 163: Don't we only need 15 partition roots here? I think `g_root[0]` is unused.

partition-alloc/CMakeLists.txt
```diff
@@ -80,12 +84,11 @@ target_compile_options(partition-alloc
 		"-Wno-invalid-offsetof"
 		"-Wno-return-type"
 )
+target_compile_options(partition-alloc
+	PUBLIC
+		"-fno-exceptions"
```
ayrtonm, line 89: Why is this necessary?

partition-alloc/src/allocator_shim_default_dispatch_to_partition_alloc.cc
```diff
@@ -0,0 +1,679 @@
...
+
+    return new_root;
+  }
+};
+
+LeakySingleton<partition_alloc::PartitionRoot, SharedPartitionConstructor>
+    g_shared_root PA_CONSTINIT = {};
+
+partition_alloc::PartitionRoot *Allocator() {
+  size_t pkey = ::ia2_get_pkey();
```
ayrtonm, line 185: This returns pkey 0 when the pkru is not one of the 16 expected values. We should fix that, but it can be in a separate change if you want.

external/chromium/src/base/allocator/partition_allocator/thread_isolation/thread_isolation.cc
```diff
@@ -52,30 +52,39 @@ int MprotectWithThreadIsolation(void* addr,
...
+  //                                    PartitionAddressSpace::setup_);
 
-  AddressPoolManager::Pool* pool =
-      AddressPoolManager::GetInstance().GetPool(kThreadIsolatedPoolHandle);
+  pool_handle handle = PoolHandleForCompartment(thread_isolation.compartment);
+  AddressPoolManager::Pool *pool =
+      AddressPoolManager::GetInstance().GetPool(handle);
   WriteProtectThreadIsolatedVariable(
-      thread_isolation, *pool,
-      offsetof(AddressPoolManager::Pool, alloc_bitset_));
```
ayrtonm, line 63: Did you intentionally drop the `offsetof` argument?

partition-alloc/CMakeLists.txt
rinon: Partition alloc shim doesn't build with exceptions:

```c++
#if (defined(__GNUC__) && defined(__EXCEPTIONS)) || \
    (defined(_MSC_VER) && defined(_CPPUNWIND))
#error This code cannot be used when exceptions are turned on.
#endif
```

partition-alloc/src/allocator_shim_default_dispatch_to_partition_alloc.cc
rinon: Is that not the behavior we want here? Early in process initialization we will have the default pkey, not 0.

external/chromium/src/base/allocator/partition_allocator/thread_isolation/thread_isolation.cc
rinon: Yes, we are protecting the entire pool, including the lock, on the same page. Previously the single thread isolation pool was carefully aligned to put the lock on the previous page, but that would double the number of pages we need and I don't really care about stats printing from out of the compartment (which is what the lock was used for outside the thread isolation thread).

partition-alloc/src/allocator_shim_default_dispatch_to_partition_alloc.cc
ayrtonm: Do we expect allocations during early process init? IIRC the dynamic linker has its own basic allocator for early allocations and I don't think our runtime does any. Regardless I'd expect the behavior for an unknown pkey to be terminating the process, but I guess allowing shared allocations in this case doesn't hurt.

external/chromium/src/base/allocator/partition_allocator/thread_isolation/thread_isolation.cc
ayrtonm: ok, that sounds reasonable. I'll have to take a closer look at the struct layout again.

partition-alloc/src/allocator_shim_default_dispatch_to_partition_alloc.cc
rinon: I think for now we should allow them, we can always change this later.

ayrtonm (approved):
Changes after switching back to ld --wrap look good to me

Issue #154, opened 2022-08-22
Improve Testing Infrastructure


rinon:
Our test infra for this project is pieced together and requires a significant amount of duplication to add new tests. We should probably automate the bulk of this to avoid error-prone duplication between tests.

rinon:
We discussed moving tests around to condense header-rewriter tests and functional tests. Main goal is to make adding new tests easier, rather than having to replicate an entire existing directory.

fw-immunant:
Cross-referencing a [Slack](https://immunant.slack.com/archives/C010R2PU3NF/p1660671369087429) comment that summarized the discussion the last time this came up: 

> Frances and I were talking about how to improve our testing framework last week and here are some notes from the conversation. They're a bit unorganized, but listed roughly in order of importance. The main issue is that writing tests is cumbersome since no one has taken the time to improve how we write tests, so we usually end up copying old tests and changing them.
> - we don't have a good way to test small variations of a program's runtime behavior and currently use ad-hoc arguments to change each test program's behavior
>     - we should factor argument parsing out of our tests
>     - we should write macros to let us reuse one set of source files for various related tests
>     - should each test variation be a separate binary? it may be easier to build just one binary like we do now
> - we should treat variations of each program as separate tests
>     - the LIT commands in each header show up as one test, so if one variation of a program fails you have to run it manually to see which failed
>     - if LIT always treats each file as one test, we could either generate files with the LIT commands at build time or use something other than LIT for testing the runtime
> - we should consider switching from `diff` to `FileCheck` for comparing stdout from tests
>     - this would let us write tests that print addresses
>     - pointer comparisons should still be done by the test program at runtime, but this would provide quicker ways to sanity-check a program
> - we don't distinguish between rewriter tests and runtime tests
>     - the older tests are mainly for the rewriter and have simple runtime behavior
>     - the newer tests have more complex runtime behavior but we're sometimes still adding the LIT annotations for rewriter checks even though it doesn't add to test coverage
>     - could consider separating runtime/rewriter tests into subdirectories or at least stop adding LIT annotations that don't increase coverage (editado)

PR #286, opened 2023-09-21
Improve testing framework

rinon:


rinon:
@sim-immunant here's my start to cleaning up the testing framework. I don't think we're getting much out of `lit` now that we can't do in other ways, perhaps with custom cmake tests that run FileCheck directly?

More context is in #154

fw-immunant (approved):
Looks good--criterion test entrypoints are way better than having to parse arguments to decide which test case to run.

fw-immunant:
I would absolutely like to get rid of `lit` because it has way more complexity than I think we benefit from. The other major thing I think we're doing wrong in testing is testing some properties at every site where they're interacted with instead of once or in one all-encompassing pass. For example, we have `// LINKARGS: --wrap=foo` comments above every wrapped function to test if the wrappers we generate are correct. I don't think this is important coverage, and it would be better to separate unit-testing the rewriter (which is what the linkargs validation does) from our integration testing for IA2 as a whole.

rewriter/tests/ro_sharing/plugin.c
```diff
@@ -35,14 +37,15 @@ const uint32_t *get_plugin_uint(bool secret) {
...
 
 // LINKARGS: --wrap=read_main_uint
 void read_main_uint(const uint32_t *shared, const uint32_t *secret) {
   // Check that we can read a pointer to rodata passed from main
-  LOG("0x%x", *shared);
+  cr_log_info("0x%x", *shared);
 
   // Check that we can't read a pointer to data passed from main
+  // TODO can we change this LOG to cr_log_info?
   CHECK_VIOLATION(LOG("0x%x", *secret));
```
rinon, line 50: Yes, change this to a cr_log and put the `CHECK_VIOLATION` around the `*secret` if we can?

rewriter/tests/global_fn_ptr/operations.c
```diff
@@ -2,18 +2,17 @@
...
-        //printf("%s\n", operations[i].desc.data);
-        uint32_t x = 18923;
-        uint32_t y = 24389;
-        uint32_t res = operations[i].function(x, y);
-        printf("%d = f(%d, %d)\n", res, x, y);
-    }
+// LINKARGS: --wrap=call_operation
+uint32_t call_operation(size_t i) {
+    // TODO: Add a way to share strings between compartments
+    //printf("%s\n", operations[i].desc.data);
```
ayrtonm, line 13: nit: I think we can remove the comment on the printf and the TODO We previously treated the `.rodata` sections (where string literals are placed) as secret, but we have since moved on to only treating writable sections that way.

ayrtonm (approved):

rewriter/tests/sighandler/main.c
```diff
@@ -47,19 +40,25 @@ void install_sighandler(struct handler *h) {
...
-int main() {
-    setbuf(stdout, NULL);
-    install_sighandler(NULL);
-    test_handler();
-
+Test(sighandler, main_ignore) {
     static struct handler h = {
         .handler = IA2_SIGHANDLER(trap_handler),
     };
     install_sighandler(&h);
```
rinon, line 52: How is this any different from the NULL case? It seems to install the same signal handler? Maybe this used to behave differently, but it seems the same to me now...

rewriter/tests/simple1/main.c
```diff
@@ -2,6 +2,9 @@
 RUN: cat main.c | FileCheck --match-full-lines --check-prefix=REWRITER %s
 RUN: cat simple1_call_gates_0.ld | FileCheck --check-prefix=LINKARGS %s
 */
+#include <criterion/criterion.h>
+#include <criterion/logging.h>
+#include <criterion/new/assert.h>
```
rinon, line 7: Do we use the new asserts here?

sim-immunant: Yes, `cr_fatal` is declared in this header.

rewriter/tests/sighandler/main.c
sim-immunant: Maybe I misunderstood, but I thought the idea was to assemble the struct manually to make sure that the rewriter/sighandler macro was setting it up correctly. @ayrtonm thoughts?

ayrtonm: yeah @sim-immunant is right. The behavior at runtime is the same, but the idea was to make sure that `sa.sa_handler = $FN_PTR_EXPR` and `sa.sa_handler = $FN_IDENTIFIER` both work (in terms of both it builds and it runs). This seemed important since this field is one of those function pointer types that we don't rewrite (because the type is defined in a header outside the root directory), so I wanted to make sure the rewriter wasn't doing something wrong like rewriting the function pointer expressions or inserting call gates that shouldn't be there.

rewriter/tests/macro_attr/functions.c
```diff
@@ -1,36 +1,36 @@
...
 RUN: cat macro_attr_call_gates_1.ld | FileCheck --check-prefix=LINKARGS %s
 */
-#include <stdio.h>
+#include <criterion/logging.h>
 #include "functions.h"
 
 // LINKARGS: --wrap=f
 void f() {
-    printf("Called `f()`\n");
+    cr_log_info("Called `f()`");
```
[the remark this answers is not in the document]
sim-immunant: Yes, I compiled in Release mode and the log messages still get printed.

PR #288, opened 2023-09-22
Switch to compartment 0 when calling exit(3)

### LINKED ISSUE #285: Destructors run from a compartment crash after running
If the `__wrap_ia2_compartment_destructor_N` function is invoked from a compartment other than the shared compartment 0, it will crash when it sets the pkey to 0 after running. I assumed destructors are only run from outside `main()` but that's an incorrect assumption.

I'm not sure how to fix this. We could try to transition to a shared stack before running any destructors, perhaps?

rinon:
Destructors are wrapped to run in their respective compartments, but transition back to compartment 0 after running. This means that the caller stack must be writable in compartment 0. This change wraps exit(3) and transitions to compartment 0 and the shared stack before running exit() from libc.

Fixes #285

### DIFF
--- a/libia2/CMakeLists.txt
+++ b/libia2/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.12)
 project(libia2)
 
-add_library(libia2 ia2.c threads.c main.c)
+add_library(libia2 ia2.c threads.c main.c exit.c)
 target_compile_options(libia2 PRIVATE "-fPIC")
 
 if(LIBIA2_DEBUG)
--- a/libia2/exit.c
+++ b/libia2/exit.c
@@ -1,0 +1,33 @@
+#include "ia2.h"
+#include <dlfcn.h>
+
+__attribute__((used))
+static void call_libc_exit(int status) {
+  void (*exit_ptr)(int) = dlsym(RTLD_NEXT, "exit");
+  if (!exit_ptr) {
+    printf("Could not find exit(3) in the next DSO\n");
+    _exit(status);
+  }
+  exit_ptr(status);
+}
+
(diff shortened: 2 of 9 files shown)


fw-immunant (approved):
The changes in the first commit look good to me.

The threads-and-thread-specific-destructors test was a bit confusing to read, but ultimately does make more sense after tracing out the control flow. I was specifically confused by a couple things:
1. the two kinds of "destructors" involved (*thread-specific data* destructors and *shared object* destructors, where only the latter is what we were segfaulting on in #285 aiui)
2.  we refer to the former of these as "tls_destructors" in the test names, but specifically this is using explicit TLS aka "thread-specific storage" rather than implicit TLS. I expected this to be referring to thread_local instances of data types that have destructors--but this isn't something you can do in C as far as I know (C++11 [has it](https://sourceware.org/glibc/wiki/Destructor%20support%20for%20thread_local%20variables) though).

I still don't really understand the *interaction* between the two kinds of destructors here--what situation do the thread tests check? We actually *don't* protect the pthread TSS map per-compartment at present (we'll need to wrap pthread APIs and shard it but there's not much interesting going on there afaik so I deferred on that for the time being), but this test only checks violations accessing the normal `calloc` allocation pointed at by the opaque (to pthreads) value stored in TSS.

Approved on the basis of fixing the linked bug, but I'd appreciate clarification re: the thread-specific storage test.

rinon:
Heh, the test was actually exposing the issue I wanted to check for (thread-specific destructor dispatch doesn't switch to compartment 0 before calling indirect functions. But... I didn't run with `LIBIA2_DEBUG=1`. I'll remove that test for now, and file a follow-up issue on it.

Issue #235, opened 2023-06-29
Define policy for syscalls manipulating memory mappings


fw-immunant:
We need to control access to the following syscalls that manipulate memory mappings:

`mmap`, `mremap`, `munmap`, `mprotect`, `pkey_mprotect`

fw-immunant:
# 1. `mmap`

## Usage scenarios

We have at least (possibly, only) the following scenarios in which compartmentalized programs will use `mmap` (in roughly this order):

During program startup:
- in the dynamic loader mapping `ld.so.cache`
- in the dynamic loader for internal allocations
- in the dynamic loader when mapping segments of dynamic libraries
- in the dynamic loader allocating the TLS region

After program startup:
- inside the compartmentalized allocator to back allocations
- when allocating a thunk for our `pthread_create` wrapper
- directly in application or library code

## Threats

The threats posed by `mmap` to compartmentalization are:
1. `mmap` with the `MAP_FIXED` flag can overwrite existing mappings, changing their permissions and contents
2. newly-mapped memory can be accessed by any compartment, which can allow a compartment other than the one causing the allocation (for example, if running in another thread) to race to write into the newly-mapped region and corrupt it

## Mitigations

The latter of these cannot happen until after the program has started up enough to spawn new threads (i.e. main or constructors), so we don't need to guard against it when the dynamic loader is doing program startup.

To mitigate the first of these threats, we can forbid `MAP_FIXED` after program startup.

To mitigate the second, we need to:
- make sure the compartmentalized allocator follows a race-free sequence of an initial `mmap` with minimal permissions, followed by `mprotect`/`pkey_mprotect` to allow only the accesses we intend. If we can't avoid races here, we'll have to clear memory once we know it's in a state where only the desired compartment can write to it, prior to returning a pointer into the memory from the allocator.
- follow a similar protocol in the `pthread_create` wrapper
- wrap `mmap` calls from the application and libraries into a similar sequence of mapping and protection.

fw-immunant:
# 2. `mremap`

## Usage scenarios

The only usage scenarios I know of for `mremap` are:
1. inside the compartmentalized allocator to resize one of its existing mappings
2. in application or library code to resize an existing allocation

## Threats

[...]

Issue #231, opened 2023-06-28
Syscall categorization


fw-immunant:
Controlling system calls isn't as simple as filtering them on syscall identity, but identity is the first criterion we filter on. I've gone through x86_64 syscalls and categorized them on the basis of what I think we ultimately want/need to do to support them and how they interact with our sandbox.

In general, we will disallow forbidden syscalls with seccomp-bpf discriminating on the syscall number; syscalls that require more complex filtering will have some amount of argument inspection done directly by logic in seccomp-bpf and some by userspace helper.

In some cases we can give the userspace helper more latitude for inspecting arguments (e.g. those in buffers pointed at by syscall arguments) by wrapping syscalls in a shim that "freezes" these arguments by copying them to a read-only memory region. This allows us to avoid TOCTTOU by checking pointers against a known map of frozen memory regions and to racelessly inspect pointed-to data if it is within a frozen region.

I've attempted to categorize all syscalls for x86_64, but we should implement filtering as an allowlist prioritized based on the syscalls used by common programs, our tests, and nginx. I'll file a separate issue on syscall usage characterization and which subsets of [this table](https://docs.google.com/spreadsheets/d/1ZhXn6gU9cNBlIoNlQGyQlAXozXaMTq4x79gnUVbQPZA/edit#gid=0) to prioritize.

I was hoping to [inline the table](https://thisdavej.com/copy-table-in-excel-and-paste-as-a-markdown-table/) into a github comment, but it's too big to fit (>65535 characters).

Source for syscalls: https://syscalls.mebeim.net/?table=x86/64/x64/v6.3

rinon:
`open` can be dealt with by selinux, can we mix selinux and seccomp-bpf?

`brk` seems like it might be dangerous, we need to test how it interacts with existing mappings. In general it is deprecated anyway, so we may not need to allow it.

`shmat` is dangerous - I believe that can remap memory.

We can go through the whole list in more detail while allow-listing each that we need, these just stuck out to me.

fw-immunant:
> `open` can be dealt with by selinux, can we mix selinux and seccomp-bpf?

[...]

PR #290, opened 2023-09-25
Integrate runtime components into single runtime

fw-immunant:
This is not yet applied to our compartmentalized binaries and tests, but runs a tiny (uncompartmentalized) demo program I've been linking against it manually for validation.

This PR will be ready for merging when the build-system work is done to integrate the runtime with our tests themselves, but we'll want a switch to build without the runtime (or disable it) so that we can debug tests, as the runtime (being a `ptrace`r) conflicts with gdb and other debuggers.

This implements an initial coarse-grained syscall policy very similar to that described in #233, and integrates fine-grained syscall filtering as described in #235. As we validate this on our tests/demons/nginx we'll probably add syscalls to our allowlist to get closer to what #231 outlines.


fw-immunant:
One wart here is that we use `SECCOMP_FILTER_FLAG_NEW_LISTENER` to get a seccomp_unotify fd, but we don't actually use this fd for anything. If we need to do more detailed inspection of syscall arguments in memory or if we need to inject fds into the inferior to simulate some syscalls, we likely will need this fd; see `man 2 seccomp_unotify`. But for now, we're setting up this infrastructure without it being load-bearing.

runtime/seccomp_filter.c
```diff
@@ -79,6 +70,6 @@ int configure_seccomp(void) {
     perror("seccomp(SECCOMP_FILTER_FLAG_TSYNC)");
     return -1;
   }
-  // prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, &prog);
```
fw-immunant, line 85: This was debugging cruft; this form is deprecated and strictly less flexible compared to the standalone `seccomp()` syscall.

runtime/seccomp_filter_demo.c
```diff
@@ -0,0 +1,60 @@
...
+
+#include "seccomp_filter.h"
+
+int main(int argc, char *argv[]) {
+  int infd, outfd;
+  ssize_t read_bytes;
+  char buffer[1024];
+
+  if (argc < 3) {
+    printf("Usage:\n\tdup_file <input path> <output_path>\n");
```
rinon, line 15: dup_file?

runtime/track_memory_map.c
```diff
@@ -346,40 +359,51 @@ void return_syscall_eperm(pid_t pid) {
   fprintf(stderr, "wrote -eperm to rax\n");
 }
 
-void track_memory_map(pid_t pid, struct memory_map *map) {
+/* track the inferior process' memory map.
+
+returns true if the inferior exits, false on trace error.
+
+if true is returned, the inferior's exit status will be stored to *exit_status_out if not NULL. */
```
rinon, line 366: Probably want to put this in the header?

rinon (commented):
So as is the usage for this is to add the `ia2-main` library which wraps main to fork and run the binary? How does this interact with our existing wrapping of main to set the pkey?

Did you decide to do this rather than the external binary that runs a separate child with fork+exec? Or was this implementation before we discussed that?

runtime/CMakeLists.txt
```diff
@@ -26,11 +26,26 @@ add_executable(track-memory-map
...
 )
 
 add_executable(landlock
+  landlock_demo.c
   landlock.c
   strv.c
   forbid_paths.c
 )
+
+add_library(ia2-main
```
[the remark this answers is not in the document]
fw-immunant: Now called `ia2-sandbox`.

fw-immunant:
This is without the changes needed to support multiple threads and fork/exec; see #307.

PR #296, opened 2023-09-28
Mark the end of IA2 init

fw-immunant:
This lets us avoid errors when running programs with destructors under the memory-map tracking runtime.

I want to factor out the magic constant--where should it live?

### DIFF
--- a/runtime/track_memory_map.c
+++ b/runtime/track_memory_map.c
@@ -184,6 +184,15 @@ unsigned char pkey_for_pkru(uint32_t pkru) {
 #undef CHECK
 }
 
+/* Pass to mmap to signal end of program init */
+#define IA2_FINISH_INIT_MAGIC 0x1a21face1a21faceULL
+
+bool event_marks_init_finished(enum mmap_event event, const union event_info *event_info) {
+  return event == EVENT_MMAP &&
+         event_info->mmap.range.start == IA2_FINISH_INIT_MAGIC &&
+         event_info->mmap.flags & MAP_FIXED;
+}
+
 /* query pid to determine the mmap-relevant event being requested. returns true
  * unless something horrible happens */
 bool interpret_syscall(struct user_regs_struct *regs, unsigned char pkey,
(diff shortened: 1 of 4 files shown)


runtime/track_memory_map.c
```diff
@@ -52,6 +52,10 @@ bool is_op_permitted(struct memory_map *map, int event,
                                                       false);
     if (impacts_only_unprotected_memory)
       return true;
+    /* during init, we allow re-mprotecting memory, which we need to alter
+    initially-RO destructors */
+    else if (!memory_map_is_init_finished(map))
+      return true;
```
rinon, line 58: Don't we want to track ownership of things like stacks mappings? If I read this correctly we won't actually mark these regions allocated during init as owned by a particular compartment?

runtime/track_memory_map.c
```diff
@@ -388,6 +401,22 @@ void track_memory_map(pid_t pid, struct memory_map *map) {
       fprintf(stderr, "could not interpret syscall!\n");
     }
 
+    /* pick up signal marking IA2 init finished to start forbidding init-only operations */
+    if (event_marks_init_finished(event, &event_info)) {
+      if (!memory_map_mark_init_finished(map)) {
+        fprintf(stderr, "attempting to re-finish init! (rip=%p)\n", (void *)regs.rip);
+        return;
```
rinon, line 408: Do we want to exit with error or something here?

rinon (commented):
I don't personally mind the duplicated constant as long as you reference the other place it is defined. Alternatively, we could have an internal shared header for this that is included from the runtime? Can we include the whole internal header without picking up any actual code?

runtime/track_memory_map.c
fw-immunant: Yeah--in the runtime integration branch `track_memory_map` returns a `bool`. I'll take this into account when rebasing.

runtime/track_memory_map.c
fw-immunant: You're right that our memory map doesn't know about stack mappings and other "preexisting" mappings like the vDSO. If we want to fully mirror the kernel's idea of the process' memory map, we would need those, but I'm not 100% sure whether this is a security concern. Specifically, I guess an attack could involve unmapping the original stack and remapping other data in its place, which would then allow corrupting data used during process termination (post-`main` and dtors). But the original stack is also compartment zero anyway, right?

I'll punt on this for now but will file a bug.

fw-immunant: Note that allowing re-protecting memory during init here is not required *because* we don't know about stack mappings; these are separate concerns.

runtime/track_memory_map.c
fw-immunant: Done.

PR #298, opened 2023-10-03
Add heap compartmentalization to nginx

rinon:
Also adds a new RTMP streaming config for a more interesting demo.

### DIFF
--- a/external/nginx/src/core/ngx_buf.h
+++ b/external/nginx/src/core/ngx_buf.h
@@ -12,6 +12,7 @@
 #include <ngx_config.h>
 #include <ngx_core.h>
 
+#include <ia2.h>
 
 typedef void *            ngx_buf_tag_t;
 
@@ -121,6 +122,7 @@ typedef struct {
 
 #define NGX_CHAIN_ERROR     (ngx_chain_t *) NGX_ERROR
 
+IA2_BEGIN_NO_WRAP
 
 #define ngx_buf_in_memory(b)       ((b)->temporary || (b)->memory || (b)->mmap)
 #define ngx_buf_in_memory_only(b)  (ngx_buf_in_memory(b) && !(b)->in_file)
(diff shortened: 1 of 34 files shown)


rinon:
The bottom three commits are actually part of #288, I just don't want to rebase twice. This will need to wait to land until #288 is in.

external/nginx/src/os/unix/ngx_alloc.h
```diff
@@ -9,10 +9,11 @@
 #define _NGX_ALLOC_H_INCLUDED_
 
 
-#include <ia2.h>
 #include <ngx_config.h>
 #include <ngx_core.h>
 
+#include <ia2.h>
+
```
ayrtonm, line 16: Why does #include order in this commit matter?

external/nginx/src/core/ngx_buf.h
```diff
@@ -152,7 +154,10 @@ ngx_chain_t *ngx_alloc_chain_link(ngx_pool_t *pool);
 
 
 ngx_int_t ngx_output_chain(ngx_output_chain_ctx_t *ctx, ngx_chain_t *in);
+IA2_END_NO_WRAP
+/* ngx_chain_writer is used as a callback, so must be wrapped */
```
ayrtonm, line 158: Callbacks don't necessarily have to be wrapped, it would just run with pkey 0 if it wasn't. Does `ngx_chain_writer` have to have access to nginx's compartment?

ayrtonm (approved):
The allocator changes themselves LGTM.

external/nginx/src/core/ngx_buf.h
rinon: Yeah, it calls some nginx logging functions, which aren't wrapped because they're in the same library. At least iirc that was the issue. It definitely needed to be wrapped.

external/nginx/src/os/unix/ngx_alloc.h
rinon: When IA2 is disabled we now don't set _GNU_SOURCE but we include headers. Once a glibc header has been included, the configuration of *_SOURCE is then fixed. We could skip including the headers if IA2 is disabled?

partition-alloc/src/get_pkey.cc
```diff
@@ -0,0 +1,77 @@
...
+  case 0xF3FFFFFC: {
+    return 13;
+  }
+  case 0xCFFFFFFC: {
+    return 14;
+  }
+  case 0x3FFFFFFC: {
+    return 15;
+  }
+  // TODO: We currently treat any unexpected PKRU value as pkey 0 (the shared
```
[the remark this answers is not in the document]
rinon: Technically the kernel has a default pkey value which is globally configurable? We could hard code the expected value to zero, that would catch any weirdness early which is good.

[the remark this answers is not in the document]
rinon: Still though, I think that should be a followup.

PR #300, opened 2023-10-10
Add support for more than two compartments

### LINKED ISSUE #276: Add support for more than two compartments
This is non-trivial because each call gate is specific to a pair of compartments. This means when compartment A and B both call `foo` in C we need separate call gates for them. Currently we insert call gates using `ld --wrap` which requires the interposed symbols to be globally unique so we'll have to rename these symbols (which can be done after compiling but before linking). See #174 for the preliminary plan.

ayrtonm:
Right now this just separates functions that need direct call gates based on whether they'll need one or more call gates, modifies `define_shared_lib` to allow invoking it more than once per tests and updates `read_config` to use 3 compartments. The expected behavior is an ld error when building the `read_config_main_wrapped` target since `define_test` doesn't pull in the third compartment DSO yet. Closes #276.


ayrtonm:
See [this comment](https://github.com/immunant/IA2-Phase2/issues/276#issuecomment-1756158945) for latest status.

ayrtonm:
CI is failing because I added a dummy `.c` to test out the objcopy step with more than one `.o`, but tests pass locally.

ayrtonm:
TODO: bump CMake req to 3.27 for LIST JOIN in generator expression

rinon (commented):
cmake changes look alright to me after a quick pass

ayrtonm:
rebased on main but I think I dropped the nginx commit. The three pkeys test builds but I didn't get a chance to run it on an MPK machine

cmake/ia2.cmake
```diff
@@ -309,12 +320,15 @@ function(add_ia2_call_gates NAME)
...
+    DEPENDS ${CALL_GATE_SRC} ${CALL_GATE_HDR}
+            ${LD_ARGS_FILES} ${OBJCOPY_ARGS_FILES} ${REWRITTEN_SOURCES})
 
   add_library(${CALL_GATE_TARGET} SHARED ${CALL_GATE_SRC})
   add_dependencies(${CALL_GATE_TARGET} ${NAME}-rewrite)
+  # This is needed to enable LIBIA2_DEBUG=1
+  target_compile_definitions(${CALL_GATE_TARGET} PRIVATE IA2_ENABLE=1)
   if(LIBIA2_DEBUG)
-    target_compile_definitions(${CALL_GATE_TARGET} PRIVATE LIBIA2_DEBUG=1)
+    target_compile_definitions(${CALL_GATE_TARGET} PRIVATE LIBIA2_DEBUG=1 LIBIA2_ENABLE=1)
```
fw-immunant, line 331: I don't see any other mentions of `LIBIA2_ENABLE` (with "LIB") in the codebase--is this actually needed?

ayrtonm: ah I forgot to commit the typo fix, it should be IA2_ENABLE

ayrtonm:
I'm not currently working on this @fw-immunant in case you have time to take a look at the test failure or make changes to this PR to minimize conflicts with the library-only mode stuff. Also the last two commits can probably be split off and merged independently though we should double check that the `ignore_function` doesn't break anything.

fw-immunant:
Rebased now that #349 has landed.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -145,7 +149,7 @@ static bool ignore_file(const Filename &filename) {
   if (is_empty) {
     return false;
   }
-  return !filename.starts_with(RootDirectory);
+  return !filename.starts_with(OutputDirectory) && !filename.starts_with(RootDirectory);
```
ayrtonm, line 152: I agree my change was wrong, but I think this still mishandles the case of root directory/not output directory. We shouldn't encounter this case because it means we're either trying to annotate an original source file in-place or picking up one of the original (non-system) headers. I think it'd be clearer to just ignore files not in the output directory and assert that if it's not in the output directory it's also not in the root directory.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -103,6 +103,10 @@ static Filename get_filename(const clang::SourceLocation loc,
       abort();
     }
   }
+  // if we could not get a filename for the spelling loc, use the expansion filename
```
ayrtonm, line 106: Would it make more sense to explicitly use `get_expansion_filename` at the callsites instead of silently falling back to the expansion loc? I haven't looked at the uses in detail enough to know which is cleaner.

tools/rewriter/SourceRewriter.cpp
fw-immunant: You're right, done. I had thought the `RootDirectory` check was necessary but in testing it is never encountered and as you observe shouldn't be. I've added an error if we do somehow encounter that case in the future.

PR #306, opened 2023-10-31
Make lit optional

sim-immunant:
Make lit opt-in rather than opt-out, and remove segment checks.


fw-immunant (approved):
LGTM. I'd prefer this as one commit since the scandir change is just a fix to new code.

ayrtonm (approved):

PR #307, opened 2023-11-01
Support clone()/fork()/exec() in runtime

fw-immunant:
We need `clone()` to sandbox multithreaded programs; `fork()`/`exec()` support is needed for Criterion tests as well as any sandboxed program that wants to spawn a child.

This raises some questions: seccomp filters are *necessarily* inherited, and if our ptrace() tracer doesn't attend to forked child processes, they will see any `SECCOMP_RET_TRACE` syscalls as forbidden and will generally not work. As such, our tracer must be responsible for *all* child processes, even if they aren't actually sandboxed with IA2.

However, we also have the situation (with Criterion) where the child process *is* sandboxed with IA2. For now, we simply try to treat all child processes like IA2-sandboxed ones.

~This is a draft because it's still pretty messy as it's where things stood when I gave up on making Criterion work (for some reason one of its child processes gets ENOENT when trying to open a shm object that was created by another process... the shm object does indeed not exist, but I don't know why this is the case under our sandbox and not otherwise).~

I've gotten this to work with most of our tests, the exceptions being `threads`, `protected_threads`, and `permissive_mode`. The largest problem was that we were not handing ptrace "group-stop" correctly, and we couldn't until switching from `PTRACE_TRACEME` to `PTRACE_SEIZE`.


cmake/define-test.cmake
```diff
@@ -142,7 +142,7 @@ endfunction()
 #                  cmake test infrastructure.
 function(define_test)
     # Parse options
-    set(options NEEDS_LD_WRAP NOT_IN_CHECK_IA2 NO_LIBS CRITERION_TEST)
+    set(options NEEDS_LD_WRAP NOT_IN_CHECK_IA2 NO_LIBS CRITERION_TEST WITHOUT_RUNTIME)
```
rinon, line 145: Please add to doc string

rinon: Also, what do you think about `WITHOUT_SANDBOX`?

runtime/seccomp_filter.c
```diff
@@ -47,22 +48,71 @@ struct sock_filter ia2_filter[] = {
...
+    BPF_SYSCALL_POLICY(geteuid, ALLOW),
+    BPF_SYSCALL_POLICY(getgid, ALLOW),
+    BPF_SYSCALL_POLICY(getuid, ALLOW),
+    BPF_SYSCALL_POLICY(getppid, ALLOW),
+    BPF_SYSCALL_POLICY(setpgid, ALLOW), // used by criterion
+    BPF_SYSCALL_POLICY(gettid, ALLOW),
     BPF_SYSCALL_POLICY(getpid, ALLOW),
     BPF_SYSCALL_POLICY(getrandom, ALLOW),
+    BPF_SYSCALL_POLICY(prctl, ALLOW),
+    BPF_SYSCALL_POLICY(ioctl, ALLOW),
```
rinon, line 67: `ioctl` and friends deserves a comment at least, we can't just allow all ioctls through, can we? Or are we sandboxing what devices can be opened? Not saying we should do this now, just don't want to forget.

rinon (approved):

runtime/seccomp_filter.c
fw-immunant: Good point; this is needed by Criterion but we can fairly easily limit it to only the one `ioctl` that Criterion actually performs.

fw-immunant: Done in b449ff0.

cmake/define-test.cmake
fw-immunant: Done in d542238.

PR #309, opened 2023-11-15
Rewrite cmake files to simplify build system

rinon:
Replaces the `define_test` and `define_shared_lib` functions with a more generic and hopefully cleaner `add_ia2_compartment` function.


fw-immunant (commented):
Mostly looks good, but I don't understand the second commit. Were we padding the TLS segment of partition-alloc before? I don't think we should need to do this, because we don't `pkey_mprotect` its thread locals as it isn't itself "in" any compartment. That said, I don't think I thought hard about the interaction of TLS usage *in* partition-alloc with our TLS scheme; I don't think I had really realized it was using TLS at all.

It does, though, for `partition_alloc::(anonymous namespace)::g_disallow_allocations`, `partition_alloc::internal::(anonymous namespace)::ReentrantScannerGuard::guard_`, `partition_alloc::internal::g_thread_cache`, `partition_alloc::internal::base::(anonymous namespace)::g_thread_id`, and `partition_alloc::internal::base::(anonymous namespace)::g_is_main_thread`.

cmake/ia2.cmake
```diff
@@ -0,0 +1,272 @@
...
+  set_property(TARGET ${NAME} PROPERTY ORIGINAL_INCLUDE_DIRECTORIES ${ARG_INCLUDE_DIRECTORIES})
+
+  if (ARG_ENABLE_UBSAN)
+    # UBSAN requires passing this as both a compiler and linker flag
+    target_compile_options(${NAME} PRIVATE "-fsanitize=undefined")
+    target_link_options(${NAME} PRIVATE "-fsanitize=undefined")
+  endif()
+
+  target_link_libraries(${NAME} PRIVATE dl libia2 partition-alloc)
+  target_link_options(${NAME} PRIVATE "-Wl,--export-dynamic")
```
ayrtonm, line 61: We previously only had this flag for the main executable, though I think for shared libs it's a no-op anyway(?), so it's not a big deal.

partition-alloc/CMakeLists.txt
```diff
@@ -65,37 +65,43 @@ set(PA_SRCS
...
 list(TRANSFORM PA_SRCS PREPEND ${EXTERNAL_DIR}/chromium/src/base/allocator/partition_allocator/)
 
-add_library(partition-alloc SHARED
+add_library(partition-alloc_unpadded SHARED
 	src/allocator_shim_default_dispatch_to_partition_alloc.cc
 	src/get_pkey.cc
 	src/shared_allocator.cc
     ${PA_SRCS})
 
+pad_tls_library(partition-alloc_unpadded partition-alloc)
```
ayrtonm, line 74: This creates a target named `partition-alloc-padding` but nginx tries to build a CMake target named `partition-alloc` which is why CI is failing. Since the `INPUT` parameter to `pad_tls_library` is usually `$X_unpadded` it might make sense to name the target for the padded lib exactly like the `OUTPUT` parameter instead of with the `-padding` suffix.

ayrtonm (changes requested):
LGTM with mostly minor nits. There's just one case we need to fix where the `-include` flag is wrong and there's an issue with the partition alloc target that nginx tries to build.

ayrtonm:
@rinon I saw you effectively removed the `NEEDS_LD_WRAP` from the new cmake functions and I think passing an ld args file unconditionally is fine here. For the objcopy step added in #300 (that's implemented in a similar way) were you also planning on doing that unconditionally or keeping `NEEDS_OBJCOPY_FILE`?

ayrtonm:
I'm rebasing #300 on this PR so for now I'll leave `NEEDS_OBJCOPY` as a no-op arg in the original cmake functions (like `NEEDS_LD_WRAP`) and do it unconditionally (or just for tests with more than two pkeys) in the new `add_ia2_compartment`. Then we can figure out how to check if it's necessary (maybe building the original targets and looking for duplicate symbols with `nm`).

partition-alloc/CMakeLists.txt
ayrtonm: actually I just noticed that `pad_tls_library` should create a target named `OUTPUT` so I'm not entirely sure what's going on here

cmake/ia2.cmake
rinon: Should be fine, yeah.

partition-alloc/CMakeLists.txt
rinon: It also creates a library target of `${OUTPUT}`, so I'm not sure what the problem is exactly. I'll investigate.

rinon: I think my issue was that I needed a separate target to use `add_library` because I can't `add_library` for a custom command without a custom target in between.

rinon: I worked around this in 1ce7fba8f

ayrtonm (approved):

ayrtonm:
LGTM. Could you update the target name for PartitionAlloc in docs/build_instructions.md?

PR #311, opened 2023-12-12
Add initial support for ARM MTE

ayrtonm:
This PR adds `IA2_USE_MPK` and `IA2_USE_MTE` macros to libia2 and replaces calls to `pkey_mprotect` with an arch-independent function for tagging loaded writable ELF segments with a memory key. On ARM this uses `addg` to increment a pointer through the range of pages passed to `mprotect` and `st2g` to tag the memory.

This should already pass tests because the `#ifdef`s used in CI ignore ARM for now, but I'd like to at least build libia2 for ARM in CI before merging this.

edit: nginx build failure may currently be expected(?)


ayrtonm:
Looks like CI works as expected. Also I think the `addg`s in `insert_tag` may be a security issue but I need to read a bit more about what kind of CFI we expect to require and how it'll be implemented.

ayrtonm:
I'm still trying to figure out if the `addg`s are a security issue but either way I think this is lower priority than setting up CI for arm. As a first step we could just build libia2 for aarch64 on donna. I'm not sure if there's a clean way to build for two archs with one cmake config step so we might just do one config per arch.

ayrtonm:
I got libia2 building in CI with a cross-compiler for aarch64 on x86. I see the preprocessor warnings I added for `IA2_USE_MTE` so it is building for MTE, but we might want to test a bit more thoroughly even before the ARM call gates are ready. I'm still trying to think if there's a good intermediate test between this and running a compartmentalized program in user-mode qemu though.

libia2/CMakeLists.txt
```diff
@@ -3,6 +3,12 @@ project(libia2)
 
 add_library(libia2 ia2.c threads.c main.c exit.c)
 target_compile_options(libia2 PRIVATE "-fPIC")
+if(LIBIA2_MTE)
+    target_compile_definitions(libia2 PUBLIC IA2_USE_MTE=1)
+    target_compile_options(libia2 PUBLIC "-march=armv8.5-a+memtag" "-ffixed-x18")
```
sim-immunant, line 8: We should take a closer look at the behavior of `-ffixed-x18`. In my armtest repo, even when compiling with that flag, X18 still seems to get saved and restored between function calls. System libraries also frequently alter x18, so we'll have to compile the whole shebang from libc up with this flag as well.

libia2/main.c
```diff
@@ -41,3 +42,4 @@ __attribute__((naked)) int __wrap_main(int argc, char **argv) {
       /* clang-format on */
       ::);
 }
```
ayrtonm, line 44: TODO: Add todo for wrap main

ayrtonm:
The issue with CMake auto-adding `-I/usr/include` to tests was caused by LLVM include dirs and fixed by replacing `include_directories` in `rewriter/CMakeLists.txt` with the usual `target_include_directories`. The whole repo doesn't really support cross-compiling correctly because CMake makes it a pain to use more than one toolchain and the "just set CMAKE_C_COMPILER" hack I used for pad-tls won't work for the rewriter, so instead we need to use [`ExternalProject_Add`](https://cmake.org/cmake/help/latest/module/ExternalProject.html) for host tools.

ayrtonm:
Remaining things to merge this PR are
1. set up basic CI for Aarch64 that just builds the no-criterion minimal test and runs it in QEMU. The current issue here is that the rewriter seems to not recognize our custom command line arguments (?) though it does suggest `-p`. This is probably differences between the version of LLVM/libclang ubuntu uses and the way that cl opts are declared. 
2. It would be nice to keep the ia2-rewriter and pad-tls targets building as-is. Right now we have to build a tools target which builds both. This isn't an issue for ia2-rewriter since pad-tls is fairly small, but it does significantly increase build time when all you want to build is pad-tls.

ayrtonm:
> set up basic CI for Aarch64 ...

actually I just dropped the commit adding the cl opt during the PR reorg...

ayrtonm:
Turns out the issue was that the libclang/LLVM version wasn't being propagated from the top-level CMake project to the tools project so the rewriter was using an older version which didn't support that way of declaring command-line arguments.

ayrtonm:
Seems that the libia2 refactoring in the first commit broke the way the call gates track the stack pointers. This was already sketchy since we were declaring a variable with different types in two different places as noted in the last commit, but stepping through in gdb shows that the stack pointer management definitely broke in this PR.

ayrtonm:
I might split the libia2 refactoring out into a separate PR before landing this.

ayrtonm:
I simplified the libia2 refactoring and it seems the stack pointer management is fixed now and the x86 CI failures are the spurious test failures mentioned in #324. This is also rebased on #309 so we can merge this as soon as that lands.

ayrtonm:
I think the spurious CI failures aren't actually related to #324 since the error message is
```
23/25 Test #23: rewrite_fn_ptr_eq ................***Failed    0.01 sec
Failed to set up path allowlist
PTRACE_CONT: No such process
could not spawn child process
```
for various tests. I think the issue is caused by running the tests through ctest (which I assumes forks to run the tracer which forks again to run each test), but I haven't been able to reproduce it manually on donna.

ayrtonm:
I added a hack for the procfs race issue and now the CI failures look like #324 so I'll look into that.

ayrtonm:
The transient mmap error is probably going to be hard to reproduce manually for [this same reason](https://github.com/immunant/IA2-Phase2/pull/311#issuecomment-2008249319) so I just added a function to print /proc/self/maps to stderr when this happens.

PR #318, opened 2024-01-22
Trace `madvise` syscall

fw-immunant:
`madvise(MADV_DONTNEED)` is equivalent to zeroing memory; we cannot allow this to be performed cross-compartment.

Because some other 'advices' may also have similar effects, for now we simply forbid all cross-compartment `madvise(advice)` calls regardless of the value of `advice`. We can loosen this policy if needed by some applications down the line, but for now this patches the security hole while keeping our tests passing.

#### Postscript

This took a few iterations to get tests green again. Round-by-round:

1. The first commit failed tests for a couple reasons, the primary of which is that the `madvise()` policy was being tripped and forbidding a `madvise(MADV_DONTNEED)` inside `free()` in PartitionAlloc. To debug this, I added the next few commits, through `runtime/track_memory_map: print details of forbidden operations`.
2. At this point, I had enough information to see that for some reason the tracer observed the program running inside compartment zero (pkru==0x00000000) the whole time. This led to the realization that the runtime could not properly read the pkru on the machine running our CI.
3. I added debugging to avoid using the pkru being seen as zero in some cases where our logic to read it via `ptrace(PTRACE_GETREGSET)`might fail. This immediately triggered, indicating that this is where the problem was (`runtime/get_inferior_pkru: initialize buffer with noticeable garbage`).
4. Investigating this further, I discovered that the approach that we (and gdb) had been using to read the pkru was non-portable. gdb has since [changed to a more portable approach](https://git.linaro.org/toolchain/binutils-gdb.git/commit/?id=c689d1fe58b2c0faf51e4f574d50271f1d0648e3), and I rewrote our code to do as the Intel documentation suggests, which should be future-proof (`runtime/get_inferior_pkru: query CPUID to determine the location of PKRU in xstate`).
5. Validating this change, I saw tests passing despite being killed by seccomp for syscall policy violation. I realized that my reworking to handle tracing of multiple threads and processes was masking the propagation of failure from processes being killed by fatal signals.
6. I fixed the masking bug and found that tests were using `open()` which is not on our syscall allowlist. I added it to the list as this was a mere oversight, caused by the fact that our tests do not use `open()` (preferring `openat()`) except via ubsan.
7. CI still failed due to a probabilistic abort caused by racy access to the filesystem when setting up our landlock FS sandbox. I fixed the error handling here (it's benign if we happen to be unable to allow access to a directory because it no longer exists) and now finally everything is green.

### DIFF
--- a/runtime/memory-map/src/lib.rs
+++ b/runtime/memory-map/src/lib.rs
@@ -337,6 +337,28 @@ pub extern "C" fn memory_map_region_get_prot(map: &MemoryMap, needle: Range) -> 
 /** memory_map_region_get_prot found no or multiple protections in the given range */
 pub const PROT_INDETERMINATE: u32 = 0xFFFFFFFFu32;
 
+const PKEY_MULTIPLE: u8 = 255;
+const PKEY_NONE: u8 = 254;
+
+/// return the pkey (0-15) that covers the given range, or PKEY_MULTIPLE or
+/// PKEY_NONE if the entire range is not protected by exactly one pkey
+#[no_mangle]
+pub extern "C" fn memory_map_region_get_owner_pkey(map: &MemoryMap, needle: Range) -> u8 {
+    let mut pkey = None;
+    let same = map.all_overlapping_regions(needle, |region| match pkey {
+        None => {
+            pkey = Some(region.state.owner_pkey);
+            true
+        }
+        Some(pkey) => region.state.owner_pkey == pkey,
+    });
+    if same {
+        pkey.unwrap_or(PKEY_NONE) // all same (return pkey) or no pkey
+    } else {
+        PKEY_MULTIPLE // multiple pkeys
+    }
+}
+
 #[no_mangle]
 pub extern "C" fn memory_map_unmap_region(map: &mut MemoryMap, needle: Range) -> bool {
     map.split_out_region(needle).is_some()
--- a/runtime/forbid_paths.c
+++ b/runtime/forbid_paths.c
@@ -21,7 +21,8 @@ static int allow_path(const char *path, const int ruleset_fd, bool shallow) {
   struct stat statbuf;
 
   path_beneath.parent_fd = open(path, O_PATH | O_CLOEXEC);
-  if (path_beneath.parent_fd < 0) {
+  /* error on inability to open dir, unless it simply no longer exists */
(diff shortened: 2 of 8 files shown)


rinon (approved):

runtime/memory-map/src/lib.rs
```diff
@@ -337,6 +337,28 @@ pub extern "C" fn memory_map_region_get_prot(map: &MemoryMap, needle: Range) ->
 /** memory_map_region_get_prot found no or multiple protections in the given range */
 pub const PROT_INDETERMINATE: u32 = 0xFFFFFFFFu32;
 
+const PKEY_MULTIPLE: u8 = 255;
```
ayrtonm, line 340: Do we need to keep these in sync with anything e.g. some libc function return value or are they arbitrary?

ayrtonm:
LGTM. The issue fixed by the last commit is a race where a process forks, then the tracer tries calls `forbid_paths` to restrict the child from opening `/proc/self/mem` and the child terminates before the tracer can open `/proc/$PID/mem`, right?

ayrtonm (approved):

runtime/memory-map/src/lib.rs
fw-immunant: `PROT_INDETERMINATE` just needs to be distinguishable from any valid combination of `PROT_` flags, and `PKEY_MULTIPLE` just needs to be outside the range of valid pkeys (0-15). They need to be synchronized between the Rust (this file) and the C usage, but cbindgen does that for us.

fw-immunant:
> LGTM. The issue fixed by the last commit is a race where a process forks, then the tracer tries calls `forbid_paths` to restrict the child from opening `/proc/self/mem` and the child terminates before the tracer can open `/proc/$PID/mem`, right?

The problem can come up with *any* transient process that exists when `nftw` reads entries from `/proc` and exits before we allowlist its process directory. It isn't a security problem because the worst-case is that we fail to allowlist the procfs dir of a transient process.

ayrtonm:
> It isn't a security problem because the worst-case is that we fail to allowlist the procfs dir of a transient process.

Ah ok so in the window before the tracer initializes the allowlist, the tracee is just running w/o access to procfs. That seems fine.

Issue #321, opened 2024-02-29
Consider reorganizing repo directories


ayrtonm:
CMake made the fundamental design choice to only support one toolchain per invocation which makes it slightly more annoying to add support for ARM. The issue is that we need to compile both arch-specific code and host tools. Luckily we have no mixed-arch directories but we do have places where nested subdirectories have different archs (e.g. repo root is native arch, `rewriter` is host, `rewriter/tests` is native). I already added support for this in #311 but since that PR already splits the config step into multiple CMake invocations using `ExternalProject_Add` I thought it would make sense to reorganize things a bit by arch before merging any ARM stuff anyway.

Also I'm updating the README/docs with an overview of the repo directory structure and I realized that the repo is a bit confusing to navigate. I propose reorganizing it as follows

```
cmake - no changes
docs - no changes
external - no changes
examples - no changes
framework (new top-level dir)
  - libia2 (previously top-level libia2 dir)
  - partition-alloc (previously top-level partition-alloc dir)
     - move contents of scripts/partition_alloc here and remove top-level scripts dir
     - this is still just our shim for PA, all the third-party code should remain in external/chromium
  - process-tracer (previously top-level runtime dir)
tests (previously rewriter/tests dir)
tools (new top-level dir)
  - rewriter (previously top-level rewriter dir)
  - pad-tls (previously top-level pad-tls dir)
```

I'm very open to naming suggestions for the `framework` and `process-tracer` directories (and even `libia2` but that would require more changes so I think we can keep it as-is for now). The idea is that `framework` (maybe `sandbox`?) includes anything that is part of a compartmentalized application (both in the process and for tracing it). Also I think we should rename `runtime` since we've sometimes used it to refer to things that are part of `libia2.a` or the call gates and I think some variation of "tracer" would be more descriptive.

ayrtonm:
I could also see a case for naming the top-level `framework` dir `lib` instead (though it'd be a bit unexpected to have the tracer stuff there in that case). It might also make sense to move the things that are part of the compartmentalization "API" (e.g. `IA2_FN`, `IA2_CALL`, etc.) to a top-level `include` directory to make it more easily discoverable.

ayrtonm:
TODO: framework -> runtime, no top level include dir

ayrtonm:
Fixed by #322

PR #322, opened 2024-03-01
Update docs and reorganize repo directory structure

ayrtonm:
This PR makes most of the changes initially proposed in #321 and updates the docs. The directory reorganization took pretty minimal work so I'm open to suggestions about any changes that would make things more easily discoverable.


ayrtonm:
Still need to finish updating docs but I've made all the directory changes and matching CMake fixes

ayrtonm:
Pushed a potential fix from the last commit of PR #318. I'm not sure if ignoring the race condition is the right approach, but I still need to get a better understanding of the problem.

runtime/CMakeLists.txt
```diff
@@ -1,50 +1,6 @@
 cmake_minimum_required(VERSION 3.12)
-project(IA2Runtime)
+project(framework)
```
fw-immunant, line 2: Should this be `runtime`?

docs/directory_structure.md
```diff
@@ -0,0 +1,59 @@
...
+suite)
+
+
+## external
+Third-party code. These are git subtrees of upstream codebases and each commit
+may only modify one subtree or the main repository.
+
+#### chromium
+The subset of chromium needed to build PartitionAlloc. See
+framework/partition-alloc for subtree's latest upstream commit and details on
```
fw-immunant, line 22: `runtime/partition-alloc`

docs/directory_structure.md
```diff
@@ -0,0 +1,59 @@
...
+
+#### libia2
+The main compartmentalization library. Includes code for interposing libc and
+the API for compartmentalizing codebases where manual changes are required.
+
+#### partition-alloc
+A compartment-aware shim for PartitionAlloc.
+
+#### tracer
+TODO:
```
fw-immunant, line 45: The implementation of a separate process that tracks compartment memory ownership and enforces system call limitations to prevent compartment escape.

fw-immunant:
TODOs notwithstanding, looks good to me. This is a much more sane layout.

docs/directory_structure.md
ayrtonm: Thanks!

ayrtonm:
Decided with @rinon to merge this before #309 (and as a result #300). Also #311 still needs a lot of cleanup so we should also merge this before that (and before #323 as a result).

ayrtonm:
Tests are failing for the same reason as #324 but the docs are in a good enough state to merge. Note that the new docs are not comprehensive, but better than before.

PR #323, opened 2024-03-05
Generate no-op call gates for ARM64

sim-immunant:
Merge in my branch with the no-op call gates and call gates refactoring for multi-arch.

WIP, still working on making sure this is the correct assembly, but the `minimal_arm_main_wrapped` target compiles.

### DIFF
--- a/tools/rewriter/GenCallAsm.cpp
+++ b/tools/rewriter/GenCallAsm.cpp
@@ -442,8 +442,7 @@ static void emit_set_pkru(AsmWriter &aw, uint32_t target_pkey, Arch arch) {
     add_asm_line(aw, "movq %r11, %rdx");
   } else if (arch == Arch::Aarch64) {
     // set X18 to the pointer key (compartment number left-shifted 56 bits)
-    add_asm_line(aw, "mov x18, #" + std::to_string(target_pkey & 0xF));
-    add_asm_line(aw, "lsl x18, x18, #56");
+    llvm::errs() << "TODO x18 switching is not implemented\n";
   }
 }
 
--- a/tools/rewriter/GenCallAsm.cpp
+++ b/tools/rewriter/GenCallAsm.cpp
@@ -419,7 +419,7 @@ static void emit_copy_args(AsmWriter &aw, size_t stack_return_size, size_t stack
   }
 }
 
-static void emit_fn_ptr(AsmWriter &aw, Arch arch) {
+static void emit_load_fn_ptr(AsmWriter &aw, Arch arch) {
   if (arch == Arch::X86) {
       add_asm_line(aw, "movq ia2_fn_ptr@GOTPCREL(%rip), %r12");
       add_asm_line(aw, "movq (%r12), %r12");
@@ -689,11 +689,11 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
   emit_copy_args(aw, stack_return_size, stack_return_padding, stack_alignment, stack_arg_count, stack_arg_size, caller_pkey, arch);
 
   if (reg_arg_count > 0) {
-    emit_scrub_regs(aw, target_pkey, return_locs, arch);
+    emit_scrub_regs(aw, caller_pkey, param_locs, arch);
   }
 
   if (kind == WrapperKind::IndirectCallsite) {
-    emit_fn_ptr(aw, arch);
+    emit_load_fn_ptr(aw, arch);
   }
 
   emit_set_pkru(aw, target_pkey, arch);
--- a/tools/rewriter/GenCallAsm.cpp
+++ b/tools/rewriter/GenCallAsm.cpp
@@ -421,8 +421,8 @@ static void emit_copy_args(AsmWriter &aw, size_t stack_return_size, size_t stack
 
 static void emit_load_fn_ptr(AsmWriter &aw, Arch arch) {
   if (arch == Arch::X86) {
-      add_asm_line(aw, "movq ia2_fn_ptr@GOTPCREL(%rip), %r12");
-      add_asm_line(aw, "movq (%r12), %r12");
+    add_asm_line(aw, "movq ia2_fn_ptr@GOTPCREL(%rip), %r12");
+    add_asm_line(aw, "movq (%r12), %r12");
   } else if (arch == Arch::Aarch64) {
     // TODO ARM function pointer save
     llvm::errs() << "TODO indirect calls not implemented on ARM\n";
@@ -496,7 +496,7 @@ static void emit_copy_stack_returns(AsmWriter &aw, size_t stack_return_size, siz
   }
 }
 
-static void emit_scrub_regs(AsmWriter &aw, uint32_t pkey, const std::vector<ParamLocation> &locs, Arch arch) {
+static void emit_scrub_regs(AsmWriter &aw, uint32_t pkey, const std::vector<ParamLocation> &locs, bool preserve_regs, Arch arch) {
   // Handles register scrubbing for calls into/out of protected compartments.
   // After call: Preserves return values by pushing them to the stack before scrubbing.
   // Before call: Saves argument registers in a similar manner.
@@ -505,11 +505,13 @@ static void emit_scrub_regs(AsmWriter &aw, uint32_t pkey, const std::vector<Para
       // Zero out all unused registers. First we save all registers containing args.
       // These pushes have matching pops before calling the wrapped function so this
       // stack space is not shown in the diagram above.
-      add_comment_line(aw, "Preserve essential regs on stack");
+      if (preserve_regs) {
+        add_comment_line(aw, "Preserve essential regs on stack");
 
-      for (auto loc : locs) {
-        if (!loc.is_stack()) {
-          emit_reg_push(aw, loc);
+        for (auto loc : locs) {
+          if (!loc.is_stack()) {
+            emit_reg_push(aw, loc);
+          }
         }
       }
 
@@ -519,11 +521,13 @@ static void emit_scrub_regs(AsmWriter &aw, uint32_t pkey, const std::vector<Para
       add_comment_line(aw, "Scrub non-essential regs");
       add_asm_line(aw, "call __libia2_scrub_registers");
 
-      // Restore saved regs
-      add_comment_line(aw, "Restore preserved regs");
-      for (auto loc = locs.rbegin(); loc != locs.rend(); loc++) {
-        if (!loc->is_stack()) {
-          emit_reg_pop(aw, *loc);
+      if (preserve_regs) {
+        // Restore saved regs
+        add_comment_line(aw, "Restore preserved regs");
+        for (auto loc = locs.rbegin(); loc != locs.rend(); loc++) {
+          if (!loc->is_stack()) {
+            emit_reg_pop(aw, *loc);
+          }
         }
       }
     }
@@ -688,9 +692,7 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
 
   emit_copy_args(aw, stack_return_size, stack_return_padding, stack_alignment, stack_arg_count, stack_arg_size, caller_pkey, arch);
 
-  if (reg_arg_count > 0) {
-    emit_scrub_regs(aw, caller_pkey, param_locs, arch);
-  }
+  emit_scrub_regs(aw, caller_pkey, param_locs, reg_arg_count > 0, arch);
 
   if (kind == WrapperKind::IndirectCallsite) {
     emit_load_fn_ptr(aw, arch);
@@ -708,7 +710,7 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
 
   emit_switch_stacks(aw, target_pkey, caller_pkey, arch);
 
-  emit_scrub_regs(aw, target_pkey, return_locs, arch);
+  emit_scrub_regs(aw, target_pkey, return_locs, true, arch);
 
   emit_set_return_pkru(aw, caller_pkey, arch);
 
--- a/tools/rewriter/GenCallAsm.cpp
+++ b/tools/rewriter/GenCallAsm.cpp
@@ -462,8 +462,9 @@ static void emit_set_pkru(AsmWriter &aw, uint32_t target_pkey, WrapperKind kind,
     add_asm_line(aw, "movq %r10, %rcx");
     add_asm_line(aw, "movq %r11, %rdx");
   } else if (arch == Arch::Aarch64) {
-    // TODO ARM set compartment
-    llvm::errs() << "TODO compartment setting not implemented on ARM\n";
+    // set X18 to the pointer key (compartment number left-shifted 56 bits)
+    add_asm_line(aw, "mov x18, #" + std::to_string(target_pkey & 0xF));
+    add_asm_line(aw, "lsl x18, x18, #56");
   }
 }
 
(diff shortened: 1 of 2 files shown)


rewriter/GenCallAsm.cpp
```diff
@@ -280,21 +280,309 @@ static std::string sig_string(const CAbiSignature &sig,
...
+static void call_string(
+    const std::optional<std::string> target_name,
+    WrapperKind kind,
+    AsmWriter &aw,
+    Arch arch) {
+
+  add_comment_line(aw, "Call wrapped function");
+
+  if (arch == Arch::X86) {
+    if (!target_name) {
```
ayrtonm, line 292: This check and the following assert are duplicated. Couldn't you pull them out of the arch checks to avoid this?

rewriter/GenCallAsm.cpp
```diff
@@ -280,21 +280,309 @@ static std::string sig_string(const CAbiSignature &sig,
...
+    add_asm_line(aw, "movq %rcx, %r10");
+    add_asm_line(aw, "movq %rdx, %r11");
+    emit_mixed_wrpkru(aw, caller_pkey, target_pkey);
+    add_asm_line(aw, "movq %r10, %rcx");
+    add_asm_line(aw, "movq %r11, %rdx");
+
+    // Switch stacks to the target stack
+    emit_switch_stacks(aw, caller_pkey, target_pkey, "r11"s);
+  } else if (arch == Arch::Aarch64) {
+    // no-op for now
```
ayrtonm, line 348: nit: I added warnings while working on the ARM stuff for things that we want to punt for now. I think `#warning` doesn't make sense in this case, but you could log using `llvm::errs() << $MSG` to make sure we don't forget anything. Just a nit since this is a WIP anyway.

rewriter/GenCallAsm.cpp
```diff
@@ -280,21 +280,309 @@ static std::string sig_string(const CAbiSignature &sig,
...
+static AsmWriter get_asmwriter(bool as_macro) {
+  std::string terminator = {};
+  // Code generated as a macro needs to terminate each line with '\'
+  if (as_macro) {
+    terminator = "\\"s;
+  }
+  return {.ss = {}, .terminator = terminator};
+}
+
+static void switch_stacks(AsmWriter &aw, uint32_t caller_pkey, uint32_t target_pkey, Arch arch) {
```
ayrtonm, line 321: nit: Most of these function names could be a bit more descriptive/accurate by prefixing with either `add_` or `emit_` to point out that they're generating code to do X not doing X themselves (e.g. `emit_stack_switch`).

rewriter/GenCallAsm.cpp
```diff
@@ -280,21 +280,309 @@ static std::string sig_string(const CAbiSignature &sig,
   return ss.str();
 }
 
+static void call_string(
```
ayrtonm, line 283: nit: Maybe `emit_fn_call` or `emit_call_target_fn`?

rewriter/GenCallAsm.cpp
```diff
@@ -280,21 +280,309 @@ static std::string sig_string(const CAbiSignature &sig,
...
+  if (as_macro) {
+    terminator = "\\"s;
+  }
+  return {.ss = {}, .terminator = terminator};
+}
+
+static void switch_stacks(AsmWriter &aw, uint32_t caller_pkey, uint32_t target_pkey, Arch arch) {
+  if (arch == Arch::X86) {
+    // Save the old frame pointer and set the frame pointer for the call gate
+    add_asm_line(aw, "pushq %rbp");
```
ayrtonm, line 324: This is part of the function prologue, not exactly related to stack switching so it might make sense to move it out of this function. Especially if that means you can reuse this function for the second time we generate the stack switching code.

ayrtonm:
Overall looks like what we had in mind, the only thing that stands out aside from the few nits is how you split up the stuff after the function call. I was thinking something more similar to what you have before the function call (i.e. `switch_stacks` -> `copy_args` -> `zero_regs` -> `set_pkru` just in a different order).

Right now ` cleanup_and_restore_after_call` and `finalize_and_return_to_caller` aren't very descriptive names and combine a few of the things in the functions mentioned above.

rewriter/GenCallAsm.cpp
```diff
@@ -280,21 +280,309 @@ static std::string sig_string(const CAbiSignature &sig,
...
+    add_asm_line(aw, "movq %r11, %rdx");
+
+    // Load registers that are preserved across function calls after switching
+    // back to the caller's compartment's stack. This is on the caller's stack so
+    // it's not in the diagram above.
+    for (auto r = preserved_registers.rbegin(); r != preserved_registers.rend();
+         r++) {
+      add_asm_line(aw, "popq %"s + *r);
+    }
+    // Restore the caller's frame pointer
```
ayrtonm, line 567: The comment about the function prologue also applies to the epilogue here.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -387,241 +750,41 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
...
-    add_comment_line(
-        aw, "Push return regs to caller's stack before scrubbing registers");
-    // Push return regs to the caller's stack. These pushes have matching pops
-    // before the return so it has no net effect on the caller's stack pointer.
-    for (auto loc : return_locs) {
-      if (!loc.is_stack()) {
-        emit_reg_push(aw, loc);
-      }
-    }
+  emit_push_return_regs_to_caller_stack(aw, target_pkey, return_locs, arch);
```
ayrtonm, line 775: This and the next two functions (`emit_return_zero_regs` and `emit_pop_return_regs`) should probably be combined into one function. I was thinking something analogous to `emit_zero_regs` so it'll be easier to deduplicate some code in a later PR.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -280,21 +286,378 @@ static std::string sig_string(const CAbiSignature &sig,
...
+  } else if (arch == Arch::Aarch64) {
+    // TODO ARM stack switching
+    llvm::errs() << "TODO register saving not implemented on ARM\n";
+  }
+}
+
+static void emit_intermediate_pkru(AsmWriter &aw, uint32_t caller_pkey, uint32_t target_pkey, Arch arch) {
+  if (arch == Arch::X86) {
+    uint32_t caller_pkru = ~((0b11 << (2 * caller_pkey)) | 0b11);
+    add_raw_line(aw, llvm::formatv("ASSERT_PKRU({0:x8}) \"\\n\"", caller_pkru));
```
ayrtonm, line 344: The lines inserting `ASSERT_PKRU` could be pulled out and left in `emit_asm_wrapper` so you can combine `emit_intermediate_pkru` and `emit_return_intermediate_pkru` since they'll be identical.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -280,21 +286,378 @@ static std::string sig_string(const CAbiSignature &sig,
...
+    }
+  } else if (arch == Arch::Aarch64) {
+    // TODO ARM reg zero
+    llvm::errs() << "TODO register zeroing not implemented on ARM\n";
+  }
+}
+
+static void emit_set_pkru(AsmWriter &aw, uint32_t target_pkey, WrapperKind kind, Arch arch) {
+  if (arch == Arch::X86) {
+    if (kind == WrapperKind::IndirectCallsite) {
```
ayrtonm, line 461: This branch isn't related to the PKRU so you should pull it out of this function.

rewriter/GenCallAsm.cpp
sim-immunant: I reversed the order of these nested ifs to deduplicate the asserts.

rewriter/GenCallAsm.cpp
sim-immunant: I threw in llvm errrs for all the ARM no-ops.

rewriter/GenCallAsm.cpp
sim-immunant: I changed all these to `emit_` and altered the names after reshuffling to make them more descriptive.

rewriter/GenCallAsm.cpp
sim-immunant: Addressed this in the reshuffle.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -280,6 +286,335 @@ static std::string sig_string(const CAbiSignature &sig,
...
+
+  add_comment_line(aw, "Call wrapped function");
+
+  if (!target_name) {
+    // indirect call
+    assert(kind == WrapperKind::IndirectCallsite);
+    if (arch == Arch::X86) {
+      add_asm_line(aw, "call *%r12");
+    } else if (arch == Arch::Aarch64) {
+      add_asm_line(aw, "br x12"); // TODO, which register?
```
ayrtonm, line 303: I'd say leave this with a `llvm::errs() << TODO` for now.

rewriter/GenCallAsm.cpp
sim-immunant: Moved these into their own function `emit_save_regs`.

rewriter/GenCallAsm.cpp
sim-immunant: Moved this into a corresponding `emit_restore_regs`

tools/rewriter/GenCallAsm.cpp
sim-immunant: Done.

tools/rewriter/GenCallAsm.cpp
sim-immunant: I pulled this out into the top level since we can conditionally define that macro depending on arch.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -280,6 +286,335 @@ static std::string sig_string(const CAbiSignature &sig,
...
+    for (auto &r : preserved_registers) {
+      add_asm_line(aw, "pushq %"s + r);
+    }
+  } else if (arch == Arch::Aarch64) {
+    // TODO ARM stack switching
+    llvm::errs() << "TODO register saving not implemented on ARM\n";
+  }
+}
+
+static void emit_intermediate_pkru(AsmWriter &aw, uint32_t caller_pkey, uint32_t target_pkey, bool returning, Arch arch) {
```
ayrtonm, line 341: Having `if (returning)` makes it hard to see at a glance what assembly is generated in each code path. Can we switch the `bool returning` for an argument with a list of the two registers that need to be saved across the wrpku? So the calls to this function would be
```
emit_intermediate_pkru(..., "rcx", "rdx", ...)
...
emit_intermediate_pkru(..., "rax", "rdx", ...)
```
instead of
```
emit_intermediate_pkru(..., false /* returning */, ...)
...
emit_intermediate_pkru(..., true /* returning */, ...)
```

tools/rewriter/GenCallAsm.cpp
```diff
@@ -280,6 +286,335 @@ static std::string sig_string(const CAbiSignature &sig,
...
+    add_asm_line(aw, "movq %rcx, %r10");
+    add_asm_line(aw, "movq %rdx, %r11");
+    emit_wrpkru(aw, target_pkey);
+    add_asm_line(aw, "movq %r10, %rcx");
+    add_asm_line(aw, "movq %r11, %rdx");
+  } else if (arch == Arch::Aarch64) {
+    // set X18 to the pointer key (compartment number left-shifted 56 bits)
+    add_asm_line(aw, "mov x18, #" + std::to_string(target_pkey & 0xF));
+    add_asm_line(aw, "lsl x18, x18, #56");
+  } else if (arch == Arch::Aarch64) {
```
ayrtonm, line 485: remove dead code?

tools/rewriter/GenCallAsm.cpp
```diff
@@ -280,6 +286,335 @@ static std::string sig_string(const CAbiSignature &sig,
...
+static AsmWriter get_asmwriter(bool as_macro) {
+  std::string terminator = {};
+  // Code generated as a macro needs to terminate each line with '\'
+  if (as_macro) {
+    terminator = "\\"s;
+  }
+  return {.ss = {}, .terminator = terminator};
+}
+
+static void emit_save_regs(AsmWriter &aw, uint32_t caller_pkey, uint32_t target_pkey, Arch arch) {
```
ayrtonm, line 324: I think `emit_prologue` and `emit_epilogue` make more sense than `emit_save_regs` and `emit_restore_regs` since there are other places where we save and restore regs (e.g. across calls to libia2_scrub_registers).

sim-immunant: Done.

tools/rewriter/GenCallAsm.cpp
sim-immunant: Whoops! Fixed.

tools/rewriter/GenCallAsm.cpp
sim-immunant: Good idea, I've done this.

tools/rewriter/GenCallAsm.cpp
sim-immunant: I moved this out into its own function `emit_fn_ptr`

ayrtonm:
One of the `emit_scrub_regs` calls has the wrong arguments like we saw while talking through this.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -387,243 +678,44 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
...
-  add_asm_line(aw, "movq %r11, %rdx");
-
-  // Call wrapped function
-  add_comment_line(aw, "Call wrapped function");
-  if (!target_name) {
-    assert(kind == WrapperKind::IndirectCallsite);
-    add_asm_line(aw, "call *%r12");
-  } else {
-    add_asm_line(aw, "call "s + target_name.value());
+    emit_fn_ptr(aw, arch);
```
ayrtonm, line 696: nit: `emit_load_fn_ptr`

sim-immunant: Fixed

ayrtonm:
Also I think the first emit_scrub_regs call produces different code when reg_arg_count is zero.

sim-immunant:
> One of the `emit_scrub_regs` calls has the wrong arguments like we saw while talking through this.

Fixed.

> Also I think the first emit_scrub_regs call produces different code than before when reg_arg_count is zero.

You're right, the call to `__libia2_scrub_registers` happens regardless in the old code. This is fixed now.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -280,6 +286,300 @@ static std::string sig_string(const CAbiSignature &sig,
...
+    // and r11 as scratch registers
+    add_asm_line(aw, "movq %rcx, %r10");
+    add_asm_line(aw, "movq %rdx, %r11");
+    emit_wrpkru(aw, target_pkey);
+    add_asm_line(aw, "movq %r10, %rcx");
+    add_asm_line(aw, "movq %r11, %rdx");
+  } else if (arch == Arch::Aarch64) {
+    // set X18 to the pointer key (compartment number left-shifted 56 bits)
+    add_asm_line(aw, "mov x18, #" + std::to_string(target_pkey & 0xF));
+    add_asm_line(aw, "lsl x18, x18, #56");
```
ayrtonm, line 446: you can set x18 with one instruction with something like `movz x18, #0x0100, LSL 48". This sets x18 to `0x0100 << 48` so we just need to modify the upper byte in the 16-bit immediate field.

ayrtonm:
@sim-immunant x18 switching is blocked on building libc and ld.so with ffixed-x18 so I'll just remove it and merge this myself.

PR #325, opened 2024-03-13
Misc fixes for recent Arch linux

rinon:


fw-immunant (approved):
LGTM, I had made a similar fix locally for the rewriter.

PR #330, opened 2024-03-21
runtime/tracer: Ignore procfs race in syscall tracer

ayrtonm:
This is the fix I added in #311 for a procfs issue. On second thought the cleaner solution would be to propagate actual errors from allow_path to its callers and check those like we did in #318 with ENOENT.

### DIFF
--- a/runtime/tracer/forbid_paths.c
+++ b/runtime/tracer/forbid_paths.c
@@ -135,7 +135,8 @@ static int forbid_deep_or_shallow(const char *fpath, const struct stat *sb,
 
   /* allow whole dir or file */
   int ret = allow_path(fpath, ctx.ruleset_fd, false);
-  if (ret < 0) {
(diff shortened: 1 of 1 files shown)


fw-immunant:
What is the actual issue/race we hit here?

ayrtonm:
It's like the issue fixed [here](https://github.com/immunant/IA2-Phase2/pull/318/commits/49e2c3ddf8820dea774cdd424c18e4f8409bb07a) where a tracee process ends while we still have a reference to `/proc/$PID/` and we hit an error from trying to use it. This [run](https://github.com/immunant/IA2-Phase2/actions/runs/8350880058/job/22858199580) shows that `fpath` is `/proc/$PID` but I'm not sure exactly where in `allow_path` it fails since we just return -1.

fw-immunant (commented):
A bit ugly but LGTM, will follow up with an additional log message for the failing `fstat()` in `allow_path`.

fw-immunant (approved):

PR #333, opened 2024-03-26
Port callgate generation code to ARM

ayrtonm:
I also added some basic prologue/epilogue stuff in the rewriter to get calls to scrub_registers working.


ayrtonm:
TODO: missing ret on scrub registers

ayrtonm:
Also x16 and x17 probably shouldn't be zeroed

ayrtonm:
This now has stack switching so the main thing missing from the callgates is copying arguments and return values between stacks

rinon:
This is ready for review, it should handle all of the ARM ABI now (not including scalable vectors).

tools/rewriter/GenCallAsm.cpp
```diff
@@ -11,125 +12,168 @@
...
+    if (arch == Arch::X86) {
+      // if the return is in memory, the first integer argument is the location to
+      // write the return value
+      auto ptr_arg = ArgLocation::Register(ArgLocation::Kind::Integral, 8);
+      ptr_arg.allocate_reg(int_param_reg_order[ints_used]);
+      args.push_back(ptr_arg);
+      ints_used += 1;
+    } else if (arch == Arch::Aarch64) {
+      // memory return goes in x8
+      auto ptr_arg = ArgLocation::Register(ArgLocation::Kind::Integral, 8);
```
ayrtonm, line 65: I'm interpreting this as the return value goes in an 8 byte slot on the stack which x8 points to. Why is this 8?

ayrtonm:
I had some minor doubts when I wrote the register scrubbing part about whether zeroing x16 and x17 was ok since they're used in PLT stubs and veneers. I think it should be fine since the call gates run after the `blr x1[6,7]` and user code shouldn't be relying on the values of these registers in the return path. I just want to make sure that linker-generated code doesn't rely on them in the return path. @rinon do you think zeroing them is fine?

tools/rewriter/GenCallAsm.cpp
rinon: This is the register (x8) holding the address of the stack return slot. I'll clarify.

rinon:
The way I read the ABI, x17 and x18 are available for us by the linker between the caller and callee, e.g. in PLT code. We are the "callee" from the linker's perspective so at the point the wrapper starts executing these regs are available temporary registers. They are not preserved across calls at all.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -481,21 +668,29 @@ static void emit_copy_stack_returns(AsmWriter &aw, size_t stack_return_size, siz
...
+    if (stack_return_size > 0) {
+      add_comment_line(aw, "Copy stack returns to caller's stack");
+
+      size_t stack_return_saved_ptr_offset = stack_arg_size + stack_arg_padding;
+      size_t stack_return_start = stack_return_saved_ptr_offset + 8;
+
+      // Pop the original return value address into x8
+      add_asm_line(aw, "ldr x8, [sp, #"s + std::to_string(stack_return_saved_ptr_offset) + "]");
+      // Copy the value back to the caller's stack
+      emit_memcpy(aw, stack_return_size, "x8", "sp", "x9", stack_return_start, arch);
```
ayrtonm, line 686: Since there's no intermediate PKRU we need to use tagged pointers for memory accesses right? I think all the other calls to emit_memcpy use pointers derived from sp, but I'm not totally sure about `x8` here since it's passed in as a hidden argument.

rinon: I think we need to, yeah. I don't know how exactly we're doing this, maybe that should be a separate PR?

ayrtonm: Sure that can be a separate PR

ayrtonm: I imagine there might be more cases where we need to use tagged pointers in the call gates but we can handle those as a second pass in a separate PR. I'll just focus on whether it works w/o MTE for this PR.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -637,6 +898,45 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
     |addr |stack by the call to the wrapped function.
     +-----+
 
+    For AArch64:
+    +-----+
+    | top |Top of the stack (stack grows down on AArch64). This address is
+    |     |aligned to 16 bytes.
+    +-----+
+    |ind  |Padding for alignment prior to the indirect args
+    |args |as such memory must be 16B aligned. (TODO: verify this)
```
ayrtonm, line 907: The padding at the start of the stack frame seems unnecessary since this is a public interface, right?
```
The stack must also conform to the following constraint at a public interface:
• SP mod 16 = 0. The stack must be quad-word aligned.
```

tools/rewriter/GenCallAsm.cpp
```diff
@@ -637,6 +898,45 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
...
+    |     |placed any lower on the stack. (TODO: verify this)
+    +-----+
+    |stack|Space required for stack arguments. This is initialized from the
+    |args |analogous part of the caller's stack. If all arguments are passed in
+    |     |registers this space isn't allocated.
+    +-----+
+    |frame|Previous link register contents. TODO: update this when we transition stacks?
+    |ptr  |
+    +-----+
+    |ret  |Return address for wrapped function. This is implicitly placed on the
```
ayrtonm, line 937: nit: `bl` on arm64 only updates the link register. it doesn't implicitly put it on the stack right?

tools/rewriter/GenCallAsm.cpp
```diff
@@ -652,19 +952,22 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
...
   size_t compartment_stack_space = start_of_ret_space + stack_return_size +
-                                   stack_return_padding + stack_arg_size;
-  // Compute what the stack alignment would be before calling the wrapped
-  // function to check if we need to insert 8 bytes for alignment. We add 8
-  // bytes to the compartment_stack_space since the frame is initially off by 8
-  // bytes.
-  size_t stack_alignment = (compartment_stack_space + 8) % 16;
+                                   stack_return_padding + stack_arg_size + stack_arg_padding;
+  size_t stack_alignment = 0;
+  if (arch == Arch::X86) {
```
ayrtonm, line 964: I think we have the alignment block that this corresponds to on both the arm and x86 diagrams so I don't see why this check is x86-specific.

tools/rewriter/GenCallAsm.cpp
ayrtonm: Also this and the frame ptr blocks are in the wrong place since `emit_prologue` is the first thing called.

tools/rewriter/GenCallAsm.cpp
rinon: It's for aligning the indirect args space becuase the stack grows down. The start of each indirect arg region must be aligned to the correct alignment for that arg though, so maybe we don't need a separate slot here to do this.

tools/rewriter/GenCallAsm.cpp
rinon: Ah sorry I meant frame pointer and ret addr pushed on by the prologue of the callee. The frame pointer and return address of the wrapper is on the caller's stack. I'll clarify that this is part of the next frame.

tools/rewriter/GenCallAsm.cpp
rinon: We don't actually insert it on ARM. I'll try to figure out some of the TODOs on the diagram.

rinon:
I want to redo how we compute all the stack offsets for this, but I have another PR pending on top and I'd rather do that after both land.

tools/rewriter/GenCallAsm.cpp
ayrtonm: oh I right I forgot there are potentially multiple indirect args. I think if it makes the implementation easier and bugs less likely then a padding slot per arg is preferable for now.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -11,125 +12,168 @@
...
+      args.push_back(arg);
       floats_used += 1;
       break;
-    case CAbiArgKind::Memory:
-      // memory return also returns address in first return register
-      if (locs.empty()) {
-        locs.push_back(ParamLocation::Register(int_ret_reg_order[0]));
+    case ArgLocation::Kind::Memory:
+      // memory return also returns address in first return register on x86
+      if (arch == Arch::X86) {
```
ayrtonm, line 167: Maybe assert `arch == Arch::X86` here?

ayrtonm:
I haven't looked at the generated code, but the changes look correct to me.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -681,13 +979,15 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
 
   add_raw_line(aw, llvm::formatv("ASSERT_PKRU({0:x8}) \"\\n\"", ~((0b11 << (2 * caller_pkey)) | 0b11)));
```
ayrtonm, line 980: nit: this is a no-op on ARM but we still emit it with a PKRU value

tools/rewriter/GenCallAsm.cpp
rinon: We can have memory returns on ARM, so I don't think this can be an assert. We just don't use one of the regular return registers to point to the address, there's a dedicated register.

ayrtonm: If there's a memory return (`func.ret` has one item w/ `ArgLocation::Kind::Memory`) then `allocate_return_locations` returns an empty vector and `stack_return_size` in `emit_asm_wrapper` stays 0 right? This seems like an issue unless I'm missing something.

rinon: Good point. All memory returns were classified as indirect on AArch64, but that resulted in us preserving an extra return register as well, which wasn't necessary. I've clarified and fixed this up now.

Issue #314, opened 2024-01-05
Add support for ARM64 using MTE


ayrtonm:
To add support for ARM64 we can use MTE and the x18 register in place of the PKRU on x64. We need to port code from x64 to ARM in a few places and I think we'll need a WIP for all of them before we can run basic compartmentalized programs. The main things that I can think of that we'll need are:

## Port libia2 code to ARM
The items handled in #311 should be enough to get a basic program running depending on what the call gates do. The other three points will be needed for things like threading and stack switching but we can ignore them for now (may require modifying build flags for ARM to avoid ld errors about missing symbols).
- [ ] `#ifdef` out x64 code to allow building libia2.a with aarch64 compiler (#311)
- [ ] port code for protecting loaded, writable ELF segments (#311)
- [ ] port wrapper for `exit`
- [ ] port wrapper for `main`
- [ ] port `ia2_thread_begin`

## Modify rewriter to generate call gates for ARM
Here we need to modify three things:
1. call gate generation (GenCallAsm.cpp)
2. mapping from args/ret to ABI (DetermineAbi.cpp)
3. review usage of pkey 0 since MTE tag 0 does not differ from the others (mainly SourceRewriter.cpp)

For the call gate generation we can port it incrementally in the following way and validating manually with gdb along the way
- [ ] a WIP that generates "no-op" call gates with a single `call` instruction (only expected to work for register-only ABIs)
- [ ] add x18 switching
- [ ] add in scrub registers (includes saving/restoring callee-saved registers, only expected to work for no args/ret)
- [ ] add in stack switching

Stack switching depends on changes to DetermineAbi.cpp, but everything before that can be done in parallel with the ABI stuff.

## Modify LLVM to instrument memory accesses
We also need to modify LLVM to combine the tag in x18 with pointers before they're dereferenced. LLVM has two pass managers ([legacy](https://llvm.org/docs/WritingAnLLVMPass.html), [new](https://llvm.org/docs/NewPassManager.html)) and I'm not sure how well supported [modifying the codegen pipeline](https://llvm.org/docs/NewPassManager.html#status-of-the-new-and-legacy-pass-managers) with the new pass manager is, so it may be best to just make in-tree changes to LLVM.

[...]

PR #334, opened 2024-03-26
runtime/libia2: enable MTE

ayrtonm:
This PR actually enables MTE after #311 added support for cross-compiling for ARM. I wrote some basic tests that protects static data but we'll need to test a different way as explained in [this comment](https://github.com/immunant/IA2-Phase2/issues/314#issuecomment-2019145083).

### DIFF
--- a/runtime/libia2/ia2.c
+++ b/runtime/libia2/ia2.c
@@ -127,6 +127,7 @@ int ia2_mprotect_with_tag(void *addr, size_t len, int prot, int tag) {
     int res = mprotect(addr, len, prot | PROT_MTE);
     if (res != 0) {
         /* Skip memory tagging if mprotect returned an error */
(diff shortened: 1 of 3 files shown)


ayrtonm:
Rebased this on #353 which adds support for running tests in QEMU. Note that the qemu from ubuntu repository is too old to recognize the `-one-insn-per-tb` flag and uses the `-singlestep` alias instead while our forked qemu is too new to accept the deprecated `-singlestep` flag. So the latest CI failure on this PR will not actually be an issue.

fw-immunant:
This appears to work correctly: we see MTE violations in test logs [here](https://github.com/immunant/IA2-Phase2/actions/runs/9375462801/job/25813623593). We can wait to merge this until we fix most (if not all) of these with call-gates.

PR #336, opened 2024-03-27
pad-tls: Add support for ARM64 to pad-tls and fix libc lookup in CMake

ayrtonm:
Previously CMake was picking up libc.so for x86 and running pad-tls on that. We were seemingly linking against it but the basic ARM test wasn't failing since QEMU doesn't actually load the guest libc.so into the address space only ld.so. This fixes both libc lookup paths and modifies pad-tls to allow EM_AARCH64. Since this has no effect on the ARM test when emulating with QEMU I checked the libc copy in the build dir manually.


ayrtonm:
correction: dynamically-linked QEMU wasn't loading libc.so immediately only ld.so and the main binary. ld.so would then try to load libc before main and hang or crash because of the arch mismatch. Statically linked QEMU loads them all at once.

PR #341, opened 2024-04-02
Initial determineAbi work for ARM

sim-immunant:
Add some comments in determineABI as I work through supporting ARM. There are several question points I've marked with TODO.

Regarding `determineAbi`, On aarch64, section 6.9 of the ABI writes:

>If the type, T, of the result of a function is such that `void func(T arg)` would require that arg be passed as a value in a register (or set of registers) according to the rules in Parameter passing, then the result is returned in the same registers as would be used for such an argument.

This is a little confusing but basically means return types are returned according to the same rules as for parameters, if I'm understanding this correctly. Composite (ie struct) arguments larger than 16 bytes (two registers) must go on the stack, so we still have the same two-register limitation as on x86.

Some non-composite types might be larger than 16 bytes, namely floating point or vector types, which are passed and returned in the `v0`-`v7` registers. We have a similar thing on x86 with the `xmmX` and `st0` registers.

All told, I think that we may not have to do anything particular to this file to support ARM. We might not even have to add architecture arguments as I've done here. The ARM-specific stuff may all belong in GenCallAsm.

The next step would be if someone could confirm my understanding of the above and take a look at the items I've marked TODO in this PR.

### DIFF
--- a/tools/rewriter/DetermineAbi.cpp
+++ b/tools/rewriter/DetermineAbi.cpp
@@ -1,4 +1,5 @@
 #include "CAbi.h"
+#include "GenCallAsm.h"
 #include "clang/AST/AST.h"
 #include "clang/AST/RecordLayout.h"
 #include "clang/Basic/CodeGenOptions.h"
@@ -135,7 +136,8 @@ static CAbiArgKind classifyLlvmType(const llvm::Type &type) {
 static std::vector<CAbiArgKind>
 abiSlotsForArg(const clang::QualType &qt,
                const clang::CodeGen::ABIArgInfo &argInfo,
-               const clang::ASTContext &astContext) {
+               const clang::ASTContext &astContext,
+               Arch arch) {
   // this function is most similar to Clang's `ClangToLLVMArgMapping::construct`
   typedef enum clang::CodeGen::ABIArgInfo::Kind Kind;
   switch (argInfo.getKind()) {
@@ -221,7 +223,7 @@ cgFunctionInfo(clang::CodeGen::CodeGenModule &cgm,
 }
 
 CAbiSignature determineAbi(const clang::CodeGen::CGFunctionInfo &info,
-                           const clang::ASTContext &astContext) {
+                           const clang::ASTContext &astContext, Arch arch) {
   // get ABI for return type and each parameter
   CAbiSignature sig;
   sig.variadic = info.isVariadic();
@@ -235,7 +237,7 @@ CAbiSignature determineAbi(const clang::CodeGen::CGFunctionInfo &info,
 
   // Get the slots for the return value.
   auto &returnInfo = info.getReturnInfo();
-  sig.ret = abiSlotsForArg(info.getReturnType(), returnInfo, astContext);
+  sig.ret = abiSlotsForArg(info.getReturnType(), returnInfo, astContext, arch);
 
   auto is_integral = [](auto &x) { return x == CAbiArgKind::Integral; };
 
@@ -257,14 +259,14 @@ CAbiSignature determineAbi(const clang::CodeGen::CGFunctionInfo &info,
   // Now determine the slots for the arguments
   for (auto &argInfo : info.arguments()) {
     clang::QualType paramType = argInfo.type;
-    auto slots = abiSlotsForArg(paramType, argInfo.info, astContext);
+    auto slots = abiSlotsForArg(paramType, argInfo.info, astContext, arch);
     sig.args.insert(sig.args.end(), slots.begin(), slots.end());
   }
 
   return sig;
 }
 
-CAbiSignature determineAbiForDecl(const clang::FunctionDecl &fnDecl) {
+CAbiSignature determineAbiForDecl(const clang::FunctionDecl &fnDecl, Arch arch) {
   clang::ASTContext &astContext = fnDecl.getASTContext();
 
   // set up context for codegen so we can ask about function ABI
@@ -306,11 +308,11 @@ CAbiSignature determineAbiForDecl(const clang::FunctionDecl &fnDecl) {
            info.getASTCallingConvention());
     abort();
   }
-  return determineAbi(info, astContext);
+  return determineAbi(info, astContext, arch);
 }
 
 CAbiSignature determineAbiForProtoType(const clang::FunctionProtoType &fpt,
-                                       clang::ASTContext &astContext) {
+                                       clang::ASTContext &astContext, Arch arch) {
   // FIXME: This is copied verbatim from determineAbiForDecl and could be
   // factored out. This depends on what we do with PR #78 so I'm leaving it as
   // is for now.
@@ -339,5 +341,5 @@ CAbiSignature determineAbiForProtoType(const clang::FunctionProtoType &fpt,
       cgm,
       fpt.getCanonicalTypeUnqualified().castAs<clang::FunctionProtoType>());
 
-  return determineAbi(info, astContext);
+  return determineAbi(info, astContext, arch);
 }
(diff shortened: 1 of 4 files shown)


ayrtonm:
> Composite (ie struct) arguments larger than 16 bytes (two registers) must go on the stack, so we still have the same two-register limitation as on x86.

Does the two register restriction apply to bit fields as well? How would this be returned/passed in?
```
struct s7 {
   bool b1 : 1;
   char c1 : 7;
   char c2;
}
```

tools/rewriter/SourceRewriter.cpp
```diff
@@ -475,7 +475,8 @@ class FnPtrCall : public RefactoringCallback {
     auto *fpt = expr_ty->castAs<clang::PointerType>()
                     ->getPointeeType()
                     ->getAsAdjusted<clang::FunctionProtoType>();
-    fn_ptr_abi_sig[expr_ty_str] = determineAbiForProtoType(*fpt, ctxt);
+    // TODO figure out the correct arch
```
ayrtonm, line 478: should just be `Target` defined on line 51 right?

tools/rewriter/DetermineAbi.cpp
```diff
@@ -66,13 +73,16 @@ static std::vector<CAbiArgKind> classifyDirectType(const clang::Type &type,
           pending_kind.reset();
         }
 
-        // Update pending_kind based on ABI rules
+        // Update pending_kind based on ABI rules.
+        // We're passing in a scalar so we expect we will only get one element.
```
fw-immunant, line 78: It isn't actually always the case that we pass a scalar type here. A struct like `struct s16 { struct { int a; int b; } inner; };` will also take this path for its first field but the recursive call does return a vector of a single eightbyte classification here because the inner struct fits in a single register.

tools/rewriter/DetermineAbi.cpp
```diff
@@ -157,18 +168,22 @@ abiSlotsForArg(const clang::QualType &qt,
...
         auto elems = STy->elements();
         std::vector<CAbiArgKind> out = {};
         for (const auto &elem : elems) {
           out.push_back(classifyLlvmType(*elem));
         }
         return out;
       } else {
+        // TODO why a single int slot when not flattenable?
         return {CAbiArgKind::Integral}; // one pointer in register, see `clang's
                                         // ClangToLLVMArgMapping::construct`
```
fw-immunant, line 184: As indicated by the comment below, non-flattenable types passed as `Direct` are a single pointer in a register. See the branch on `AI.isDirect() && AI.getCanBeFlattened() && STy` [here](https://clang.llvm.org/doxygen/CGCall_8cpp_source.html).

tools/rewriter/DetermineAbi.cpp
```diff
@@ -157,18 +168,22 @@ abiSlotsForArg(const clang::QualType &qt,
...
       }
     }
     // if not flattenable, classify the single-register value
+    // TODO this seems to be the no-StructType case, not the not-flattenable case
     return classifyDirectType(*qt.getCanonicalType(), astContext);
```
fw-immunant, line 189: I think you're right here; the comment should be corrected to say that we have a scalar type here.

tools/rewriter/DetermineAbi.cpp
```diff
@@ -204,32 +219,53 @@ cgFunctionInfo(clang::CodeGen::CodeGenModule &cgm,
...
+  // Each slot is of one of the types described in CAbiArgKind.
+
+  // Get the slots for the return value.
   auto &returnInfo = info.getReturnInfo();
   sig.ret = abiSlotsForArg(info.getReturnType(), returnInfo, astContext);
 
   auto is_integral = [](auto &x) { return x == CAbiArgKind::Integral; };
+
+  // num_regs is the number of registers in which the return value is stored.
+  // This may be one for a value up to 64 bits or two for a 128-bit value.
```
fw-immunant, line 245: This is specifically only counting the number of integral registers--floating-point registers (`xmmN`) will not be counted here, and I'm not sure if that's a bug. I believe @ayrtonm originally wrote this code (see #86), and it never appears to have counted float-kind eightbytes.

tools/rewriter/DetermineAbi.cpp
```diff
@@ -204,32 +219,53 @@ cgFunctionInfo(clang::CodeGen::CodeGenModule &cgm,
...
   size_t num_regs = std::count_if(sig.ret.begin(), sig.ret.end(), is_integral);
-  if (num_regs > 2) {
+  // It's possible that clang gives us back a value of three or more integers,
+  // for example a struct full of ints. However, in reality the whole value
+  // goes on the stack in this case according to the x64 ABI.
+  // We handle that case here.
```
fw-immunant, line 250: I wish Clang documented which transformations it does and doesn't do by this point in compilation, but sadly we're really flying by the seat of our pants. I would maybe ask to qualify more what kind of "possible" this is--one where we don't know what Clang does right now (possibly for some types we haven't observed), or one where we don't know what Clang may do in the future.

tools/rewriter/DetermineAbi.cpp
```diff
@@ -42,6 +42,13 @@ static std::vector<CAbiArgKind> classifyDirectType(const clang::Type &type,
           "unsupported scalar type (obj-C object, Clang block, or C++ member) found during ABI computation");
     }
   } else {
+    // Handle the case where we pass a struct directly in registers.
+    // The strategy here is to iterate through each field of the struct
+    // and record the ABI type each time we exit an eightbyte chunk.
+
+    // TODO I think this branch is dead code. We only call this in abiSlotsForArg
+    // with single-register values.
+
```
fw-immunant, line 52: This is definitely not dead code; it runs for the following types in our "structs" test: `s1`, `s2`, `s4`, `s7`. I believe this ends up being called whenever a struct type fits in a single register, but I don't know any particular guarantee Clang gives us around this, so this function does have some subsequent logic that may be dead.

fw-immunant (changes requested):
Various comments inline. Some of the added comments are excellent stepping stones for future readers, others may not be worth spacing out the adjacent logic--just keep the more informative ones imo.

sim-immunant:
> Does the two register restriction apply to bit fields as well? How would this be returned/passed in?

From 10.1.8:

> For each bit-field, the type of its container is:
> * Its declared type if its size is no larger than the size of its declared type.
> * The largest integral type no larger than its size if its size is larger than the size of its declared type.

I think this means your example struct occupies three bytes and would get passed in a single register.

tools/rewriter/SourceRewriter.cpp
sim-immunant: Whoops, fixed.

tools/rewriter/DetermineAbi.cpp
```diff
@@ -204,32 +219,53 @@ cgFunctionInfo(clang::CodeGen::CodeGenModule &cgm,
...
     clang::QualType paramType = argInfo.type;
     auto slots = abiSlotsForArg(paramType, argInfo.info, astContext);
     sig.args.insert(sig.args.end(), slots.begin(), slots.end());
   }
+
   return sig;
 }
 
 CAbiSignature determineAbiForDecl(const clang::FunctionDecl &fnDecl) {
+// Given a C function declaration, find the layout of the argument and return value slots.
```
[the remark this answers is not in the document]
sim-immunant: I dropped these comments as you suggested below. If you know what a CAbiSignature is, it's self-explanatory.

tools/rewriter/DetermineAbi.cpp
sim-immunant: Fixed!

tools/rewriter/DetermineAbi.cpp
sim-immunant: I see it now. I was getting "scalar" and "direct" confused I think.

sim-immunant:
At our impromptu sync this morning we discussed merging this and getting my classifyDirectType and abiSlotsForArg ARM changes in another PR. If that still sounds good drop and approval and I'll merge.

fw-immunant (approved):

PR #343, opened 2024-04-04
misc/spoofed_criterion: Fork for tests on ARM

### LINKED ISSUE #342: Make spoofed criterion fork for tests
In #337 we added a lib matching criterion's API to avoid having to rebuild it and it's transitive deps with -ffixed-x18 on ARM. This works for most tests, but a few of our tests call `exit` (e.g. tests/destructors) so we need to have our lib fork for each test instead of just calling them. We still don't support [running tests in QEMU from cmake](https://github.com/immunant/IA2-Phase2/issues/314#issue-2068101763) so this is probably lower priority than that.

ayrtonm:
Closes #342

### DIFF
--- a/misc/spoofed_criterion/include/criterion/criterion.h
+++ b/misc/spoofed_criterion/include/criterion/criterion.h
@@ -2,6 +2,6 @@
 #include <criterion/new/assert.h>
 
 #define Test(suite, name) \
-    void suite##_##name(void); \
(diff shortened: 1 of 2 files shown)


fw-immunant (changes requested):
This needs to wait for the child and propagate failure if it exits with failed status.

fw-immunant:
The child terminating without yet being waited for results in a zombie process that sits around until the parent reaps it with wait/waitpid.

fw-immunant (approved):
Now forks for tests. Another reviewer welcome.

PR #344, opened 2024-04-05
Add glibc submodule and prebuilts with -ffixed-x18

ayrtonm:
This PR adds glibc from sourceware.org as a git submodule instead of subtree since I don't anticipate needing to modify it. I also added prebuilts for libc.so and ld.so built with `-ffixed-x18` to avoid needing to build them on each run. These prebuilts mostly don't clobber x18*, but they don't have memory accesses instrumented so they unblock x18 switching but not enabling MTE. I kept the configure command I used and added it to CMakeLists.txt in runtime/libia2 (gated under `LIBIA2_REBUILD_GLIBC`) for when we have the LLVM fork.

*comparing to the aarch64 libc.so from AUR with objdump, I see a lot fewer uses of x18 but there are still a few. In getcontext, swapcontext, etc. that's expected since it's probably written in asm. However I also see it in some soft-fp functions (mutlfv3, divft3) but I'm not sure where this code is coming from since the .c's with those functions don't show up in the build logs.


ayrtonm:
Test failed since I forgot to point QEMU to the build dir sysroot

ayrtonm:
current status: headers are copied over to the right places, but it's failing because it's using two sysroots on ARM in the rewriter step. The issue is that the CMake target has `--sysroot=`and the rewrite target is based off of this target but needs to use the clang headers instead. We should fix this before landing this.

ayrtonm:
Another option is to set the right sysroot in the rewriter itself (either aarch64-only or for all archs). Either way it should be the clang sysroot and since the rewriter is built with the same sysroot it should know the clang version/path to sysroot. This might be the more sane option since it means that users wouldn't need to modify the build system to explicitly add `-isystem` or `--sysroot`. Instead the rewriter would just know where to find the standard headers just like compilers normally do.

ayrtonm:
CI failing because I need to point qemu to the build dir sysroot.

ayrtonm:
CI is passing but a quick sanity check showed that aarch64 isn't actually using the prebuilt headers, it's still using the system aarch64 headers.

ayrtonm:
> Another option is to set the right sysroot in the rewriter itself

As much as I would've liked to do this, turns out it's a bit harder than expected since clang needs some of gcc's include-fixed headers anyway and I'd rather not reimplement that part of clang's Driver.cpp. To see how I fixed the multiple sysroot issue see the last two commits. Aarch64 doesn't actually use `-nostdinc` since the glibc headers rely on some system header that I seem to be missing (I only have the `-print-file-name=include` and `-print-file-name=include-fixed` headers rn). I did do a small sanity check to make sure that the glibc headers that are copied to the build dir are the ones that are being used which is what I ultimately cared about.

PR #346, opened 2024-04-08
WIP ABI classification for ARM

sim-immunant:


tools/rewriter/DetermineAbi.cpp
```diff
@@ -43,80 +140,89 @@ static std::vector<CAbiArgKind> classifyDirectType(const clang::Type &type,
           "unsupported scalar type (obj-C object, Clang block, or C++ member) found during ABI computation");
     }
   } else {
-    // Handle the case where we pass a struct directly in a register.
-    // The strategy here is to iterate through each field of the struct
-    // and record the ABI type each time we exit an eightbyte chunk.
+    if (arch == Arch::Aarch64) {
+      auto result = classifyARMAggregate(type, astContext);
+      for (auto i: result)
+        std::cout << (int)i << ' ';
```
ayrtonm, line 146: this might not work. use "llvm::outs() <<" or llvm::errs instead

tools/rewriter/DetermineAbi.cpp
```diff
@@ -43,80 +133,97 @@ static std::vector<CAbiArgKind> classifyDirectType(const clang::Type &type,
...
     }
   } else {
-    // Handle the case where we pass a struct directly in a register.
-    // The strategy here is to iterate through each field of the struct
-    // and record the ABI type each time we exit an eightbyte chunk.
+    if (arch == Arch::Aarch64) {
+      // TODO we have a bug here where `type` is not an aggregate
+      // (ie, type.getAsStructureType() returns nullptr). This shouldn't happen
+      // because if we're in this branch, we've already determined that we're
+      // not a scalar type (I think?). End result is a segfault in
```
ayrtonm, line 140: I don't think you can assume types are either scalars or structures. For reference [these docs](https://clang.llvm.org/doxygen/classclang_1_1Type.html) have a lot more `getAs*` methods. I'm not saying we have to explicitly handle every case, just pointing out that the assumption here may be wrong.

sim-immunant:
I'm closing this PR and moving the work into #333.

PR #347, opened 2024-04-08
Don't fail fast if tracing-enabled tests fail

### LINKED ISSUE #339: Avoid cancelling test run if tests with tracer fail
PR #329 added `-DIA2_TRACER` to the CMake config in the test matrix to run the test suite with and w/o the syscall tracer. This was partially done to avoid a transient error in the syscall tracer #324 which we're looking into. This is also good to have in general since the syscall tracer doesn't yet have an exhaustive list of the syscalls/memory mapping operations we can support.

However even though half our tests should now be more reliable, the test run is still cancelled if any of the tracer tests happen to run and fail before the non-tracer tests finish. Ideally we should have the non-tracer tests run first, then run them with the tracer.

thedataking:
Close #339.

Non-tracing-enabled runs will now continue but the overall run will still fail if a tracing-enabled one does. LMK if this is the desired functionality or not.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,9 @@ on:
 jobs:
   test-on-pkru-enabled-host:
     runs-on: self-hosted
+    continue-on-error: ${{ matrix.ia2-tracer == 'ON' }}
     strategy:
+      fail-fast: true
       matrix:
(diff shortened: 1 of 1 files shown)


ayrtonm (approved):
LGTM

PR #349, opened 2024-04-16
Rewriter cleanups

fw-immunant:
Split out of lib-mode work so we can merge them eagerly. Most commits are semantics preserving, a few are bug fixes.

### DIFF
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -883,40 +883,8 @@ static void write_to_ld_file(llvm::raw_fd_ostream *file[MAX_PKEYS], int i,
   *file[i] << contents;
 }
 
-int main(int argc, const char **argv) {
-#if LLVM_VERSION_MAJOR >= 13
-  auto parser_ptr =
-      CommonOptionsParser::create(argc, argv, SourceRewriterCategory);
-  if (!parser_ptr) {
-    auto error = parser_ptr.takeError();
-    llvm::errs() << error;
-    return 1;
-  }
-  CommonOptionsParser &options_parser = *parser_ptr;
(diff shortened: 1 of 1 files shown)


tools/rewriter/SourceRewriter.cpp
```diff
@@ -1029,15 +1029,20 @@ int main(int argc, const char **argv) {
...
+  // not try to rewrite our own implementation details
   tool.appendArgumentsAdjuster([&](const CommandLineArguments &args, llvm::StringRef filename) {
     CommandLineArguments new_args(args);
     // Try to remove existing definition from command line to avoid warnings.
     new_args.erase(std::remove_if(new_args.begin(), new_args.end(),
                                   [](std::string &x) { return x.starts_with("-DIA2_ENABLE="); }),
                    new_args.end());
-    new_args.push_back("-DIA2_ENABLE=0"s);
+    // Insert prior to the "end of flags" double hyphen; if not present, append
+    auto double_hyphen_pos = std::find(new_args.begin(), new_args.end(), "--");
```
ayrtonm, line 1042: Are you sure this vector of arguments includes `--` and the arguments before it? I thought clang just took arguments from the compile commands JSON or those after the `--` instead if it's used.

fw-immunant: In some situations (e.g. if `-p` is not passed to the writer but a `compile_commands.json` is found and Clang synthesizes an argv from it), this argument vector will contain `--`. In cases where `--` is not present, this still does the right thing.

I really hate Clang's argv handling.

ayrtonm:
LGTM, thanks for splitting it up by commit

ayrtonm (approved):

PR #353, opened 2024-05-24
cmake: Disable check-ia2 on ARM and support for running tests in QEMU

ayrtonm:
~~This adds a new `check-exec` CMake target that just runs the tests w/o running LIT tests. The `check` target retains its old behavior by running this and the existing `check-ia2` target.~~

This disables the check-ia2 target on arm and adds support for running QEMU with the correct arguments. Disabling check-ia2 was necessary because we omit some test directories on arm for various reasons which caused LIT tests for those directories to always fail.

It also modifies the `add_test` invocations on arm to use qemu with the prebuilt glibc/ld.so. Setting `CMAKE_CROSSCOMPILING_EMULATOR` sets each target's `CROSSCOMPILING_EMULATOR` property which `add_test` should use automatically but for whatever reason it isn't using it so I went ahead and implemented this manually. ~~Also the QEMU path is hard-coded but~~ we should set this through an env var or something and have it default to a build of our patched QEMU on donna.


ayrtonm:
I think adding the `check-exec` target broke x86 CI because of the interaction between `add_test`, the child processes it creates and the tracer. I really would like to avoid arch-specific CMake targets, but if figuring out what's happening turns out to be too difficult because of CMake/CTest, then making `check-exec` arm-only (and keeping `check`'s old definition on x86) should be a viable option.

ayrtonm:
structs test is cause the rewriter to segfault like this
```
(gdb) bt
#0  0x00007ffff56c1c14 in clang::TagType::getDecl() const () from /usr/lib/libclang-cpp.so.17
#1  0x0000555555588112 in clang::RecordType::getDecl (this=0x0) at /usr/include/clang/AST/Type.h:4881
#2  0x000055555558541e in classifyDirectType (type=..., astContext=...) at /home/ayrton/immunant/IA2-Phase2/tools/rewriter/DetermineAbi.cpp:51
#3  0x0000555555585606 in classifyDirectType (type=..., astContext=...) at /home/ayrton/immunant/IA2-Phase2/tools/rewriter/DetermineAbi.cpp:78
#4  0x0000555555585c8e in abiSlotsForArg (qt=..., argInfo=..., astContext=..., arch=Arch::Aarch64) at /home/ayrton/immunant/IA2-Phase2/tools/rewriter/DetermineAbi.cpp:190
#5  0x0000555555585ff5 in determineAbi (info=..., astContext=..., arch=Arch::Aarch64) at /home/ayrton/immunant/IA2-Phase2/tools/rewriter/DetermineAbi.cpp:240
#6  0x000055555558655c in determineAbiForDecl (fnDecl=..., arch=Arch::Aarch64) at /home/ayrton/immunant/IA2-Phase2/tools/rewriter/DetermineAbi.cpp:311
#7  0x00005555555c2800 in FnDecl::run (this=0x7fffffffd8a0, result=...) at /home/ayrton/immunant/IA2-Phase2/tools/rewriter/SourceRewriter.cpp:879
```
so I assume @sim-immunant's rewriter changes should fix this

ayrtonm:
Note that the qemu from ubuntu repository is too old to recognize the `-one-insn-per-tb` flag and uses the `-singlestep` alias instead while our forked qemu is too new to accept the deprecated `-singlestep` flag. So the latest CI failure on this PR will not actually be an issue.

fw-immunant (approved):
This seems to be working now. I was worried about CI times inflating due to the uncached QEMU build but as it's only a 73 seconds or so I think we can address caching later.

I'll squash merge to keep these changes self-contained as the history isn't super clean and a number of the CMake changes are ones we want to eventually resync with x86_64 when ARM support is more mature.

PR #354, opened 2024-05-24
spoofed_criterion: Support non-zero exit code and enable should_segafult test on ARM

ayrtonm:
This does not handle forking correctly so we need to combine it with #343


fw-immunant:
This is pretty trivial and just improves our test coverage. Merging.

PR #356, opened 2024-05-28
libia2: Allow defining compartment-aware signal handlers on ARM and re-enable test

ayrtonm:
The sighandler test now builds and runs. It fails at a `CHECK_VIOLATION` (as expected) because MTE is not enabled. I don't think it makes much sense to enable MTE before adding in real call gates (we currently just jump to the target function) so this and the tests in the other PRs should get enabled after #333 lands.


fw-immunant:
This is a pretty straightforward port of our x86_64 signal handler behavior to ARM/MTE: it applies the specified tag and jumps to the signal handler function.

PR #358, opened 2024-06-04
ci: cache built QEMU

fw-immunant:
this only saves us about a minute of CI time, but I'd rather debug our first CI caching against a 1-minute QEMU build than Clang's interminable build


fw-immunant:
I cancelled other tests but the relevant one (aarch64-wip-test) passed, and on the second run it successfully found the cached QEMU build; merging.

PR #359, opened 2024-06-04
ci: cache apt packages

fw-immunant:
rav1d does this and it seems like a good idea; maybe it'll shave a few more seconds off of CI


fw-immunant:
This seems to save most of the time from the "installing packages" step. I cancelled test jobs not touched by this CI change.

PR #360, opened 2024-06-04
ci: build+cache Clang

fw-immunant:
First some `ci.yml` tweaks to QEMU that we'll want for Clang, then the Clang build itself.


fw-immunant:
When building clang we come in under the disk limit on the GH runner by the skin of our teeth. Now that clang has built it should be cached for subsequent PRs.

PR #362, opened 2024-06-13
Port runtime to ARM

fw-immunant:
The tracer changes here add relatively clean support for compile-time choice of architecture. The partition-alloc changes here are enough to get things building, but don't implement tagging with MTE instructions and will need some revisiting.


fw-immunant:
Currently blocked on #373.

PR #373, opened 2024-07-23
Flesh out `spoofed_criterion` for AArch64

fw-immunant:
Fixes some failing tests, not-building tests, ~and spuriously passing tests~ on AArch64. The tests depend on other work in addition to these fixes before they'll actually pass on AArch64, but this fills in missing testing infrastructure.

EDIT: I've split the `test_fault_handler` changes into a separate PR (#378) because they uncover brokenness that we need to fix separately while most of this infrastructure is strictly additive and needed to run more AArch64 tests.

### DIFF
--- a/misc/spoofed_criterion/include/criterion/criterion.h
+++ b/misc/spoofed_criterion/include/criterion/criterion.h
@@ -3,6 +3,7 @@
 
 struct fake_criterion_test {
   void (*test)(void);
+  void (*init)(void);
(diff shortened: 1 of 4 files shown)


fw-immunant:
Based on bisecting it appears the three_keys_minimal test has never worked but its failures have always been masked by the test_fault_handler bug that this fixes. I'll look into things more tomorrow.

Issue #368, opened 2024-07-23
System headers missing when using `meson` (`dav1d`)


kkysen:
When using `meson`, like for `dav1d`, system headers appear to be missing.

For example, after running

```sh
git clone https://code.videolan.org/videolan/dav1d.git
cd dav1d
mkdir build
cd build
meson setup ..
```

I get this output:

```shell
> ../ia2/build/tools/rewriter/ia2-rewriter --output-prefix call_gates --root-directory ./ --output-directory ia2/ -p build src/data.c
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
In file included from ../src/data.c:32:
/usr/include/stdlib.h:32:10: fatal error: 'stddef.h' file not found
   32 | #include <stddef.h>
      |          ^~~~~~~~~~
looking at commands for src/data.c
No compilation command found for src/data.c
```

Or this:

```shell
> ../ia2/build/tools/rewriter/ia2-rewriter --output-prefix call_gates --root-directory ./ --output-directory ia2/ -p build src/decode.c
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
In file included from ../src/decode.c:31:
/usr/include/limits.h:124:16: fatal error: 'limits.h' file not found
  124 | # include_next <limits.h>
      |                ^~~~~~~~~~
looking at commands for src/decode.c
No compilation command found for src/decode.c
```

An example of one of the compile commands that is causing this:

```json
{
[...]

PR #377, opened 2024-07-24
rewriter: skip copying files if the output file is the same file to avoid truncating them

### LINKED ISSUE #367: `copy_files` truncates all `#include`d files in `dav1d`
When trying to run `ia2-rewriter` on [`dav1d`](https://code.videolan.org/videolan/dav1d), it truncates (to 0 length) all of the files `#include`d from the file it was run on (including that file, too).

I ran this:

```sh
git clone https://code.videolan.org/videolan/dav1d.git
cd dav1d
mkdir build
cd build
meson setup ..
../ia2/build/tools/rewriter/ia2-rewriter --output-prefix call_gates --root-directory ./ --output-directory ia2/ -p build src/data.c
```

The result is this:

```shell
> ../ia2/build/tools/rewriter/ia2-rewriter --output-prefix call_gates --root-directory ./ --output-directory ia2/ -p build src/data.c
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
In file included from ../src/data.c:32:
/usr/include/stdlib.h:32:10: fatal error: 'stddef.h' file not found
   32 | #include <stddef.h>
      |          ^~~~~~~~~~
looking at commands for src/data.c
No compilation command found for src/data.c

> git diff --stat
 include/common/attributes.h | 203 ---------------------------------------------------------------------
 include/common/validate.h   |  61 ---------------------
 include/dav1d/common.h      |  94 --------------------------------
 include/dav1d/data.h        | 117 ----------------------------------------
 include/dav1d/dav1d.h       | 329 ----------------------------------------------------------------------------------------------------------------
 include/dav1d/headers.h     | 444 -------------------------------------------------------------------------------------------------------------------------------------------------------
 include/dav1d/picture.h     | 157 ------------------------------------------------------
 src/data.c                  | 149 ---------------------------------------------------
 src/data.h                  |  56 -------------------
 src/mem.h                   | 137 -----------------------------------------------
 src/ref.h                   |  77 ---------------------------
 src/thread.h                | 189 ----------------------------------------------------------------
 12 files changed, 2013 deletions(-)
```

This still happens if I pass `--extra-arg -isystem`, which fixes the `#include <stddef.h>` error.

Also note that errors like

```shell
looking at commands for src/decode.c
No compilation command found for src/decode.c
```

occur after `copy_files` runs and truncates `src/decode.c`.

kkysen:
* Fixes #367.

This fixes #367 in that the files are no longer truncated, but a clearer error message is printed instead.  This doesn't fix the root cause of #367, which probably has more to do with #368 and the `compile_commands.json` not getting read correctly, but it does fix the alarming behavior of deleting the contents of a ton of files.

### DIFF
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -994,9 +994,11 @@ std::set<llvm::SmallString<256>> copy_files(std::vector<std::unique_ptr<clang::A
           assert(mkdir_rc == 0);
         }
 
-        std::ifstream src(input_file.str().str(), std::ios::binary);
-        std::ofstream dst(output_file.str().str(), std::ios::binary);
-        dst << src.rdbuf();
+        if (llvm::sys::fs::equivalent(input_file, output_file)) {
+          llvm::errs() << "skipping copying file to itself: " << input_file << "\n";
+          continue;
+        }
+        llvm::sys::fs::copy_file(input_file, output_file);
       }
     }
   }


fw-immunant (approved):

fw-immunant:
There might be a prior invariant we're violating if we actually try to copy files from one location to itself (I don't think inputs and outputs should overlap in this way), but this does at least solve the symptom while whining loudly even if the underlying cause is still present. LGTM.

PR #378, opened 2024-07-24
test_fault_handler: exit nonzero if expected segfault missing

fw-immunant:
We previously tested for the printed message with FileCheck, but now our Criterion-based tests only care about exit codes.

This was hiding test failures on both AArch64 and x86_64.

Based on bisecting it appears the `three_keys_minimal` test has never worked but its failures have always been masked by the `test_fault_handler` bug that this fixes. I'll look into things more tomorrow.

### DIFF
--- a/runtime/libia2/include/test_fault_handler.h
+++ b/runtime/libia2/include/test_fault_handler.h
@@ -29,7 +29,7 @@
     asm volatile("" : : : "memory");                                           \
     volatile typeof(expr) _tmp = expr;                                         \
     printf("CHECK_VIOLATION: did not seg fault as expected\n");                \
-    _exit(0);                                                                  \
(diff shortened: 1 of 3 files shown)


kkysen (approved):
This makes sense to me from what I can tell.

kkysen (commented):
Did you also want to disable `three_keys_minimal` in CI until it's fixed?

fw-immunant:
I think we can wait to merge this until we figure out what's wrong with that test and `heap_two_keys`.

rinon:
Do we have an issue on `three_keys_minimal` and `heap_two_keys`? If we're not working on that currently I think we should disable those tests and land this PR.

rinon:
It looks like we need to delete the build directory in CI? #418

PR #380, opened 2024-07-24
ci: try freeing space up by deleting LLVM libs after building Clang

fw-immunant:
Running out of disk space to create the cache archive may be why our caching is not working.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,6 +130,8 @@ jobs:
           df -h /
           cd -
           ./build.sh
+          # free up space used by libs; our desired clang binary statically links them
+          rm -rf build/lib/*.a
       - name: Test ARM build
         run: |
           LLVM_DIR=`llvm-config-15 --cmakedir`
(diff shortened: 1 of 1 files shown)


fw-immunant:
Smoking gun (if this is it) [here](https://github.com/immunant/IA2-Phase2/actions/runs/10056959690/job/27796919909?pr=370#step:18:3):
```
Post job cleanup.
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/IA2-Phase2/IA2-Phase2 --files-from manifest.txt --use-compress-program zstdmt
zstd: error 70 : Write error : cannot write block : No space left on device 
/usr/bin/tar: cache.tzst: Wrote only 8192 of 10240 bytes
/usr/bin/tar: Child returned status 70
/usr/bin/tar: Error is not recoverable: exiting now
Warning: Failed to save: "/usr/bin/tar" failed with error: The process '/usr/bin/tar' failed with exit code 2
```

kkysen (commented):
Coincidentally, would zstd help compressing the binaries?  Or should there be plenty of space left now that the other libs are deleted?

PR #382, opened 2024-07-24
ci: test building Clang in release mode

fw-immunant:
Our custom Clang seems very slow and this may be why.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,7 @@ jobs:
           path: |
             llvm-project/build/bin
             llvm-project/build/lib/clang
-          key: ${{ runner.os }}-clang
+          key: ${{ runner.os }}-clang-release
       - name: Check out Clang
         if: steps.cache-clang.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -131,6 +131,7 @@ jobs:
           '!' -path ./runner '!' -path ./runner-cache -exec rm -rf '{}' ';'
           df -h /
           cd -
+          sed -i -r -e s/Debug/RelWithDebInfo/ build.sh
```
kkysen, line 134: What's the `build.sh` script?  Is there a way to directly set the build type?  Because this seems like it could change some other `Debug`s that happen to be in the script.

fw-immunant: It's https://github.com/immunant/llvm-project/blob/am/mte_instrumentation/build.sh which I didn't want to touch on that branch because it's used by all branches of this project. If this change helps I'll patch the script over there and invalidate the cache on this side.

fw-immunant:
This seems to have worked; I've committed the build script change to our LLVM fork.

Issue #324, opened 2024-03-11
Transient test failure (trusted_direct mmap interferes with existing allocation)


fw-immunant:
This [CI run](https://github.com/immunant/IA2-Phase2/actions/runs/8175479087/job/22352689895) failed for commit [8fff09a](https://github.com/immunant/IA2-Phase2/commit/8fff09aa1124e3b874808d5a20cb22b5b6d7812d), but it seems to succeed most of the time. I wasn't able to reproduce when I tried.

Full log:
<details>

```
Run pushd build
~/actions-runner/_work/IA2-Phase2/IA2-Phase2/build ~/actions-runner/_work/IA2-Phase2/IA2-Phase2
[0/2] cd /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/rewriter/tests && /mnt/ssd2/home/perl/.local/bin/lit --verbose -sv /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/rewriter/tests
lit: /mnt/ssd2/home/perl/.local/lib/python3.10/site-packages/lit/llvm/config.py:459: note: using clang: /usr/lib/llvm-14/bin/clang
-- Testing: 46 tests, 24 workers --
Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90.. 

Testing Time: 0.09s
  Passed: 46
[2/2] cd /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/rewriter/tests && /usr/bin/ctest --output-on-failure
FAILED: rewriter/tests/CMakeFiles/check /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/rewriter/tests/CMakeFiles/check 
cd /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/rewriter/tests && /usr/bin/ctest --output-on-failure
Test project /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/rewriter/tests
      Start  1: destructors
 1/25 Test  #1: destructors ......................   Passed    0.13 sec
      Start  2: header_includes
 2/25 Test  #2: header_includes ..................   Passed    0.09 sec
      Start  3: heap_two_keys
[...]

PR #383, opened 2024-07-26
rewriter: use `llvm::sys::fs::create_directories` instead of doing the logic ourselves

kkysen:
Also, ignore existing directories when copying files.

Also, do the skipping if the file is the same before anything else (like creating the directories, or marking the file as copied).

### DIFF
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -974,7 +974,7 @@ std::set<llvm::SmallString<256>> copy_files(std::vector<std::unique_ptr<clang::A
       if (needs_copy && !copied_files.contains(input_file)) {
         copied_files.insert(input_file);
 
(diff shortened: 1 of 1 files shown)


kkysen:
I'm not sure why this is causing CI errors like

```
[74caf437e000, +00081000) interferes with [74caec3ff000, +08000000)
could not update memory map! (operation=MMAP, rip=0x74caf691ea27)
```

fw-immunant:
> I'm not sure why this is causing CI errors like
> 
> ```
> [74caf437e000, +00081000) interferes with [74caec3ff000, +08000000)
> could not update memory map! (operation=MMAP, rip=0x74caf691ea27)
> ```
> 
> since this PR seems pretty unrelated.

That's #324, feel free to re-run those CI jobs.

fw-immunant (approved):

PR #384, opened 2024-07-26
Improve error for invalid pkey range

randomPoison:
Add an error message in the case where pkeys are setup incorrectly so that users can more easily debug their pkey setup.


fw-immunant (approved):
This is much more user-friendly, thanks.

PR #385, opened 2024-07-29
Disable more header defs when IA2_ENABLE=0

randomPoison:
This fixes issues I ran into when trying to run the source rewriter on zlib (https://github.com/immunant/zlib). `ia2.h` was not importing `ia2_internal.h` when `IA2_ENABLE=0`, but `ia2_compartment_init.inc` was importing it unconditionally. It seems like we don't want `ia2_internah.h` imported when running the source rewriter, and the rewriter specifically sets `-DIA2_ENABLE=0` to make sure that IA2 internals don't get included in the source files.

I then also hit a similar issue with some of the symbols defined in `ia2.h`, so I moved those inside the `#if IA2_ENABLE` block so that they won't be defined when the source rewriter runs.

### DIFF
--- a/runtime/libia2/include/ia2.h
+++ b/runtime/libia2/include/ia2.h
@@ -142,7 +142,6 @@
 /// the type of `IA2_FN(func)` are ABI-compatible since no extra type-checking is
 /// done.
 #define IA2_CAST(func, ty) (ty) { (void *)IA2_FN_ADDR(func) }
-#endif // !IA2_ENABLE
 
 /// Convert a compartment pkey to a PKRU register value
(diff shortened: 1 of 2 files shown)


rinon:
It _should_ have been safe to include `ia2_internal.h` regardless, what did you hit? I think it's ok to not include it though.

Re: the symbols, were you adding `-lia2` to the zlib build? You'll need to depend on libia2 regardless, so you should probably link against it in either mode.

randomPoison:
I went back to try to reproduce this and I'm no longer hitting these errors, so it probably was caused by something I was doing wrong. I've opened tickets for the various undocumented build steps I uncovered, so hopefully addressing those fixes the issue I hit here.

PR #387, opened 2024-08-29
Fix struct copying for structs with an unaligned size

### LINKED ISSUE #365: Structs test fails on Arch (LLVM 18)
```
[----] ./tests/structs/main.c:143: Assertion Failed
[----]
[----]   The expression s.i1 == 1 + 100 + 8043 / 2 is false.
[----]
[FAIL] structs::main: (0.00s)
[====] Synthesis: Tested: 1 | Passing: 0 | Failing: 1 | Crashing: 0
```

rinon:
Clang started packing structs immediately after
stack arguments, but we used to copy memory
arguments as 8-byte chunks. Instead, copy only up
to the size of the struct argument to avoid
overwriting other bytes above it on the stack.

Fixes #365

### DIFF
--- a/tools/rewriter/DetermineAbi.cpp
+++ b/tools/rewriter/DetermineAbi.cpp
@@ -20,8 +20,8 @@
 
 // Compute sequence of eightbyte classifications for a type that Clang has
 // chosen to pass directly in registers
-static std::vector<CAbiArgKind> classifyDirectType(const clang::Type &type,
-    const clang::ASTContext &astContext) {
+static std::vector<CAbiArgLocation> classifyDirectType(const clang::Type &type,
+                                               const clang::ASTContext &astContext) {
   if (type.isVoidType()) {
     return {};
   } else if (type.isScalarType()) {
@@ -30,9 +30,9 @@ static std::vector<CAbiArgKind> classifyDirectType(const clang::Type &type,
     case clang::Type::ScalarTypeKind::STK_Bool:
     case clang::Type::ScalarTypeKind::STK_Integral:
     case clang::Type::ScalarTypeKind::STK_FixedPoint:
-      return {CAbiArgKind::Integral};
+      return {CAbiArgLocation{CAbiArgKind::Integral}};
     case clang::Type::ScalarTypeKind::STK_Floating:
-      return {CAbiArgKind::Float};
+      return {CAbiArgLocation{CAbiArgKind::Float}};
     case clang::Type::ScalarTypeKind::STK_IntegralComplex:
     case clang::Type::ScalarTypeKind::STK_FloatingComplex:
       llvm::report_fatal_error(
@@ -54,8 +54,8 @@ static std::vector<CAbiArgKind> classifyDirectType(const clang::Type &type,
     const clang::RecordDecl *decl = rec->getDecl();
 
     if (decl->canPassInRegisters()) {
-      std::vector<CAbiArgKind> out; // Classifications for the entire record
-      std::optional<CAbiArgKind> pending_kind;
+      std::vector<CAbiArgLocation> out; // Classifications for the entire record
+      std::optional<CAbiArgLocation> pending_location;
       int64_t prev_end_offset = 0; // Initially, first eightbyte
 
       const clang::ASTRecordLayout &layout =
@@ -69,9 +69,9 @@ static std::vector<CAbiArgKind> classifyDirectType(const clang::Type &type,
         // Save the classification of the last eightbyte if we've left it
         bool same_eightbyte = offset / 64 == prev_end_offset / 64;
         if (!same_eightbyte) {
-          assert (pending_kind.has_value());
-          out.push_back(*pending_kind);
-          pending_kind.reset();
+          assert(pending_location.has_value());
+          out.push_back(*pending_location);
+          pending_location.reset();
         }
 
         // Update pending_kind based on ABI rules.
@@ -83,24 +83,24 @@ static std::vector<CAbiArgKind> classifyDirectType(const clang::Type &type,
           llvm::report_fatal_error(
               "unexpectedly classified register-passable field as multiple eightbytes");
         }
-        auto new_kind = recur[0];
+        auto new_location = recur[0];
         // This block sets pending_kind = new_kind regardless.
         // However we format it this way to match §3.2.3.4 of the x86_64 ABI.
         // In the future, if we add support for X87 types etc, this logic will
         // be more complex.
-        if (pending_kind) {
-          if (*pending_kind != new_kind) {
-            if (new_kind == CAbiArgKind::Memory) {
-              pending_kind = CAbiArgKind::Memory;
-            } else if (new_kind == CAbiArgKind::Integral) {
-              pending_kind = CAbiArgKind::Integral;
-            // TODO: handle X87/X87UP/COMPLEX_X87
+        if (pending_location) {
+          if (pending_location->kind != new_location.kind) {
+            if (new_location.kind == CAbiArgKind::Memory) {
+              pending_location = new_location;
+            } else if (new_location.kind == CAbiArgKind::Integral) {
+              pending_location = new_location;
+              // TODO: handle X87/X87UP/COMPLEX_X87
             } else {
-              pending_kind = new_kind;
+              pending_location = new_location;
             }
           }
         } else {
-          pending_kind = new_kind;
+          pending_location = new_location;
         }
 
         // Update prev_end_offset for next iteration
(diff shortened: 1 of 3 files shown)


tools/rewriter/GenCallAsm.cpp
```diff
@@ -184,6 +189,19 @@ static void emit_reg_pop(AsmWriter &aw, const ParamLocation &loc) {
   }
 }
 
+// Emit code to copy `byte_count` bytes from `src` to `dst` using `scratch` as a
+// temporary register.
+static void emit_memcpy(AsmWriter &aw, unsigned byte_count, const std::string &dst,
```
fw-immunant, line 194: It's probably worth calling out that this copies a rounded-up multiple of 8-byte chunks rather than `byte_count` exactly.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -391,27 +409,31 @@ static void emit_copy_args(AsmWriter &aw, size_t stack_return_size, size_t stack
...
-        // The index into the caller's stack is backwards since pushq will copy to
-        // the compartment's stack from the highest addresses to the lowest.
-        add_asm_line(aw,
-                     "pushq " + std::to_string(caller_stack_size - i) + "(%rax)");
-      }
+      // We must take the preserved registers we pushed on the caller's stack
+      // into account (including rbp and the return address) when determining
+      // the location of the stack args.
+      size_t offset = (preserved_registers.size() + 2) * 8;
+      emit_memcpy(aw, stack_arg_size, "rsp", "rax", "r10", offset);
```
fw-immunant, line 425: Don't split up construction and use of `expr`.

Is it preferable to use `mov`s and a separate modification of `rsp` here as opposed to the `push`es used before, or is there another reason to change this?

If we do use `mov` and a temporary, we should probably reuse `r12` which we just clobbered.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -391,27 +409,31 @@ static void emit_copy_args(AsmWriter &aw, size_t stack_return_size, size_t stack
...
+      //   // We must take the preserved registers we pushed on the caller's stack
+      //   // into account (including rbp) when determining the location of the stack
+      //   // args
+      //   size_t caller_stack_size =
+      //       stack_arg_size + ((preserved_registers.size() + 1) * 8);
+      //   // The index into the caller's stack is backwards since pushq will copy to
+      //   // the compartment's stack from the highest addresses to the lowest.
+      //   add_asm_line(aw,
+      //                "pushq " + std::to_string(caller_stack_size - i) + "(%rax)");
+      // }
```
fw-immunant, line 436: No need to leave this around in a comment.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -446,12 +468,12 @@ static void emit_set_pkru(AsmWriter &aw, uint32_t target_pkey, Arch arch) {
...
 }
 
-static void emit_free_stack_space(AsmWriter &aw, size_t stack_arg_size, int stack_alignment, Arch arch) {
+static void emit_free_stack_space(AsmWriter &aw, size_t stack_arg_size, size_t stack_arg_padding, int stack_alignment, Arch arch) {
   if (arch == Arch::X86) {
     // Free stack space used for stack args on the target stack
     if (stack_arg_size > 0) {
       add_comment_line(aw, "Free stack space used for stack args");
-      add_asm_line(aw, "addq $"s + std::to_string(stack_arg_size) + ", %rsp");
+      add_asm_line(aw, "addq $"s + std::to_string(stack_arg_size + stack_arg_padding) + ", %rsp");
```
fw-immunant, line 476: Is the padding here doing something other than rounding the arg size up to the passed alignment?

tools/rewriter/GenCallAsm.cpp
```diff
@@ -481,13 +503,27 @@ static void emit_copy_stack_returns(AsmWriter &aw, size_t stack_return_size, siz
...
+      }
+      if (i + 2 <= stack_return_size) {
+        add_asm_line(aw, "movw "s + std::to_string(i) + "(%rsp), %r10w");
+        add_asm_line(aw, "movw %r10w, "s + std::to_string(i) + "(%rax)");
+        i += 2;
       }
+      if (i < stack_return_size) {
+        add_asm_line(aw, "movb "s + std::to_string(i) + "(%rsp), %r10b");
+        add_asm_line(aw, "movb %r10b, "s + std::to_string(i) + "(%rax)");
+      }
```
fw-immunant, line 524: If the other memcpy gets factored out into its own function this one probably should too.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -657,13 +707,13 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
     start_of_ret_space += 8;
 
     // See what unpadded alignment would be for the start of the stack return.
-    size_t unaligned = start_of_ret_space % stack_return_align;
+    size_t unaligned = (start_of_ret_space + stack_return_size) % stack_return_align;
```
fw-immunant, line 710: This comment doesn't seem to make sense anymore: the start of the stack return wouldn't include its size.

fw-immunant (commented):
Various comments inline.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -11,18 +12,21 @@
...
 
 struct ParamLocation {
 private:
-  ParamLocation(const char *reg) : reg(reg) {}
+  ParamLocation(const char *reg, unsigned size, unsigned align) : reg(reg), _size(size), _align(align) {}
 
 public:
   const char *reg;
+  const unsigned _size;
+  const unsigned _align;
```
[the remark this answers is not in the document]
rinon: They were intended to be private, good catch.

tools/rewriter/DetermineAbi.cpp
```diff
@@ -254,9 +252,11 @@ CAbiSignature determineAbi(const clang::CodeGen::CGFunctionInfo &info,
...
     // Replace the integer slots with an equal number of memory (stack) slots
     std::erase_if(sig.ret, is_integral);
-    for (int i = 0; i < num_regs; i++) {
-        sig.ret.push_back(CAbiArgKind::Memory);
-    }
+    sig.ret.push_back(CAbiArg{
+        .kind = CAbiArgKind::Memory,
+        .size = static_cast<unsigned>(num_regs) * 8,
+        .align = 8,
+    });
```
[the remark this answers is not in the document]
rinon: I believe branch only handles multiple integral values. Strangely sized structs are handled in `abiSlotsForArg` above.

tools/rewriter/GenCallAsm.cpp
rinon: I originally thought we would copy the exact size, not a padded to 8 bytes size. I can revert this bit.

rinon:
Somehow this bitrotted and is failing again now? I'm investigating.

tools/rewriter/GenCallAsm.cpp
rinon: Reverted back to `pushq`s.

tools/rewriter/GenCallAsm.cpp
rinon: It now copies the exact number of bytes.

tools/rewriter/GenCallAsm.cpp
rinon: No, that's exactly what we're doing. `stack_arg_size` may not be a multiple of 8 now.

tools/rewriter/GenCallAsm.cpp
rinon: Good point, this make more sense?

rinon:
@fw-immunant I think I've addressed all your feedback, so this needs another look when you can.

tools/rewriter/GenCallAsm.cpp
fw-immunant: I guess my question is: if `stack_arg_padding` is just `round_up(stack_arg_size, 8) - stack_arg_size`, why is it a separate argument? Or am I misunderstanding?

fw-immunant:
> @fw-immunant I think I've addressed all your feedback, so this needs another look when you can.

One last question, otherwise all looks good.

tools/rewriter/GenCallAsm.cpp
rinon: It doesn't need to be and I actually change this in https://github.com/immunant/IA2-Phase2/commit/3caefe7baa423b4796e6f466dfff8ee4b4ca07ee but that's not landed yet.

rinon:
I ported over the free space simplification that is going to happen for ARM anyway. Should be good now?

fw-immunant (approved):

PR #388, opened 2024-09-06
rewriter: add script to canonicalize all paths in a `compile_commands.json`

### LINKED ISSUE #368: System headers missing when using `meson` (`dav1d`)

kkysen:
* Fixes #368.

This canonicalization of `compile_commands.json` paths ensures that `libclangTooling` and `ia2-rewriter` handle paths correctly when using relative and non-canonical paths, like `meson` generates for `dav1d`.

### DIFF
--- a/tools/rewriter/canonicalize_compile_command_paths.py
+++ b/tools/rewriter/canonicalize_compile_command_paths.py
@@ -11,6 +11,16 @@ import json
 from pathlib import Path
 
 def main():
+    """
+    Assuming we're in a build directory containing a `compile_commands.json`,
+    canonicalize (including making absolute) all paths in every compile command,
+    including as part of arguments (detect as a best effort).
+
+    libclangTooling is finicky with relative paths,
+    and it and other tools don't always handle non-canonical paths well,
+    so making all paths in a `compile_commands.json` canonical is helpful.
+    """
+
     cc_db = Path("compile_commands.json")
     cc_text = cc_db.read_text()
     cmds = json.loads(cc_text)
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -26,6 +26,22 @@ want to generate and use a `compile_commands.json` to ensure the rewriter
 preprocesses each source file with the same command-line arguments as when it is
 compiled.
 
+### Compile Command Paths
+
+Note that `libclangTooling`, which `ia2-rewriter` uses, is very finicky
+about relative paths, so we suggest making all relative paths absolute.
+Furthermore, `ia2-rewriter` itself sometimes uses paths as keys,
+so we suggest making all paths canonical.  We have a script,
+[`canonicalize_compile_command_paths.py`](../tools/rewriter/canonicalize_compile_command_paths.py),
+that does this automatically (doing its best to detect paths embedded in args).
+`libclangTooling` also requires compilation databases to be named exactly
+`compile_commands.json`, so this script assumes that as well,
+and must be run in the directory of the `compile_commands.json`, like this:
+
+```sh
+(cd $build_directory && $ia2_dir/tools/rewriter/canonicalize_compile_command_paths.py)
+```
+
 ## Manual source changes
 
 ### Defining compartments


tools/rewriter/canonicalize_compile_command_paths.py
```diff
@@ -0,0 +1,62 @@
...
+                path = dir / path
+            path = path.resolve()
+            return str(path)
+
+        cmd["file"] = convert_path(cmd["file"])
+        if "output" in cmd:
+            cmd["output"] = convert_path(cmd["output"])
+        
+        def try_convert_arg(arg: str) -> str:
+            for prefix in ["", "-I"]:
```
rinon, line 43: Don't you want this order reversed? All strings start with an empty prefix.

rinon (commented):
I'm uneasy about changing arbitrary command-line argument internals with this. Remind me what the problem you hit that needed this was? Just trying to figure out if there's another way to resolve whatever that was. Can we get away without that part and just canonicalize the input/output paths?

tools/rewriter/canonicalize_compile_command_paths.py
kkysen: Is the order important here?  I'm just checking if the whole argument is a path.  And then checking if, minus a `-I` prefix, the rest of the argument is a path.

rinon: Checking the other direction is half the work (assuming the `-I` matches). It doesn't matter much.

kkysen: Yeah, but we don't know which will match.  Performance doesn't really matter here, and it's Python anyway, so as long as it's correct, I'll just keep it as is.

tools/rewriter/canonicalize_compile_command_paths.py
```diff
@@ -0,0 +1,62 @@
...
+        def convert_path(path: str) -> str:
+            path: Path = Path(path)
+            if not path.is_absolute():
+                path = dir / path
+            path = path.resolve()
+            return str(path)
+
+        cmd["file"] = convert_path(cmd["file"])
+        if "output" in cmd:
+            cmd["output"] = convert_path(cmd["output"])
```
ayrtonm, line 40: Do we need to convert output paths?

kkysen:
@rinon, I tried disabling canonicalization of paths inside `command` and `arguments`, but I ran into some errors, so it seems like it's still needed.

The error seems to be

```
pkey not known for function allocate_stack
```

though I'm not sure if there are more (a lot of other stuff is printed, too).  A lot of the paths that are printed while rewriting are slightly different.  That might just be different paths to the same file (i.e. non canonicalized), but it does end up with an error.

tools/rewriter/canonicalize_compile_command_paths.py
kkysen: I don't believe so, since the field is optional in the first place (and the rewriter still seems to work without doing this), but since it's an easy field to rewrite, as we don't have to parse it unlike `arguments` and `command`, I think it's better to just canonicalize this path as well.

rinon (approved):

PR #392, opened 2024-09-20
rewriter: avoid segfaulting on non-scalar, non-struct direct types

### LINKED ISSUE #390: rewriter: segfault in `classifyDirectType` for non-scalar, non-struct types

kkysen:
* Fixes #390.

This doesn't actually handle the type, but it avoids segfaulting and dumps the unexpected type. In this case, it happens to be a `union`, which we need to handle I believe, but we can do that in a follow-up PR.


fw-immunant (approved):

fw-immunant:
> @fw-immunant, can you let me merge my own PRs? This one was stacked and wasn't meant to be merged as is yet.

Oh, my bad. I didn't realize it was stacked, but I don't need to be trigger-happy with merges regardless. Will do.

PR #397, opened 2024-09-27
Fix indirect call stack return handling

### LINKED ISSUE #396: Indirect calls that return a value on the stack are broken
It looks like indirect function call returns break when there is also a return value region on the stack.

rinon:
We don't save the function pointer to the stack anymore, so we shouldn't be adding a slot in the indirect call wrapper case. This wasn't breaking anything unless we had a return value on the stack.

Fixes #396


tools/rewriter/GenCallAsm.cpp
```diff
@@ -646,10 +646,6 @@ std::string emit_asm_wrapper(const CAbiSignature &sig,
   size_t stack_return_align = 16;
   size_t stack_return_padding = 0;
   size_t start_of_ret_space = 0;
-  if (kind == WrapperKind::IndirectCallsite) {
```
ayrtonm, line 649: nit: we should also remove `ret ptr` from the comment above this

ayrtonm:
The part where you remove the stack slot looks good to me. I'm not sure if we need the `.set << wrapper_name ...` though. I thought that was a case where the rewriter shouldn't emit annotations in the first place but I've only briefly skimmed that part so far.

tools/rewriter/GenCallAsm.cpp
ayrtonm: oh wait `ret ptr` is for the return value. nvm

tools/rewriter/SourceRewriter.cpp
```diff
@@ -1421,6 +1421,10 @@ int main(int argc, const char **argv) {
       wrapper_out << "asm(\n";
       wrapper_out << asm_wrapper;
       wrapper_out << ");\n";
+    } else {
+      header_out << "asm(\n";
+      header_out << "  \".set " << wrapper_name << ", __real_" << fn_name << "\\n\"\n";
```
ayrtonm, line 1426: I think this branch is dead code and the TODO on line 1413 should be deleted. I mean if you take the address of a static function that runs in compartment 0 then you must always be in compartment 0 (since it's static) and the rewriter doesn't add annotations to source files in compartment 0, right?

ayrtonm: also if it isn't dead code wouldn't we want to add this in `wrapper_out` instead of the header?

rinon: This isn't for static functions, this is for taking the address of a function in compartment 0 from compartment 1. The new ABI test has an example of this, I needed this to get that to even compile.

rinon: `ret_in_memory` is in compartment 0, so for us to call `IA2_FN(ret_in_memory)` we need a wrapper, but the wrapper is a no-op so we just need to provide a symbol alias.

ayrtonm: sorry I was building the wrong targets when I went to check if it was dead code (still getting the hang of things again). That all makes sense. Also having it in `header_out` is fine since that gives simpler control flow than having a no-op call gate in the callgates .so and it shouldn't be an issue that it's in a header since the symbol name already can't be duplicated in different translation units.

ayrtonm (approved):
LGTM

ayrtonm:
oh there's a TODO comment we can remove above the `.set` directive

PR #398, opened 2024-09-30
rewriter: warn on multiple compile commands per src and truncate to first one

### LINKED ISSUE #393: rewriter: `assert(comp_cmds.size() == 1)` fails on templated files like in `dav1d`
`dav1d` uses templated source files that are compiled multiple times, but `ia2-rewriter` currently has an `assert(comp_cmds.size() == 1)`.  It might be good enough to just turn this into a warning instead and run things on the first compile command.

kkysen:
* Fixes #393.

We discussed this separately, and besides just cases like `dav1d` where this is done for templating reasons, this is also done in other places, I think for static and dynamic libraries maybe?  Given that, it's simpler to just use the first compile command and emit a warning about it.  We already emit warnings on other things that might be an issue (e.x. unsupported variadic function pointers), so I think this is okay.

### DIFF
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -1029,7 +1029,14 @@ std::optional<Pkey> pkey_from_commands(std::function<std::optional<std::vector<C
     return {};
   }
 
+  if (comp_cmds.size() != 1) {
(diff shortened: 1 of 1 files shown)


rinon (approved):

kkysen:
Ugh, oops I merged it into the wrong branch again 😭.

PR #400, opened 2024-09-30
runtime: use builtin `__{x86_64, aarch64}__` instead of `LIBIA2_{X86_64,AARCH64}`

kkysen:
After seeing @randomPoison run into this same issue that I ran into as well, and since @fw-immunant said we probably don't need our own `LIBIA2_{X86_64,AARCH64}` macros and can use the builtin `__{x86_64,aarch64}__` macros, I thought it'd be simpler to fix this here in ia2 vs. handle defining them in the correct places in `dav1d`'s `meson.build`s.

`dav1d` also uses `-Wundef`, so `#if`/`#elif`s directly on `__{x86_64,aarch64}__` warn about being undefined on the other architecture, so I changed them all to `#if`/`#elif` `define(...)`.  I could use `#ifdef ...` and then `#elif defined(...)`, but for consistency between the two, I just stuck to using `defined` in both places.

### DIFF
--- a/cmake/ia2.cmake
+++ b/cmake/ia2.cmake
@@ -1,4 +1,4 @@
-if(LIBIA2_AARCH64)
+if (LIBIA2_AARCH64)
     set(UBSAN_FLAG "")
 else()
     set(UBSAN_FLAG "-fsanitize=undefined")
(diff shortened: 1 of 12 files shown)


fw-immunant (approved):
I was going to get around to this myself if nobody else did. Thanks!

PR #407, opened 2024-10-03
Fix stack tagging on AArch64

fw-immunant:
~We weren't tagging the entire requested region but only its first page (which also impacted protection of non-stack pages).~ (Fixed separately.)

After allocating and tagging a fresh compartment stack, we returned an untagged pointer to it, which led to MTE violations. I was surprised by this because the violations I observed seemed to come from `sp`-relative accesses, which I thought would not perform MTE checks, but they did anyway--it would be good to write a test program to characterize exactly which instructions do and don't perform MTE checks.

In addition, we need to apply tags for the access to TLS when accessing `ia2_stackptr_N`.

### DIFF
--- a/runtime/libia2/init.c
+++ b/runtime/libia2/init.c
@@ -28,6 +28,10 @@ char *allocate_stack(int i) {
       exit(-1);
     }
   }
+#ifdef __aarch64__
+  /* Tag the allocated stack pointer so it is accessed with the right pkey */
+  stack = (char *)((uint64_t)stack | (uint64_t)i << 56);
+#endif
   /* Each stack frame start + 8 is initially 16-byte aligned. */
   return stack + STACK_SIZE - 8;
 }
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -281,11 +281,12 @@ asm(".macro movz_shifted_tag_x18 tag\n"
         "mov x10, sp\n"                                                        \
         /* switch to old stack */                                              \
         "mov sp, x9\n"                                                         \
-        /* calculate location to save pointer to newly allocated stack */      \
+        /* calculate location to save addr of newly allocated stack */         \
         "mrs x12, tpidr_el0\n"                                                 \
         "adrp x11, :gottprel:ia2_stackptr_" #i "\n"                            \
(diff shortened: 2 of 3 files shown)


runtime/libia2/init.c
```diff
@@ -26,6 +26,10 @@ char *allocate_stack(int i) {
       exit(-1);
     }
   }
+#ifdef __aarch64__
+  /* Tag the allocated stack pointer so it is accessed with the right pkey */
+  stack = (char *)((uint64_t)stack | (uint64_t)i << 56);
```
ayrtonm, line 31: are we sure `stack`'s upper bits are initially zero or should we explicitly clear them out first?

ayrtonm (approved):
I had one question but otherwise LGTM. Thanks for catching this oversight.

ayrtonm:
> the violations I observed seemed to come from sp-relative accesses, which I thought would not perform MTE checks

This is expected for some instructions using sp. I'll find a link in a sec

ayrtonm:
Can't find a link from ARM at the moment but here's what I thought was a bug in QEMU which pointed out that some sp-relative accesses are checked. https://gitlab.com/qemu-project/qemu/-/issues/514. I think it's basically those that access "sp + runtime offset" since that can't be known to access the right thing statically.

fw-immunant:
CI here is failing for a really interesting reason: passing tagged pointers into the stack as system call arguments returns `EFAULT` (bad address)...

ayrtonm:
> passing tagged pointers into the stack as system call arguments returns EFAULT (bad address)...

to mprotect?

ayrtonm:
oh I'm guessing it probably needs the `prctl` stuff from #334.

ayrtonm:
https://www.kernel.org/doc/html/next/arm64/tagged-address-abi.html#aarch64-tagged-address-abi-exceptions

fw-immunant:
> oh I'm guessing it probably needs the `prctl` stuff from #334.

Ah, yep--that's why it was working when I wrote these commits but not when I pulled them out in to their own PR.

fw-immunant:
I confirmed visually that only the 4 ARM tests we expect to fail are producing MTE violations in CI:
- recursion
- read_config
- simple1
- static_addr_taken

Next we should make these failures act as *test failures* in CI and also actually fix them.

Issue #413, opened 2024-10-03
heap_two_keys and three_keys_minimal 


rinon:
#378 exposed the issue that `heap_two_keys` and `three_keys_minimal` are failing due to not segfaulting as expected. We are disabling these tests until this is investigated to unblock further development.

rinon:
Fixed by #419 and #422

PR #419, opened 2024-10-04
Fix propagation of partition-alloc link options

rinon:
The partition-alloc module exported public linker options, but this was not propagated through the padded version of the library. This change propagates PUBLIC LINK_OPTIONS from partition-alloc to its consumers.


rinon:
Addresses half of #413 (fixes `heap_two_keys`)

fw-immunant (approved):
LGTM.

ayrtonm (approved):

PR #421, opened 2024-10-04
Label /proc/pid/maps entries in permissive mode logs

### LINKED ISSUE #405: Add mmap region descriptions to permissive mode logs
Permissive mode currently prints the state of `/proc/self/maps` when the process is terminating. This technically includes everything but it's hard to tell which memory regions correspond to the stacks and TLS so we should add this information explicitly.

ayrtonm:
* Fixes #405.

This removes unnecessary info from the /proc/pid/maps entries in the permissive mode logs and labels the compartment stacks in the column with file paths. Still a WIP since this only labels the first thread's stacks. Aside from the stacks created by other threads, we could also label compartment heaps, compartment TLS regions (not sure if there's one per thread or just one per DSO?).

### DIFF
--- a/runtime/libia2/ia2.c
+++ b/runtime/libia2/ia2.c
@@ -246,6 +246,8 @@ static int segment_flags_to_access_flags(Elf64_Word flags) {
       ((flags & PF_R) != 0 ? PROT_READ : 0);
 }
 
+uintptr_t tls_addr[16][2] IA2_SHARED_DATA = {0};
+
 int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
   if (!data || !info) {
     printf("Passed invalid args to dl_iterate_phdr callback\n");
@@ -258,6 +260,7 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
     // Continue iterating to check the next object
     return 0;
   }
+  const int pkey = search_args->pkey;
 
   // Protect TLS segment.
   for (size_t i = 0; i < info->dlpi_phnum; i++) {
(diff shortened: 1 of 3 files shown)


runtime/libia2/include/permissive_mode.h
```diff
@@ -214,7 +214,7 @@ void permissive_mode_handler(int sig, siginfo_t *info, void *ctxt) {
 static bool exiting IA2_SHARED_DATA = false;
 
 // Process-specific name for log file
-static char log_name[256] IA2_SHARED_DATA = {0};
```
ayrtonm, line 217: I don't think this change is necessary, right?

kkysen: I'm not sure.  Do you remember why you made it?

runtime/libia2/init.c
```diff
@@ -12,6 +13,8 @@
 extern __thread void *ia2_stackptr_0[PAGE_SIZE / sizeof(void *)]
     __attribute__((aligned(4096)));
 
+uintptr_t ia2_stacks[16] IA2_SHARED_DATA = {0};
+
```
fw-immunant, line 17: Because this is `IA2_SHARED_DATA`, this is definitely not secure. That's fine if it's just for logging in permissive mode, but we should have a comment to that effect so nothing security-critical starts relying on it.

runtime/libia2/ia2.c
```diff
@@ -246,6 +246,8 @@ static int segment_flags_to_access_flags(Elf64_Word flags) {
       ((flags & PF_R) != 0 ? PROT_READ : 0);
 }
 
+uintptr_t tls_addr[16][2] IA2_SHARED_DATA = {0};
```
ayrtonm, line 249: Can we use add `const int`s for these integer literals? I understand the first is number of pkeys, but I don't know what the other one is.

kkysen: Yeah.  I was gonna ask if there's a macro for the max number of compartments.  The 2 is just because sometimes the TLS regions are on the side (so there are two) and sometimes it's just whole.

ayrtonm: I don't understand why TLS is split if we don't support shared TLS yet, but I'll take a look. Either way shouldn't it show up as a single segment in proc maps?

ayrtonm: Only the first compartment should have a split TLS region so I think this can be a 16 entry array plus 2 more variables for the split TLS.

runtime/libia2/init.c
kkysen: Moved to #554.

runtime/libia2/ia2.c
fw-immunant: Yeah, (reiterating from Slack for posterity) the split TLS region for compartment 0 is because there is a page of unprotected data for `ia2_stackptr_0` plus padding, as we don't have a general implementation of shared TLS yet but `ia2_stackptr_0` is special-cased for now as it must be stored in TLS and unprotected.

kkysen:
Moved to
* #555
* #556
* #557
* #560

PR #422, opened 2024-10-04
Fix ELF segment end computation

rinon:
ELF segment end computation was using the page-aligned start, not the virtual address start (which may not be page-aligned). This was resulting in not protecting the BSS section.

### DIFF
--- a/runtime/libia2/ia2.c
+++ b/runtime/libia2/ia2.c
@@ -391,7 +391,7 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
 
     struct AddressRange *relro_range = &shared_ranges[shared_range_count++];
     relro_range->start = (info->dlpi_addr + phdr.p_vaddr) & ~0xFFFUL;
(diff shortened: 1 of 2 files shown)


ayrtonm (approved):

ayrtonm:
wait do we need an analogous fix for `protect_tls_pages`?

ayrtonm:
I don't remember off-hand how TLS is laid out in memory so I'm not sure.

rinon:
TLS looks ok to me, but RELRO needed the same fix. Not that it matters there because RELRO start is page-aligned already in the ELF, but this just ensures we do it right even if it's not.

rinon:
> maybe point out the pkey_mprotected parts in permissive mode logs (https://github.com/immunant/IA2-Phase2/pull/421).

I don't really understand, what do you want documented?

ayrtonm:
Oh I meant mark the segments or parts of a segment that are protected in the /proc/pid/maps that we write to the end of each permissive mode log. But I can do it.

rinon:
Oops, wrong branch. I'll just push these two commits to master as well.

PR #423, opened 2024-10-04
Add verbose debug logging to libia2

rinon:
To print verbose logs of all mprotects and ELF segment computation, build with `-DLIBIA2_DEBUG_LOG=ON`.


runtime/libia2/include/ia2_internal.h
```diff
@@ -300,9 +300,16 @@ works as a reasonable signpost no-op. */
 #if defined(__aarch64__)
 int ia2_mprotect_with_tag(void *addr, size_t len, int prot, int tag);
 #elif defined(__x86_64__)
+#if LIBIA2_DEBUG_LOG
+static int ia2_mprotect_with_tag(void *addr, size_t len, int prot, int tag) {
+  printf("ia2_mprotect_with_tag(addr=%p, len=%zu, prot=%d, tag=%d)\n", addr, len, prot, tag);
+  return pkey_mprotect(addr, len, prot, tag);
+}
```
ayrtonm, line 307: Maybe we should merge this with the definition for arm in ia2.c?

ayrtonm:
nit: do we want to namespace our runtime's stuff with `IA2_` or `LIBIA2_`? I think we generally do the former (or usually nothing at all) but we should probably try to be consistent.

runtime/libia2/include/ia2_internal.h
rinon: We could, I'm not sure we should have a single function though. I'd rather define separate functions under the conditionals. I'm wondering if it would make sense to make this a static function in the header so we can inline it everywhere?

rinon:
re: naming, I was modeling this off of the existing `LIBIA2_DEBUG`. I don't have a strong preference between the two, but if we change this to IA2_ we should change that too.

ayrtonm:
> I was modeling this off of the existing LIBIA2_DEBUG

yeah that's one of the few places we use `LIBIA2_` so I we should make both `IA2_`.

runtime/libia2/include/ia2_internal.h
ayrtonm: > I'd rather define separate functions under the conditionals.

that part's fine with me

>  I'm wondering if it would make sense to make this a static function in the header so we can inline it everywhere?

Static functions in headers should minimize compartment transitions but they make debugging more complicated since you need to be aware of which copy of the function you're calling and which DSO it's located in. So I'm leaning towards not doing this.

ayrtonm (approved):
LGTM

PR #424, opened 2024-10-05
Integrate various ARM branches

fw-immunant:
Some of the commits here from other branches are stale and others are modified compared to their upstream branch, but this successfully runs a number of tests locally with MTE enforced. PR for CI only for now.


runtime/libia2/main.c
```diff
@@ -49,13 +49,26 @@ __asm__(
     // Save old stack pointer in main_sp
     "adrp x9, main_sp\n"
     "add x9, x9, #:lo12:main_sp\n"
+    //ask for tag in x0?
+    "eor x0, x0, x0\n"
+    "ldg x0, [x9]\n"
```
rinon, line 54: Can this just be `ldg x9, [x9]`?

PR #430, opened 2024-10-07
runtime: fix `-Wsign-compare` warnings

kkysen:
Projects being rewritten and compiled may have warnings/errors like this turned on, so it's simpler to just fix them.

Note that this doesn't set `-Werror=sign-compare` in `cmake` because that'll apply to all of the tests, too.  We just want to make sure that included headers work under different possible warning/error configurations a project might set.

### DIFF
--- a/runtime/libia2/include/ia2_compartment_init.inc
+++ b/runtime/libia2/include/ia2_compartment_init.inc
@@ -146,7 +146,7 @@ COMPARTMENT_IDENT(ia2_compartment_destructor)(void) {
 
   void (**fini_array)(void) = (void (**)(void))(((uint8_t *)&__ehdr_start) +
                                                 finalizers.fini_array_offset);
-  for (int i = 0; i < (finalizers.fini_array_size / sizeof(void *)); i++) {
+  for (size_t i = 0; i < (finalizers.fini_array_size / sizeof(void *)); i++) {
     fini_array[i]();
   }
   if (finalizers.fini_offset) {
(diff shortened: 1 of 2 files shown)


kkysen:
> Maybe we should build with `-Wsign-compare` as well? I mean we already build with ubsan to catch potential issues.

Sure, we can do that, too.  I think the `-Wsign-compare` came from something like `-Wall` so we should probably just do that.

ayrtonm:
`-Wall` would be nice but it's not a priority so let's stick to just `-Wsign-compare` for now. Also when we update the docs we should keep track of which flags are required and which are just nice-to-haves.

ayrtonm (approved):
LGTM after adding `-Wsign-compare` [here](https://github.com/immunant/IA2-Phase2/blob/main/cmake/ia2.cmake#L58).

kkysen:
@ayrtonm, hmm, now I'm hitting other `sign-compare`s in the other examples/tests.  Is there a way I can only enable `-Wsign-compare` in `cmake` some of the time, because this is something that we want to work without errors no matter what flags the downstream user is setting.  Maybe it's easier to just not set `-Werror=sign-compare` for us?

ayrtonm:
Actually the last place I linked wouldn't apply it to the original places you fixed, only the sources in runtime/libia2 so I'm fine leaving the flag out for now.

kkysen:
Merging now since @ayrtonm already approved this version and we decided not to set `-Werror=sign-compare` anymore.

PR #431, opened 2024-10-07
runtime: use `__attribute__((__noreturn__))` instead of `_Noreturn` to avoid errors

kkysen:
`_Noreturn` doesn't exist before C11, and is deprecated in C23 (in favor or `[[noreturn]]`), so it's simpler just to use `__attribute__((__noreturn__))` since we don't know what C version the headers will be compiled with.

### DIFF
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -306,7 +306,7 @@ int ia2_mprotect_with_tag(void *addr, size_t len, int prot, int tag);
 char *allocate_stack(int i);
 void verify_tls_padding(void);
 void ensure_pkeys_allocated(int *n_to_alloc);
-_Noreturn void ia2_reinit_stack_err(int i);
+__attribute__((__noreturn__)) void ia2_reinit_stack_err(int i);
 
 #define _IA2_INIT_RUNTIME(n)                                                   \
   int ia2_n_pkeys_to_alloc = n;                                                \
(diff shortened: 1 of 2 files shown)


ayrtonm:
I think we should replace our uses of `_Noreturn` with `__attribute__((__noreturn__))` instead.

kkysen:
> I think we should replace our uses of `_Noreturn` with `__attribute__((__noreturn__))` instead.

In all cases, or just pre C11?

ayrtonm:
All cases. Glibc [sometimes defines](https://github.com/immunant/IA2-Phase2/blob/main/external/glibc_sysroot/include/sys/cdefs.h#L692) it as `#define _Noreturn` but I think that's just for old versions of gcc (and unrecognized attributes should be ignored anyway).

kkysen:
> All cases. Glibc [sometimes defines](https://github.com/immunant/IA2-Phase2/blob/main/external/glibc_sysroot/include/sys/cdefs.h#L692) it as `#define _Noreturn` but I think that's just for old versions of gcc and unrecognized attributes should be ignored anyway. (Also I linked to our repo's glibc sysroot out of convenience but we don't use it on x86 in case that wasn't clear.)

Should I do something like `#define _Noreturn __attribute__((__noreturn__))` or just use `__attribute__((__noreturn__))` inline in each case (just more verbose I guess)?

ayrtonm:
I think `__attribute__((__noreturn__))` is preferable.

ayrtonm:
I think `__attribute__((__noreturn__))` is preferable since it has the lowest chance of breaking with different C libraries/compilers/unexpected macro definitions.

ayrtonm (approved):
LGTM

PR #434, opened 2024-10-07
Add stack traces to permissive mode logs

ayrtonm:
This turned out to be easier to implement than labeling the /proc/pid/maps entries w/o changes across the runtime codebase so I pushed it separately since these changes are contained to permissive_mode.h. @randomPoison I tested manually by tweaking our tests but permissive mode doesn't interact well with criterion so it was pretty limited testing. Could you run this on the zlib demo and post the result?

### DIFF
--- a/runtime/libia2/include/permissive_mode.h
+++ b/runtime/libia2/include/permissive_mode.h
@@ -103,7 +103,6 @@ typedef struct mpk_err {
   uint64_t sp;
   uint64_t fp;
   uint32_t pkru;
-  uint64_t local_addr;
 } mpk_err;
 
 struct queue {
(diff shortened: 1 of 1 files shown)


ayrtonm:
Also you may need to build with `-fno-omit-frame-pointer`.

ayrtonm:
huh CI broke because the address of the return address that I'm deriving from the frame pointer is unaligned. We could "fix" this by replacing the new `*(uint64_t *)x`s with`memcpy`s but that's probably just covering up the underlying issue. `ret` on x86 doesn't require alignment so this should work when disabling ubsan, but I think the SysV ABI says the return address address has to be aligned so something else is wrong here.

ayrtonm:
This only seems to affect things below `main` on the call stack so it seems like crt stuff deciding to not follow SysV ABI as strictly at times. In this case `memcpy` would be fine, but I'm not sure whether I should do that or add the additional stopping condition `(fp & 0xF) != 0`.

randomPoison:
It seems like permissive mode isn't working when I build against your branch. I'm seeing segfaults on the cross-compartment memory accesses, and nothing is being printed to the log. I tested against main and permissive mode works as expected there, so the issue shouldn't be with my setup, but it's possible I'm doing something wrong still.

ayrtonm:
Just realized I need to grab return addresses while grabbing the frame pointers otherwise the return addresses may be overwritten.

PR #436, opened 2024-10-08
Pass indirect call target in first register

rinon:
Rather than pass the indirect call target in a global static, this change passes the target in the first parameter register to the wrapper. The wrapper then shifts the remaining parameters so that the callee receives the correct parameters.

### DIFF
--- a/runtime/libia2/include/ia2.h
+++ b/runtime/libia2/include/ia2.h
@@ -45,7 +45,7 @@
 #define IA2_ADDR(opaque) (void *)opaque
 #define IA2_AS_PTR(opaque) opaque
 #define IA2_FN(func) func
-#define IA2_CALL(opaque, id) opaque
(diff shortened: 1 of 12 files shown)


ayrtonm:
> Rather than pass the indirect call target in a global static, this change passes the target in the first parameter register to the wrapper

I agree this is a better approach than a thread-local in memory.

ayrtonm (approved):
LGTM

ayrtonm:
I think at some point we'll want to generate scrub registers in GenCallAsm.cpp to avoid the redundant push/pops and avoid spilling the function pointer to the stack in indirect calls, but this is already better than what we currently do since it's only on a protected stack.

fw-immunant (approved):
Looks good. I wish our handling of stack alignment were less ad-hoc.

PR #440, opened 2024-10-10
Secure PKRU during thread startup

### LINKED ISSUE #257: Thread startup saves PKRU in spillable register across function calls
See `ALLOCATE_COMPARTMENT_STACK_AND_SETUP_TLS`. For each compartment, we save PKRU in `r12d`, change PKRU to the pkey for compartment, call `init_tls_N` for that compartment, then restore the PKRU from `r12d`.

The latter step can end up reading the wrong value if r12 is spilled to the stack and another thread in the same compartment writes to that memory before the register is unspilled. This can change what PKRU value the thread ends up running its main function with. The PKRU is then used to decide which compartment stack gets used for the thread, so a PKRU for the wrong compartment can/will not be noticed. A PKRU value that has pkeys for multiple compartments will get `NULL` back from `ia2_stackptr_for_pkru`, which will result in a null dereference and DoS, but one for another compartment will not.

rinon:
Rather than saving and restoring the PKRU from a register when allocating and securing thread stacks, we branch on the initial value to one of a set of duplicated initialization routines, each of which hard-codes the PKRU value to restore after initialization.

Fixes #257

### DIFF
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -249,6 +249,7 @@ asm(".macro movz_shifted_tag_x18 tag\n"
         "mov $" #i ",%%rdi\n"                                                  \
         "call ia2_reinit_stack_err\n"                                          \
         "ud2\n"                                                                \
+        /* %= produces a unique (to this inline asm block) label value */      \
         ".Lfresh_init%=" #i ":\n"                                              \
         /* store the stack addr in the stack pointer */                        \
         "mov %%r10,%%fs:(%%r11)\n"                                             \
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -47,22 +47,22 @@ struct dl_phdr_info;
 #define _IA2_CALL(opaque, id, pkey) __IA2_CALL(opaque, id, pkey)
 
 /* clang-format off */
-#define REPEATA0(fn, basefn, ...) basefn(0, ##__VA_ARGS__)
-#define REPEATA1(fn, basefn, ...) REPEATA0(fn, basefn, ##__VA_ARGS__) fn(1, ##__VA_ARGS__)
-#define REPEATA2(fn, basefn, ...) REPEATA1(fn, basefn, ##__VA_ARGS__) fn(2, ##__VA_ARGS__)
-#define REPEATA3(fn, basefn, ...) REPEATA2(fn, basefn, ##__VA_ARGS__) fn(3, ##__VA_ARGS__)
-#define REPEATA4(fn, basefn, ...) REPEATA3(fn, basefn, ##__VA_ARGS__) fn(4, ##__VA_ARGS__)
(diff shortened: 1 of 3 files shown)


runtime/libia2/include/ia2_internal.h
```diff
@@ -210,19 +214,13 @@ asm(".macro movz_shifted_tag_x18 tag\n"
         "mov ia2_stackptr_" #i "@GOTTPOFF(%%rip),%%r11\n"                      \
         /* check that stack pointer holds NULL */                              \
         "cmpq $0x0,%%fs:(%%r11)\n"                                             \
-        "je .Lfresh_init" #i "\n"                                              \
+        "je .Lfresh_init%=" #i "\n"                                            \
```
fw-immunant, line 217: What does this `%=` syntax do?

rinon: Creates a fresh unique label (this asm block is now duplicated and labels are a global namespace).

runtime/libia2/include/ia2_internal.h
```diff
@@ -40,29 +47,46 @@ struct dl_phdr_info;
...
+#define REPEATB14(fn, basefn, ...) fn(14, ##__VA_ARGS__) REPEATB13(fn, basefn, ##__VA_ARGS__)
+#define REPEATB15(fn, basefn, ...) fn(15, ##__VA_ARGS__) REPEATB14(fn, basefn, ##__VA_ARGS__)
 /* clang-format on */
 
 /* Macro to repeatedly apply a function or function-like macro `fn` a given
 number of times, passing the index to each invocation. The passed index `n` is
 first, followed by n-1 and so on. For the base case of 0, `basefn` is applied
 instead of `fn`. */
-#define REPEATB(n, fn, basefn) REPEATB##n(fn, basefn)
+#define REPEATA(n, fn, basefn, ...) REPEATA##n(fn, basefn, ##__VA_ARGS__)
```
fw-immunant, line 88: Can you document this? As a historical note, the `REPEATB` name came from "**REPEAT** this macro application _n_ times, with a different **B**ase case". What's the intuition for the `REPEATA` name and how is it used? Is it just expanding in the opposite order to give an ascending sequence?

runtime/libia2/include/ia2_internal.h
fw-immunant: This might be worth commenting because I couldn't find any docs for this syntax in a cursory google search.

fw-immunant: Ohh, this is an inline ASM thing rather than an assembler feature: https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Special-format-strings

fw-immunant (approved):
Semantically LGTM, a couple non-functional nits.

rinon:
@fw-immunant how about `REPEATB_REV()`?

fw-immunant:
Thanks! Feel free to merge.

Issue #428, opened 2024-10-07
rewriter: `static` fns with the same name result in duplicate `extern IA2_fn_ptr_* __ia2_*` declarations


kkysen:
When there are multiple `static` functions with the same name, they result in duplicate `extern IA2_fn_ptr_* __ia2_*` declarations in `callgate_wrapper.h`.  I'm not certain this an error, since other errors are in the way first, but it seems like an error.

For example, in `dav1d`, this happens with `init_internal`, which is defined in `lib.c` and `scan.c` and used when calling `pthread_once`.

ayrtonm:
We took the issue of call gates for static functions with the same name into account in a few places but ultimately decided to declare them all in the output header which prevents them. This was just to minimize the number of build artifacts and simplify the build process, but I think there's multiple ways we might be able to handle this.

>  I'm not certain this an error, since other errors are in the way first, but it seems like an error. For example, in dav1d, this happens with init_internal, which is defined in lib.c and scan.c and used when calling pthread_once.

If both these files define a `static` `init_internal` function and take their address (must happen in both .c's) then it's an issue we need to address.

ayrtonm:
`-Wl,-zunique-symbol` might come in handy here if call gate names cannot be local.

ayrtonm:
`-Wl,-zunique-symbol` might come in handy here if call gate names cannot be local.

ayrtonm:
I think the easiest thing to handle the most common cases of this would be to assert that duplicate static functions have the same function signature and generate only one declaration in the header.

kkysen:
> I think the easiest thing to handle the most common cases of this would be to assert that duplicate static functions have the same function signature and generate only one declaration in the header.

Yeah that would improve things.  Move the error to the rewriter, and if the functions are the same, then just remove duplicates.

ayrtonm:
My tests for this cover static functions defined in headers, duplicate definitions of static functions in source files and different definitions of static functions with the same name in source files. @kkysen any other cases I should cover?

ayrtonm:
oh I just realized the duplicate declarations don't actually cause an error and if the function signatures don't match it'll trigger a compile error. Since the generated header isn't considered developer-facing this is much lower priority than I thought. I'll push the tests I wrote and the change I made for #414 (it's needed for static functions defined in headers) but I don't think it's worth spending time on moving the errors to the rewriter for now.

kkysen:
[...]

PR #441, opened 2024-10-11
rewriter: Redeclare static functions that have their address taken with used

### LINKED ISSUE #414: `static` address-taken functions generate linker errors
If we have a function `foo` that is used as a function pointer but is also marked `static`, we get linker errors because the call gates in `libcallgates.so` try to call `foo`, but `foo` can't be seen outside of the library it's in. The work around for now is to remove the `static` modifier so that the linker can find the symbol. I'm not sure if this can be addressed by improving the rewriter, or if this is just an IA2 limitation that we should document.

ayrtonm:
When we make call gates for static functions that have their address taken, the static function is only referenced from the asm for the call gate. Compiling with optimizations may get rid of the static functions unless they are marked with `__attribute__((__used__))`. The rewriter previously prepended this attribute in this case to avoid that. This commit removes the prepended attribute and instead redeclares these functions with that attribute. Closes #414 and fixes a test added for #428.

### DIFF
--- a/runtime/libia2/include/ia2.h
+++ b/runtime/libia2/include/ia2.h
@@ -49,7 +49,7 @@
 #define IA2_CAST(func, ty) (ty) (void *) func
 #else
 #define IA2_DEFINE_WRAPPER(func) \
-    typeof(func) func __attribute__((__used__)); \
+    __attribute__((__used__)) typeof(func) func; \
     IA2_DEFINE_WRAPPER_##func
 #define IA2_SIGHANDLER(func) ia2_sighandler_##func
 /// Create a wrapped signal handler for `sa_sigaction`
--- a/runtime/libia2/include/ia2.h
+++ b/runtime/libia2/include/ia2.h
@@ -48,7 +48,9 @@
 #define IA2_CALL(opaque, id) opaque
 #define IA2_CAST(func, ty) (ty) (void *) func
 #else
-#define IA2_DEFINE_WRAPPER(func) IA2_DEFINE_WRAPPER_##func
+#define IA2_DEFINE_WRAPPER(func) \
+    typeof(func) func __attribute__((__used__)); \
+    IA2_DEFINE_WRAPPER_##func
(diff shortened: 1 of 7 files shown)


ayrtonm:
Still need to add LIT annotations and a reference for the expected stdout to the new test.

ayrtonm:
guess clang is more strict than gcc here. I could add a __used__ pointer referencing the function in `IA2_DEFINE_WRAPPER` but I'll see if there's a less silly solution.

ayrtonm:
I went with the `__used__` pointer solution and the build looks good. I'm going to hold off on debugging this until I replace criterion with our test runner.

ayrtonm:
TODO: push the tests with the case mentioned in #414

runtime/libia2/include/ia2.h
```diff
@@ -48,7 +48,9 @@
 #define IA2_CALL(opaque, id) opaque
 #define IA2_CAST(func, ty) (ty) (void *) func
 #else
-#define IA2_DEFINE_WRAPPER(func) IA2_DEFINE_WRAPPER_##func
+#define IA2_DEFINE_WRAPPER(func) \
+    __attribute__((__used__)) static void *keep_##func = (void *)func; \
```
rinon, line 52: I don't love adding a function pointer to data just to mark a function as used. We don't have any other way?

ayrtonm: I haven't put much thought into this. There might be a way to get `__attribute__((used)) typeof(func) func;` before the function definition, but I'd need to think about how that would work with the types `func`'s signature depends on.

kkysen:
> Compiling with optimizations may get rid of the static functions unless they are marked with `__attribute__((__used__))`.

Would marking them with `__attribute__((visibility("default"))`, which we need for #443, supersede marking them as used?

ayrtonm:
@rinon will handle this by checking that the macro expansion starts at the beginning of the function definition location instead. There's a test in this PR you might want to reuse, but it mainly covers the cases in #428 which is super relevant here.

PR #445, opened 2024-10-14
(WIP) rewriter: add `__attribute__((visibility("default"))` to `IA2_FN` address taken functions if they don't already default visibility

### LINKED ISSUE #443: rewriter: support compiling with `-fvisibility=hidden`
* Part of #435.

For the runtime support for `-fvisibility=hidden`, we also need to mark all functions with call gates as `__attribute__((visibility("default")))` because their call gate wrappers are defined in the call gate DSO.

kkysen:
* Fixes #443.

I tried copying the similar approach used to insert the `__attribute__((used))`s, and while its finding the right functions to add the `__attribute__((visibility("default")))`s to, the replacements are going in totally wrong places places, and I'm not sure what I'm doing wrong or quite how the replacements API is supposed to work.  @ayrtonm or anyone else, do you have any suggestions?

### DIFF
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -686,6 +686,22 @@ public:
       }
     }
 
+    if (fn_decl->getVisibility() != clang::Visibility::DefaultVisibility) {
+      auto decl_start = fn_decl->getBeginLoc();
+      if (!decl_start.isFileID()) {
+        llvm::errs() << "Error: non-file loc for function " << fn_name << '\n';
+      } else {
+        llvm::errs() << "trying to add __attribute__((visibility(\"default\"))) for " << fn_name << '\n';
+        Replacement old_used_attr(sm, decl_start, 0,
+                                  llvm::StringRef("__attribute__((visibility(\"default\"))) "));
+        Replacement used_attr = replace_new_file(filename, old_used_attr);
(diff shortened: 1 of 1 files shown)


ayrtonm:
Do we actually need fvisibility on the address taken functions? I think it's only the wrappers that need it.

ayrtonm:
Oh it's needed for non-static address taken functions. It might be better to define the call gates in the same .c like we do for static functions that way we don't have to prepend attributes to function definitions.

I'm guessing the issues you're running into probably stem from expansion/spelling loc differences. Doing these kinds of rewrites gets trickier when functions are defined in headers/expanded from macros so I'd like to avoid them if possible.

ayrtonm:
>  It might be better to define the call gates in the same .c like we do for static functions

I just realized that these may be referenced from multiple .c files so we need to arbitrarily pick a .c per compartment for non-static address-taken function call gates or generate a .c per compartment and link that in. The first option is kinda ugly but I'm leaning towards that for expediency. @rinon thoughts?

ayrtonm:
For context I mostly want to avoid prepending attributes to functions with the rewriter to avoid cases like #414 where a function's definition has `static` or some attribute expanded from a macro and we can't reliably tell if the macro expansion includes anything else.

ayrtonm:
Superseded by #451.

PR #446, opened 2024-10-14
cmake: Disable PartitionAlloc on ARM

ayrtonm:
PartitionAlloc needs to be built with -ffixed-x18 on ARM to avoid clobbering x18. However linking it in also pulls in libstdc++ which can also clobber x18 so this commit just disables PartitionAlloc on ARM for now. It shouldn't be too hard to build LLVM's libc++ with -ffixed-x18 since that's what Android uses so we should probably just do that when we want to re-enable PA.

### DIFF
--- a/cmake/ia2.cmake
+++ b/cmake/ia2.cmake
@@ -69,7 +69,12 @@ function(add_ia2_compartment NAME TYPE)
     target_link_options(${NAME} PRIVATE ${UBSAN_FLAG})
   endif()
 
-  target_link_libraries(${NAME} PRIVATE dl libia2 partition-alloc)
+  if (LIBIA2_AARCH64)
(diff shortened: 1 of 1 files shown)


fw-immunant:
Is libstdc++ being linked causing problems even in tests that don't do heap allocation? If not, I would rather we kept PartitionAlloc in the build and just ignored test failures of heap tests until we have our own build of libc++. We have ARM-specific changes to PartitionAlloc that I would prefer to keep building in CI.

ayrtonm:
No I haven't hit runtime bugs due to this, it was to preempt potential bugs from calling into it. Though libstdc++.so also defines some unmangled symbols so I don't think we can say with certainty that not allocating will not call into libstdc++.so. If you want to keep build PA we could add `ninja partition-alloc` to CI instead.

PR #447, opened 2024-10-15
Improve static address taken function rewriting

rinon:
Handles adding `used` attribute to functions defined in headers or that begin with a macro expansion.


rinon:
Weird, on my system the main binary identical name function is called correctly, but in CI it isn't.

rinon:
I'm disabling the identically named function part of the test. That should be fixed separately, as it is unrelated to this change.

ayrtonm (approved):
LGTM

PR #448, opened 2024-10-15
Replace criterion with our test runner on x86

### LINKED ISSUE #439: Replace use of prebuilt criterion with our test runner
When using either criterion or our fake criterion lib we turn each `Test` macro into a function that has its address taken so the test runner can call it. When we rewrite the test sources each `Test` function has a wrapper generated (with real criterion it seems like it's a static fn so the wrapper is defined at the bottom of the test .c). Since the callsite for the function pointers is in libcriterion.so there's no corresponding `IA2_CALL` so we should make sure that the pointer points to the original function, not the wrapper. I think it should be the case since the function pointer is assigned inside of the `Test` macro but we should double check.

This potential bug also applies to our fake criterion but in that case the test functions are not static so the wrappers aren't defined in the rewritten test sources. Also fake criterion is just a static lib that the tests link against so I don't think it gets rewritten (and we should probably keep it that way). @fw-immunant have you thought about these issues?

ayrtonm:
This PR also moves the test_fault_handler.h functionality into the test runner. Closes #439


ayrtonm:
CI is failing because the (existing) `setbuf(stdout, NULL)` is forcing compartment pkey != 1 to flush stdout which is actually located in the main binary's .bss. Nothing seemed out of the ordinary in glibc's definition for the streams so I'm not sure which flags/what aspect of symbol versioning is causing it to be in .bss. I'll just get rid of the `setbuf` which removes the symbol though that's not solving the underlying problem.

ayrtonm:
We should probably also use `sigaction` rather than `signal` to install the handler.

misc/test_runner/include/ia2_test_runner.h
```diff
@@ -0,0 +1,69 @@
...
+  } while (0)
+
+/*
+ * This header defines a test framework for detecting MPK violations using
+ * signal handlers. This file must be included exactly once from a source file
+ * in the main binary with IA2_DEFINE_TEST_HANDLER defined by the preprocessor.
+ * This will define the functions and variables used by the test handler, ensure
+ * it is initialized before main and provide access to the LOG and
+ * CHECK_VIOLATION macros. Other files which need CHECK_VIOLATION or LOG may
+ * include the header without defining IA2_DEFINE_TEST_HANDLER. Using
```
rinon, line 51: Is this still true? Do we need to define `IA2_DEFINE_TEST_HANDLER`?

tools/rewriter/SourceRewriter.cpp
```diff
@@ -614,6 +614,11 @@ class FnPtrExpr : public RefactoringCallback {
     auto *fn_ptr_expr = result.Nodes.getNodeAs<clang::DeclRefExpr>("fnPtrExpr");
     assert(fn_ptr_expr != nullptr);
 
+    auto annotation = fn_ptr_expr->getDecl()->getAttr<clang::AnnotateAttr>();
+    if (annotation && annotation->getAnnotation() == SKIP_WRAP_ATTR) {
+        return;
+    }
```
rinon, line 620: Is this needed for this change? Just wanted to make sure it was related and not accidental.

misc/test_runner/include/ia2_test_runner.h
ayrtonm: oh right, I forgot to update the comments

tools/rewriter/SourceRewriter.cpp
ayrtonm: yeah, but it's not obvious why so I should've update the documentation in ia2.h. The attribute is added to the function declaration in the `Test` macro and this branch makes it so that expressions that refer to that dec (e.g. the RHS in `.test = fake_criterion_##suite##_##name,`) are not rewritten. In the case of `Test` it wasn't being rewritten since it's expanded from a macro, but the rewriter was printing out warnings about needing to manually rewrite that spot.

ayrtonm: Ideally I would've like to have `IA2_{BEGIN,END}_NO_WRAP` apply to only things between those macros but I don't think we can annotate expressions so I don't see an obvious way to do this.

misc/test_runner/include/ia2_test_runner.h
ayrtonm: updated comments

tools/rewriter/SourceRewriter.cpp
ayrtonm: Actually this pass already called `ignore_function` so I'll remove it since it wasn't necessary. I updated the comment on the `Test` macro definition to explain why the NO_WRAP is necessary there but I'll leave updating the `NO_WRAP` documentation to a separate change after checking exactly what it applies to (it doesn't apply to function pointers between the macros, but it does apply to pointers that reference functions defined between the macros).

PR #453, opened 2024-10-22
Print backtrace on failed ASSERT_PKRUs

### LINKED ISSUE #452: Make `IA2_DEBUG` more useful and on by default
Given that the initial and target compartments of every call gate are known statically, we don't need to check the PKRU in properly compartmentalized programs. For cases that want the extra security of runtime PKRU checks in the callgates we had `-DIA2_DEBUG=1` (disabled by default) and the call gates conceptually look like this
```
#define CALLER_PKRU N

int call_gate_for_foo(int x) {
#if defined(IA2_DEBUG)
    assert(current_pkru == CALLER_PKRU);
#endif
    switch_to_compartment_with_foo();
    int result = foo(x);
    switch_to_caller_compartment();
    return result;
}
```

However the exact details of control-flow are usually not be immediately obvious after inserting call gates. Often while debugging compartmentalized code it's easy to end up in a state with a partially-compartmentalized program and `IA2_DEBUG` can be helpful for debugging this kind of issue. Currently the PKRU assertion just does a `rdpkru` and `ud2` if the result doesn't match the expected value. That means the process ends with a `SIGILL` at that point.

This is better than letting it continue and ending with a `SIGSEGV` later on because it's easier to see that the problem is control-flow not missing memory permissions. It points to a problem earlier on though so I want to make the assertion more useful by having it print a backtrace (this is easy to do manually in gdb but gets a little annoying when processes fork). That would make our tests easier to debug while working on the runtime, but I wanted to see if @kkysen @randomPoison had suggestions about what else might be helpful here. Also I think since this is a debugging feature now it may be useful to add an additional assertion after calling the call gate's target function.

ayrtonm:
This currently prints a trace but at least brief checks on the `read_config` test shows the trace ends at compartment boundaries. Need to fix that, add anything else that might be useful and swap `IA2_DEBUG` for `IA2_RELEASE` with opposite semantics. Closes #452.


ayrtonm:
I think the issue with the shortened backtraces is specifically in `IA2_CALL` callgates, not general cross-compartment calls.

ayrtonm:
Printing the backtrace for the broken test mentioned above gives this
```
0x00007ffff7e0b874 in __wrap_parse_str () from /home/ayrton/ia2/IA2-Phase2/build/tests/read_config/libread_config_call_gates.so
(gdb) bt
#0  0x00007ffff7e0b874 in __wrap_parse_str () from /home/ayrton/ia2/IA2-Phase2/build/tests/read_config/libread_config_call_gates.so
#1  0x00007ffff7e1459c in __ia2_parse_str () from /home/ayrton/ia2/IA2-Phase2/build/tests/read_config/libread_config_lib.so
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```

Although call gate control flow is wrong here (`__ia2_foo` should call `foo` not `__wrap_foo`) this is exactly the type of case that we want to be able to handle so we should do whatever gdb is doing.

PR #454, opened 2024-10-28
docs: add detailed docs of how `dav1d` was compartmentalized

kkysen:


docs/compartmentalizing_dav1d.md
```diff
@@ -0,0 +1,577 @@
...
+* `'-I' + join_paths(ia2_path, 'runtime/libia2/include')`
+
+Then there are args for debugging:
+
+* `-DIA2_DEBUG=1` (for debug assertions)
+* `-DIA2_DEBUG_LOG=1` (for verbose logging)
+
+Then there are overrides:
+
+First just disabling some warnings/errors that `dav1d` enabled but IA2's runtime doesn't follow
```
ayrtonm, line 151: I feel like we should support building with these flags at some point.

docs/compartmentalizing_dav1d.md
```diff
@@ -0,0 +1,577 @@
...
+#### Global Args
+
+For global compile args, we add these:
+
+* `-include`, `join_paths(dav1d_src_root, 'callgate_wrapper.h')`:
+    Here, this `callgate_wrapper.h` is the header of the call gates.
+    It's named this specifically because we set our `--output-prefix` to this.
+* `-Werror=incompatible-pointer-types`:
+  To detect errors around function pointers and their wrapped call gates types.
+* `-DIA2_PERMISSIVE_MODE=1`:
```
ayrtonm, line 376: Please add a note here saying that this is specific to the compartmentalized dav1d demo. It's probably useful functionality to have in general since you found it useful as a flag, but I want to plan where to put that code (both in what source file and in what binary) before just opening up a PR that implements it.

docs/compartmentalizing_dav1d.md
```diff
@@ -0,0 +1,577 @@
...
+
+## Building IA2
+
+The IA2 tools that we need to compartmentalize and run `dav1d` are:
+
+| Binary Name             | `ninja` Target Name       |
+| ----------------------- | ------------------------- |
+| `ia2-rewriter`          | `rewriter`                |
+| `pad-tls`               | `pad-tls`                 |
+| `libpartition-alloc.so` | `partition-alloc-padding` |
```
ayrtonm, line 18: The target should've been `partition-alloc` so we should add that when updating the docs. Both build the same thing, but the one you used is an artifact of cmake limitations and only intended to be used internally.

docs/compartmentalizing_dav1d.md
kkysen: Okay, though `-Wno-unknown-warning-option` we probably want to keep since I think it's a `gcc` vs `clang` thing.

docs/compartmentalizing_dav1d.md
kkysen: Ahh, I didn't realize.  Let me fix that.

docs/compartmentalizing_dav1d.md
kkysen: Sorry, I'm not sure what you mean exactly?  I'm trying to say here you can add this if you want to enable permissive mode.

ayrtonm: I mean that nothing in this repo checks for `-DIA2_PERMISSIVE_MODE`, only the build system/code in the dav1d repo does.

kkysen: Oh, yes, I know what you mean now.  That I added this for `dav1d` as a way of including `<permissive_mode.h>` or not.  Not that it's a part of `ia2` to begin with.  Will do.

kkysen: Fixed in bd9c866.

docs/compartmentalizing_dav1d.md
kkysen: Fixed in 6cf6549.

ayrtonm (approved):

PR #461, opened 2024-10-29
ARM: build and wire up dependencies in CI

fw-immunant:


.github/workflows/ci.yml
```diff
@@ -133,13 +152,36 @@ jobs:
...
         run: |
           LLVM_DIR=`llvm-config-15 --cmakedir`
           Clang_DIR=`realpath $LLVM_DIR/../clang`
           mkdir build
           pushd build
-          cross_flags="-isystem /usr/aarch64-linux-gnu/include --gcc-toolchain=/usr -march=armv8+memtag"
+          rtlibs_dir="$(pwd)/../llvm-project/build-rtlibs"
+          cross_link_flags="-B${rtlibs_dir}/compiler-rt/lib/linux -L${rtlibs_dir}/lib \
+          --rtlib=compiler-rt --unwindlib=libunwind"
+          cross_flags="-isystem /usr/aarch64-linux-gnu/include --gcc-toolchain=/usr -march=armv8.5-a+memtag -ffixed-x18 --rtlib=compiler-rt --stdlib=libc++ -I${rtlibs_dir}/include/c++/v1"
```
rinon, line 184: Either break both sets of flags or have long lines in both places (I think long lines are fine here).

tests/CMakeLists.txt
```diff
@@ -32,7 +32,11 @@ set_target_properties(check-ia2 PROPERTIES FOLDER "tests")
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-# Set up 'check' target and check-exec on arm
+# Set up 'check' target
+if (LIBIA2_AARCH64)
+    # On aarch64, avoid swallowing QEMU MTE check failures
+    set(CMAKE_CTEST_COMMAND ${CMAKE_CTEST_COMMAND} --verbose)
```
rinon, line 38: Why not do this on x86 too?

rinon (approved):
Please document the cross*flags and linker flags somewhere? I don't think the ci.yml is the place but they need to be documented, especially `-rpath-link` and the rtlib stuff. Doesn't need to be in this PR, just soon.

.github/workflows/ci.yml
```diff
@@ -111,13 +111,32 @@ jobs:
           path: |
             llvm-project/build/bin
             llvm-project/build/lib/clang
-          key: ${{ runner.os }}-clang
-      - name: Check out Clang
-        if: steps.cache-clang.outputs.cache-hit != 'true'
+          key: ${{ runner.os }}-clang-2
```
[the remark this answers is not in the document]
fw-immunant: We don't strictly need it as these changes shouldn't break anything if run against older PRs (and we could just drop the cache from the GH UI), but this was necessary to avoid using the previously-existing cached clang without dropping it from cache while this was in development. There doesn't need to be any synchronization between this file's usage of the key and any other place, so it seems better to bump than not.

.github/workflows/ci.yml
```diff
@@ -111,13 +111,32 @@ jobs:
...
+            build-rtlibs/lib
+            build-rtlibs/include
+          key: ${{ runner.os }}-rtlibs
+        # cloning llvm is required if we need to build rtlibs, but the cached clang would already be in the directory that git wants to create
+      - name: Move cached clang out of the way of git clone
+        if: steps.cache-clang.outputs.cache-hit == 'true' && steps.cache-rtlibs.outputs.cache-hit != 'true'
+        working-directory: ${{ github.workspace }}
+        run: |
+          set -x
+          pwd
```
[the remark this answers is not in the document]
fw-immunant: Nope. Will remove.

.github/workflows/ci.yml
```diff
@@ -133,13 +152,36 @@ jobs:
           ./build.sh
           # free up space used by libs; our desired clang binary statically links them
           rm -rf build/lib/*.a
+        # if clang was cached, move it into the git tree to the same location as it would be built
+      - name: Move cached Clang
+        if: steps.cache-clang.outputs.cache-hit == 'true' && steps.cache-rtlibs.outputs.cache-hit != 'true'
+        working-directory: ${{ github.workspace }}
+        run: |
+          set -x
+          pwd
```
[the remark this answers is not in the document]
fw-immunant: Nope.

.github/workflows/ci.yml
fw-immunant: I'll break them for readability.

tests/CMakeLists.txt
ayrtonm: > Our QEMU fork effectively runs in permissive mode, so this is necessary to distinguish test runs with violations from those without.

Right, we need this for tests that fail due to unintentional violations. I don't remember exactly what currently happens with `CHECK_VIOLATION` on ARM off the top of my head.

ayrtonm:
Shouldn't we be passing [`-one-insn-per-tb`](https://github.com/immunant/qemu/commit/dc1c156a51406d6ca4b1ed534ef826c39f3d0d83) to qemu? QEMU gets invoked by ctest since we set `CROSSCOMPILING_EMULATOR` right?

fw-immunant:
> Shouldn't we be passing [`-one-insn-per-tb`](https://github.com/immunant/qemu/commit/dc1c156a51406d6ca4b1ed534ef826c39f3d0d83) to qemu? QEMU gets invoked by ctest since we set `CROSSCOMPILING_EMULATOR` right?

I believe we unconditionally do this in define_test: https://github.com/immunant/IA2-Phase2/blob/main/cmake/define-test.cmake#L151

.github/workflows/ci.yml
```diff
@@ -133,13 +149,36 @@ jobs:
...
         run: |
           LLVM_DIR=`llvm-config-15 --cmakedir`
           Clang_DIR=`realpath $LLVM_DIR/../clang`
           mkdir build
           pushd build
-          cross_flags="-isystem /usr/aarch64-linux-gnu/include --gcc-toolchain=/usr -march=armv8+memtag"
+          rtlibs_dir="$(pwd)/../llvm-project/build-rtlibs"
+          cross_link_flags="-B${rtlibs_dir}/compiler-rt/lib/linux -L${rtlibs_dir}/lib \
+          --rtlib=compiler-rt --unwindlib=libunwind"
+          cross_flags="-isystem /usr/aarch64-linux-gnu/include --gcc-toolchain=/usr \
```
ayrtonm, line 179: can't these be moved to the aarch64 toolchain file?

PR #468, opened 2024-11-12
libia2: WIP Make signal handler macro x86-only

ayrtonm:
edit: When I opened this PR I initially thought that we wouldn't need a signal handler wrapper for ARM but I think that's wrong and the wrapper we currently have is correct so I'll need to find a different way to make misusing IA2_DEFINE_WRAPPER an error.

### DIFF
--- a/runtime/libia2/include/ia2.h
+++ b/runtime/libia2/include/ia2.h
@@ -49,7 +49,10 @@
 #define IA2_CAST(func, ty) (ty) (void *) func
 #else
 #define IA2_DEFINE_WRAPPER(func) IA2_DEFINE_WRAPPER_##func
+
+#if defined(__x86_64__)
 #define IA2_SIGHANDLER(func) ia2_sighandler_##func
+
 /// Create a wrapped signal handler for `sa_sigaction`
 ///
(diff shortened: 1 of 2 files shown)


runtime/libia2/include/ia2.h
```diff
@@ -59,9 +62,10 @@
 ///
 /// Creates a new function with `ia2_sighandler_` prepended to the given
 /// function name which should be registered with sigaction().
-#define IA2_DEFINE_SIGACTION(function, pkey)                                   \
-  void ia2_sighandler_##function(int, siginfo_t *, void *);                    \
-  _IA2_DEFINE_SIGNAL_HANDLER(function, pkey)
+#define IA2_DEFINE_SIGACTION(function, pkey)                                        \
+  __attribute__((naked)) void ia2_sighandler_##function(int, siginfo_t *, void *) { \
+  _IA2_DEFINE_SIGNAL_HANDLER(function, pkey)                                        \
+  }
```
kkysen, line 68: If it's defined as a normal, albeit naked, function, I think the nested (not closure) function works fine.

PR #470, opened 2024-11-21
rewriter: add hardcoded call gate post condition call

kkysen:
For now, the name of the post condition checking function is `${target_name}_post_condition`, and it's hardcoded to only emit for `dav1d_get_picture`.

I'm not sure how to call the function with arguments, though.  I was imagining the API be that the post condition function have the same arguments as the target function and a `void` return type.

### DIFF
--- a/tools/rewriter/GenCallAsm.cpp
+++ b/tools/rewriter/GenCallAsm.cpp
@@ -406,6 +406,14 @@ static std::string sig_string(const AbiSignature &sig,
   return ss.str();
 }
 
+static void emit_direct_call(AsmWriter &aw, Arch arch, std::string_view fn_name) {
+  if (arch == Arch::X86) {
+    add_asm_line(aw, llvm::formatv("call {0}", fn_name));
+  } else if (arch == Arch::Aarch64) {
+    add_asm_line(aw, llvm::formatv("bl {0}", fn_name));
+  }
+}
+
 static void emit_fn_call(
     const std::optional<std::string> target_name,
     WrapperKind kind,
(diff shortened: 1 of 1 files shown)


kkysen:
@ayrtonm, can I merge this and split my stuff into smaller PRs?  Next PR will do the 6 register args, which I'm working on now.

ayrtonm:
To add some other notes from our talk
- we want to check arguments in post-conditions to validate out parameters
- using the first 6 regs will clobber rdx which is fine if the return value is not 128 bits
- as a next step for proper argument handling we may want to use one of the purely stack-based calling conventions for the check functions

tools/rewriter/GenCallAsm.cpp
```diff
@@ -1026,6 +1039,19 @@ std::string emit_asm_wrapper(AbiSignature &sig,
 
   emit_set_return_pkru(aw, caller_pkey, arch);
 
+  // Call the post-condition function, if one was specified.
+  // The call happens in the caller's compartment.
+  // For now, we hardcode the existence and name of the post-condition function.
+  // The name is `${target_name}_post_condition`,
+  // and we only do this for `dav1d_get_picture`.
+  std::optional<std::string> target_post_condition_name = std::nullopt;
+  if (target_name && *target_name == "dav1d_get_picture") {
```
ayrtonm, line 1048: Hardcoding this is fine, but I think this won't let us run the rewriter on the dav1d demo without a post-condition function. Since we're considering using it for the ARM demo and there will be more rewriter codegen changes like in #471 maybe we should gate the post-condition validation with a feature flag.

It can be either a build-time option (e.g. preprocessor #define) or a runtime option (e.g. extra rewriter flag), whichever is easiest. I'm also good with having it on by default to make integrating your changes into the test infra easier.

ayrtonm (approved):
LGTM after adding an option to enable/disable the post-validation codegen.

tools/rewriter/GenCallAsm.cpp
kkysen: I made it a runtime flag in eb8f17b; runtime seemed easier.  On by default, which seems simpler for now.  It'll hopefully soon get un-hardcoded, so I simplest seemed best here.

ayrtonm: thanks! that approach is good with me

PR #474, opened 2024-12-03
tests: replace `fake_criterion_tests` section with `__attribute__((constructor))`s adding to a linked list, fixing the padding bug

### LINKED ISSUE #473: `fake_criterion_tests` iteration is broken due to padding
When trying to add multiple `Test`s in a single executable, I discovered that only the first test was actually being run.  `fake_criterion` currently works by putting all of the `fake_criterion_test`s into the same section, `fake_criterion_tests`, and then iterating over the `fake_criterion_test`s in this section by using the generated symbols `__start_fake_criterion_tests` and `__stop_fake_criterion_tests`.  However, it seems there is now padding between these test elements, so the `++` on the `const fake_criterion_test*` pointer is incorrectly iterating over them.  The padding is zeroed, and so the function pointer null check fails, and the loop exits early.  I'm not sure how the rules work for multiple objects in a section (like if they're the same or different from a C array).

Another potential solution instead of putting all of the test `struct`s into the same section is generate `__attribute__((constructor))` functions that add the test `struct` to a list.  This is how trusty does it, for example: https://android.googlesource.com/trusty/lk/trusty/+/refs/heads/main/include/shared/lk/trusty_unittest.h#323.  However, we'd have to make sure this doesn't interfere with any `__attribute__((constructor))`s used by IA2.

kkysen:
* Fixes #473.

The previous way of appending `struct fake_criterion_test`s by putting them into a section was buggy because extra padding was added between the elements in the section, so iterating through them from `__{start,stop}_fake_criterion_tests` didn't work correctly.

This new way uses `__attribute__((constructor))` to add each `struct fake_criterion_test` to a singly linked list. A singly linked list is used for simplicity, though it does mean the tests are run in reverse order within a file, but I don't really think the test order should matter, so this should be okay.

I also added the suite and test name to `struct fake_criterion_test` and print them out when running tests, which was helpful in distinguishing which tests were being run.

### DIFF
--- a/misc/test_runner/include/ia2_test_runner.h
+++ b/misc/test_runner/include/ia2_test_runner.h
@@ -17,10 +17,15 @@ typedef void (*ia2_test_fn)(void);
 #endif
 
 struct fake_criterion_test {
+  const char *suite;
+  const char *name;
   ia2_test_fn test;
   ia2_test_fn init;
   int exit_code;
 };
+
+#define _STRINGIFY(a) #a
+#define STRINGIFY(a) _STRINGIFY(a)
 
 /*
  * Placing IA2_{BEGIN,END}_NO_WRAP between the function declaration stops the rewriter from creating a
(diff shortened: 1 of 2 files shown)


misc/test_runner/include/ia2_test_runner.h
```diff
@@ -17,25 +17,40 @@ typedef void (*ia2_test_fn)(void);
...
+#define Test(suite_, name_, ...)                                                           \
+  IA2_BEGIN_NO_WRAP                                                                        \
+  void fake_criterion_##suite_##_##name_(void);                                            \
+  IA2_END_NO_WRAP                                                                          \
+  struct fake_criterion_test fake_criterion_##suite_##_##name_##_##test = {                \
+      .next = NULL,                                                                        \
+      .suite = STRINGIFY(suite_),                                                          \
+      .name = STRINGIFY(name_),                                                            \
+      .test = fake_criterion_##suite_##_##name_,                                           \
+      ##__VA_ARGS__};                                                                      \
```
ayrtonm, line 48: I think these structs should have `IA2_SHARED_DATA` to prevent potential test flakiness.

kkysen: Fixed in [e9367d6..b32c3ce](https://github.com/immunant/IA2-Phase2/compare/e9367d68ce41a2f8af4370183efe3132f71dd46f..b32c3ce63e7df8569847173526285ed916ecc1f6).

ayrtonm (approved):

Issue #409, opened 2024-10-03
Document missing build steps


randomPoison:
In integrating IA2 into zlib I uncovered a number of build steps that aren't currently covered in the docs. We'll need to update the docs to include the following:

* [x] How to run the rewriter, i.e. source changes, compiler args, and rewriter args.
* [x] [`partition-alloc`](https://github.com/immunant/IA2-Phase2/tree/main/runtime/partition-alloc) needs to be included in the final containerized build, and the `--wrap=malloc` linker args need to be added.
* [x] Either `LIBIA2_X86_64` or `LIBIA2_ARM64` defines need to be set for the containerized build.
* [x] Run `pad-tls` on any shared objects that use TLS. We probably also want to document how to determine if a `.so` uses TLS.

kkysen:
> Either `LIBIA2_X86_64` or `LIBIA2_ARM64` defines need to be set for the containerized build.

This is fixed as of #400.

ayrtonm:
> We probably also want to document how to determine if a .so uses TLS.

We were thinking about adding runtime checks for this in #404, so we can also point users to the right `readelf` flags in that error message.

ayrtonm:
> [partition-alloc](https://github.com/immunant/IA2-Phase2/tree/main/runtime/partition-alloc) needs to be included in the final containerized build, and the --wrap=malloc linker args need to be added.

It might be a good idea to runtime checks for this as well.

ayrtonm:
The names of `-DIA2_DEBUG` and `-DIA2_DEBUG_LOG` are also too similar. After addressing #452 the former will probably be `-DIA2_RELEASE` with inverted semantics and we should consider renaming the former to something like `-DIA2_VERBOSE` or `-DIA2_LOG` since that seems more descriptive than "debug".

kkysen:
> The names of `-DIA2_DEBUG` and `-DIA2_DEBUG_LOG` are also too similar. After addressing #452 the former will probably be `-DIA2_RELEASE` with inverted semantics and we should consider renaming the former to something like `-DIA2_VERBOSE` or `-DIA2_LOG` since that seems more descriptive than "debug".

`-DIA2_VERBOSE` sounds good to me.

PR #476, opened 2024-12-09
Update build docs and document pad-tls

randomPoison:
This PR documents additional build flags that were needed to containerize zlib, and adds a section on `pad-tls`.

Part of https://github.com/immunant/IA2-Phase2/issues/409.


docs/usage.md
```diff
@@ -155,8 +155,14 @@ are also required:
...
 -DIA2_ENABLE=1
 -include /path/to/generated_output_header.h
+-I $IA2_PATH/runtime/libia2/include
+-I $IA2_PATH/runtime/partition-alloc/include
 -Werror=incompatible-pointer-types
 -Wl,--wrap=pthread_create
+-Wl,--wrap=calloc
+-Wl,--wrap=malloc
+-Wl,--wrap=realloc
+-Wl,--wrap=free
```
ayrtonm, line 165: `runtime/partition-alloc/CMakeLists.txt` lists a few more functions from when Stephen redid our PA shim. We should add these to the README in that directory as well as here. It's not a small list so it might make sense to make a linker @ file for this.

ayrtonm: Also the DSOs needed these as well

docs/usage.md
```diff
@@ -166,8 +172,21 @@ are also required:
...
+-lcallgates
 ```
 
 Also if the rewriter produces a linker args file for a given compartment (i.e. a
 `.ld` file), you must include `-Wl,@/path/to/generated_linker_args_$PKEY.ld` when
 linking that DSO.
+
+## Using Thread Local Storage
+
+When using TLS in a containerized app, you'll need to run the `pad-tls` tool on
```
ayrtonm, line 188:
```suggestion
When using TLS in a compartmentalized app, you'll need to run the `pad-tls` tool on
```

docs/usage.md
```diff
@@ -166,8 +172,21 @@ are also required:
...
+all shared objects in the app. This includes `libc.so` as libc both initializes
```
ayrtonm, line 189: Maybe mention `ldd` and `lddtree` as ways to list all DSOs a program uses.

randomPoison: Good idea, added.

docs/usage.md
randomPoison: I'm going to split making an @ script into another issue (https://github.com/immunant/IA2-Phase2/issues/477), but I'll add the args in the usage docs for now 👍

docs/usage.md
```diff
@@ -158,11 +158,22 @@ are also required:
 -I $IA2_PATH/runtime/libia2/include
 -I $IA2_PATH/runtime/partition-alloc/include
 -Werror=incompatible-pointer-types
--Wl,--wrap=pthread_create
--Wl,--wrap=calloc
--Wl,--wrap=malloc
--Wl,--wrap=realloc
--Wl,--wrap=free
+-Wl,--wrap,pthread_create
```
ayrtonm, line 161: each `-Wl,--wrap` must be followed by `=` not `,`

randomPoison: Ah noted, that should be fixed now.

ayrtonm (approved):

PR #479, opened 2024-12-10
Canonicalize input paths in source rewriter

randomPoison:
Make `RootDirectory` and `OutputDirectory` canonical paths immediately after reading CLI args.  #432 makes them absolute, but we need them to be fully canonical to reliably compare.

### DIFF
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -72,13 +72,13 @@ struct DirectoryParser : public llvm::cl::parser<std::string> {
       llvm::errs() << "error: directory does not exist: " << dir << "\n";
       return true; // true on error
     }
-    auto ec = llvm::sys::fs::make_absolute(dir);
+    llvm::SmallString<PATH_MAX> real_path;
+    auto ec = llvm::sys::fs::real_path(dir, real_path);
     if (ec) {
       llvm::errs() << ec.message() << '\n';
       return true;
     }
-    llvm::sys::path::remove_dots(dir, true);
-    Value = std::string(dir);
+    Value = std::string(real_path);
     return false;
   }
 };
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -77,6 +77,7 @@ struct DirectoryParser : public llvm::cl::parser<std::string> {
       llvm::errs() << ec.message() << '\n';
       return true;
     }
+    llvm::sys::path::remove_dots(dir, true);
     Value = std::string(dir);
     return false;
   }


fw-immunant:
We might prefer `llvm::sys::fs::make_absolute` just to follow the pattern of using llvm functions here and avoid the `<filesystem>` include.

kkysen (changes requested):
Isn't this already done in #432?

randomPoison:
Oh weird you're right. I'm pretty sure I've synced and rebuilt IA2 since then, so I'm not sure why I was still seeing issues with relative paths. I'll dig into it on my end, maybe I just didn't have the latest version if IA2.

kkysen:
Does it need to fully canonicalized them or just make them absolute?  If we need full canonicalization, can you fix #432 to use the canonicalization function instead?  Either LLVM's if it has one or `std`'s if it doesn't like you did here.

ayrtonm:
> Does it need to fully canonicalized them or just make them absolute?

I think it has to be canonical because we compare with paths we get from querying clang for things.

kkysen:
> > Does it need to fully canonicalized them or just make them absolute?
> 
> I think it has to be canonical because we compare with paths we get from querying clang for things.

Ah, let's canonicalize then.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -77,6 +77,7 @@ struct DirectoryParser : public llvm::cl::parser<std::string> {
       llvm::errs() << ec.message() << '\n';
       return true;
     }
+    llvm::sys::path::remove_dots(dir, true);
```
kkysen, line 80: Lol LLVM just calls this function `remove_dots`?

randomPoison:
Yeah I'm seeing bad behavior from the rewriter (e.g. trying to copy files to the wrong directory) with the current absolute path logic. Things only seem to work with a fully canonicalized path.

I've updated the PR to use `remove_dots` after making the path absolute. I couldn't find a direct canonicalize path function, but there's an internal one in LLVM that just does `make_absolute` followed by `remove_dots`, and it works when I test with relative paths.

tools/rewriter/SourceRewriter.cpp
kkysen: Does this do canonicalization or just remove the dots? Canonicalization also needs to resolve links.

randomPoison: No this doesn't fully canonicalize, it just removes `.` and `..` from the path. I couldn't find a proper canonicalize function, at least in LLVM. If we really want full canonicalization we may need to use the stdlib one (unless someone else can find an LLVM function I'm missing).

kkysen: Let's just use the `std::filesystem` one then.

ayrtonm: Might be better to copy the LLVM approach of `make_absolute` followed by `remove_dots`. They should give the same thing but we're comparing it against paths generated by LLVM so that seems slightly less likely to create weird bugs.

kkysen: `make_absolute` and `remove_dots` won't resolve symlinks, though.  If we compare it to an LLVM path, we should also fully resolve and canonicalize it.

I think I found the LLVM canonicalization function: [`llvm::sys::fs::real_path`](https://llvm.org/doxygen/namespacellvm_1_1sys_1_1fs.html#a2363fbfbd0a348f5d81fc3d3223ecae3).

randomPoison: Oh dope, it does exist it just doesn't at all have the name I expected. In that case I think using `real_path` is the right thing to do here.

kkysen (approved):

PR #483, opened 2024-12-16
runtime, tests: require `-Werror=strict-prototypes` so every function has a prototype

kkysen:
In order to have C callgate wrappers, we need to know function prototypes, but `ReturnType foo()` functions have no prototype, as `foo(void)` is required.

ABI determination (and soon to be API, too) is done eagerly for all functions, so this errors when a function has no prototype.  We can later relax this to only requiring prototypes on functions with callgates.

Furthermore, often projects enable `-Werror=strict-prototypes` themselves, like `dav1d`. Turning this on here for the IA2 runtime and tests means that we don't have to disable it in other projects like `dav1d`.


ayrtonm:
The change itself looks fine though I don't understand why generating C call gates requires this flag across the codebase rather than just the call gate source file. I'm also not sure how you planned on getting type declarations for C call gates so I might be missing something.

kkysen:
> The change itself looks fine though I don't understand why generating C call gates requires this flag across the codebase rather than just the call gate source file.

It doesn't; it only requires them for the call gate sources as you said, but it seems a lot simpler to enforce this way, and seems like a better default in general, so I just did it everywhere.

> I'm also not sure how you planned on getting type declarations for C call gates so I might be missing something.

I'm getting them from the `clang::FunctionProtoType`.  Basically adding it in parallel to the `AbiSignature` information but storing API type information.  Still trying to figure out how to set everything up so the types used are visible where the function is declared (I'm thinking an approach like defining them in the same file as the wrapped function like you did in #451 would likely work best).

ayrtonm:
> seems like a better default in general, so I just did it everywhere.

I think the fewer requirements we have in terms of compiler flags the better

> still trying to figure out how to set everything up so the types used are visible where the function is declared

yeah that's the main issue I was asking about

kkysen:
How are we determining the ABI for untyped `f()` functions in the first place?

ayrtonm:
We currently treat `fn()` like `fn(void)`. I understand that isn't always correct since the caller can pass in arguments but I'd say that handling this properly isn't a priority right now.

Issue #415, opened 2024-10-03
Version scripts that hide symbols break linkage for containerized builds


randomPoison:
If a project has a function `foo` that is used as an address-taken function that has been rewritten by the rewriter, and the project uses a linker script to mark `foo` local, then the function gates in `libcallgates.so` will generate linker errors when they can't find `foo` in the containerized library. Note that this is basically the same error as in #414, but caused by a linker script instead of a `static` modifier.

randomPoison:
The zlib linker script can be seen here: https://github.com/immunant/zlib/blob/1849c2400150304c100af9d3e3b9b92ac0337290/zlib.map

ayrtonm:
The problem here is basically the same thing as 414 (missing the right type of callgate), but I'm not sure this can be fixed in the rewriter alone. Basically we'll need something similar to `IA2_BEGIN_NO_WRAP` to get the rewriter to create the correct callgates for these functions.

randomPoison:
I think this is just something we need to document in the usage docs. The user will have to do some additional source annotation (e.g. add the `used` attribute) to handle this case since we can't do it automatically, so we should note this limitation and tell the user how to manually fix it.

ayrtonm:
Right, though I don't think the used annotation by itself does the right thing here.

rinon:
The simplest thing I can think of is a CLI param for the rewriter that tells it to treat a symbol as internal. Were you thinking of adding an annotation to the sources instead, Ayrton? That works too, probably cleaner, but technically the linker script allows the build system to hide the symbol independently from the source file, so in some configurations it might and in some it might not.

randomPoison:
This may be fixed now (possibly by https://github.com/immunant/IA2-Phase2/pull/451). I tested this against zlib and I can now include the linker script in the compartmentalized build and it links without issue. Do we want to close this? Or is there more work that needs to be done for this?

Issue #435, opened 2024-10-07
Support compiling with `-fvisibility=hidden`


ayrtonm:
Split into:
* [x] #442
* [ ] #443

Compiling with `-fvisibility=hidden` gives the following error which @kkysen hit with dav1d. This is caused because the undefined references become local symbols in the other .so so we need explicit visibility attributes on those variables to avoid this.
```
[7/7] Linking C executable tests/two_keys_minimal/two_keys_minimal
FAILED: tests/two_keys_minimal/two_keys_minimal
[...]

PR #484, opened 2024-12-17
Add documentation for linker errors around `static` functions and linker scripts

randomPoison:
This (at least partially) addresses #414 and #415 by adding docs around those potential error cases. This covers the cases that I ran into working on zlib.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -182,6 +182,20 @@ IA2_CALL(opaque, id) // Call the the wrapper that `opaque` points to
 IA2_CAST(func, ty) // Get a struct of type `ty` pointing to `func`'s wrapper
 ```
 
+## `static` Function Pointers
+
+If a `static` function is used as an address-taken function for cross-compartment calls, then the rewriter needs to add `__attribute__((used))` to the function declaration in order for it to link correctly with the generated callgates. Normally the rewriter can achieve this, but if the `static` keyword comes from a macro the rewriter will not be able to insert the attribute. For example:
+
+```c
+#define local static
+local void function() {}
+```
+
+The same issue can happen when using 
+
+If this pattern occurs in your code, you can remove the `static` modifier or add
+`__attribute__((used))` manually post-rewriter.
+
 ## Building the callgate shim
 
 Running the source rewriter produces a `.c` that will be used to build the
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -182,6 +182,20 @@ IA2_CALL(opaque, id) // Call the the wrapper that `opaque` points to
...
+## `static` Function Pointers
+
+If a `static` function is used as an address-taken function for cross-compartment calls, then the rewriter needs to add `__attribute__((used))` to the function declaration in order for it to link correctly with the generated callgates. Normally the rewriter can achieve this, but if the `static` keyword comes from a macro the rewriter will not be able to insert the attribute. For example:
+
+```c
+#define local static
+local void function() {}
+```
+
+The same issue can happen when using
```
randomPoison, line 194: Woops, I accidentally a

docs/usage.md
```diff
@@ -249,3 +266,38 @@ initializes and makes use of TLS. If you're seeing compartment violations when
...
+may run into linker errors where `libcallgates.so` can't find symbols within a
+compartment. There are a couple of reasons why that may happen.
+
+As mentioned in the rewriter section, if a function is `static`, but the
+`static` keyword comes from inside a macro, then the rewriter will fail to emit
+an `__attribute__((used))` annotation for the function. This can lead to linker
+errors if the function is used as an address-taken function across compartments.
+
+Similar issues can occur when using linker scripts. If your project uses a
+linker script to mark functions local, you'll run into linker errors where the
```
kkysen, line 282: What's meant by "local" here exactly? `static`, or does it mean something similar but slightly different?

docs/usage.md
```diff
@@ -249,3 +266,38 @@ initializes and makes use of TLS. If you're seeing compartment violations when
...
+
+If you're seeing errors along the lines of:
+
+```txt
+/usr/bin/ld: libcallgates.so: undefined reference to `foo'
+```
+
+Then IA2 needs `foo` to not be marked local.
+
+Note also that an `__attribute__((visibility ("hidden")))` annotation can have
```
kkysen, line 301:
```suggestion
Note also that an `__attribute__((visibility("hidden")))` annotation can have
```

docs/usage.md
```diff
@@ -249,3 +266,38 @@ initializes and makes use of TLS. If you're seeing compartment violations when
...
+the same effect as hiding the symbol in a linker script. Such annotation will
+have to be removed to work in a containerized app.
```
kkysen, line 303: We figured out how to handle hidden symbols in #451.  Though it appears #435 was then re-opened, but I forget why (@ayrtonm, do you remember?).

fw-immunant, line 303: "compartmentalized", not "containerized"

docs/usage.md
fw-immunant: I don't know the canonical definition of the notion of a local symbol, but this link is a decent starting point: https://sourceware.org/binutils/docs/ld.html#VERSION

At the very least, one difference is that `static` is a C notion, while local symbols are a linker notion.

fw-immunant: It might be clearer to say "mark function **symbols** as local".

docs/usage.md
ayrtonm: It was closed by mistake. There's some edge case we still don't support but I don't remember at the moment. IIRC adding the flag to the test suite should point it out.

randomPoison: Hrm, interesting I didn't realize we had already resolved part of this. I'm trying to reproduce the issues I'm documenting here and it seems like maybe these docs are now redundant?

* `__attribute__((used))` is now being inserted correctly for static functions when `static` comes from inside a macro.
* Functions marked with `__attribute__((visibility ("hidden")))` now link correctly.

I still need to test if the issue with a version script is also fixed, but if it is I think we can close this PR.

randomPoison: Looks like the fix for `visibility("hidden")` also addressed the issue with having a linker script mark a symbol `local`? Or at least the issue was fixed at some point because I can now include the zlib linker script in the compartmentalized build and I no longer hit the linker errors I originally saw. So I'm going to close this and the related bugs.

randomPoison:
The issues this PR documents no longer seem to be happening (see https://github.com/immunant/IA2-Phase2/pull/484#discussion_r1932993068), so I'm going to close this PR.

PR #485, opened 2024-12-17
WIP Add support for checking return value in post-condition call

ayrtonm:
… call


fw-immunant:
We decided in last week's meeting to defer this feature until later, but eventually it would be good to merge something implementing this support.

PR #486, opened 2024-12-19
test_runner: print expected exit code for tests

kkysen:
I did this while debugging some tests regarding exit codes, signals, etc.  Not too important but pretty trivial.


ayrtonm (approved):
Looks fine, though I don't understand the point of `STRINGIFY(X)` rather than just `%d", X`

kkysen:
> Looks fine, though I don't understand the point of `STRINGIFY(X)` rather than just `%d", X`

I wanted the textual name, not the integer value.

PR #487, opened 2024-12-19
tests: split `post_condition` `main` test into `normal` and `corrupt_bounds` tests to have a control

kkysen:
Now we actually test that an `assert` (for now an `exit`) post condition works.

### DIFF
--- a/tests/post_condition/dav1d.c
+++ b/tests/post_condition/dav1d.c
@@ -2,18 +2,18 @@
 RUN: cat dav1d_call_gates_1.ld | FileCheck --check-prefix=LINKARGS %s
 */
 
-#include <ia2_test_runner.h>
 #include "dav1d.h"
+#include <ia2_test_runner.h>
 
 // LINKARGS: --wrap=dav1d_get_picture
 int dav1d_get_picture(Dav1dContext *const c, Dav1dPicture *const out) {
-    out->stride[0] = -1;
-    return 0;
+  out->stride[0] = -1;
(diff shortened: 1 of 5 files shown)


tests/post_condition/dav1d.c
```diff
@@ -2,18 +2,27 @@
 RUN: cat dav1d_call_gates_1.ld | FileCheck --check-prefix=LINKARGS %s
 */
 
-#include <ia2_test_runner.h>
 #include "dav1d.h"
+#include <ia2_test_runner.h>
+#include <signal.h>
+
+bool corrupt_stride = false;
```
ayrtonm, line 9: Doesn't this need to be marked as shared? It's accessed from another compartment.

kkysen: Yup, good catch.

ayrtonm: The lib compartment is using pkey 0 so it's fine.

kkysen: How do I get the lib compartment to not use pkey 0?  `#define IA2_COMPARTMENT 2` in `dav1d.c` doesn't seem to fix it either.

kkysen: Okay I think I figured it out.  I didn't realize how to define the test in cmake properly before.

fw-immunant (approved):
LGTM. The test failures on ARM are unrelated and I'm looking into them separately.

PR #488, opened 2024-12-19
runtime/tracer: passthrough unexpected signals by default

kkysen:
This allows signals like `SIGABRT` to work in tests, so now `assert`s work correctly, like in the `post_condition` test. Since post condition functions can be arbitrary and the user can handle things any way, we don't want to impose an unnecessary restriction on what fatal signals they may raise.

### DIFF
--- a/runtime/tracer/track_memory_map.c
+++ b/runtime/tracer/track_memory_map.c
@@ -475,17 +475,29 @@ enum control_flow {
   }
 
 enum wait_trap_result {
+  // Child is performing an intercepted syscall.
   WAIT_SYSCALL,
+  // Child is stopped as the result of a signal.
   WAIT_STOP,
+  // Child process group is stopped as the result of a signal.
   WAIT_GROUP_STOP,
+  // Child exited.
   WAIT_EXITED,
+  // Child was killed by an unhandled signal.
   WAIT_SIGNALED,
+  // Child received SIGSEGV.
   WAIT_SIGSEGV,
+  // Child received SIGCHLD.
   WAIT_SIGCHLD,
+  // An error occurred in waitpid() or subsequent tracing calls.
   WAIT_ERROR,
+  // Child returned from execution of clone() syscall and the child PID is available.
   WAIT_PTRACE_CLONE,
+  // Child returned from execution of fork() syscall and the child PID is available.
   WAIT_PTRACE_FORK,
+  // Child performed the exec() syscall.
   WAIT_EXEC,
+  // Child received SIGCONT.
   WAIT_CONT,
   // Child received another signal that does not require special handling.
   WAIT_NONFATAL_SIGNAL,
--- a/runtime/tracer/track_memory_map.c
+++ b/runtime/tracer/track_memory_map.c
@@ -487,32 +487,38 @@ enum wait_trap_result {
   WAIT_PTRACE_FORK,
   WAIT_EXEC,
   WAIT_CONT,
+  // Child received another signal that does not require special handling.
+  WAIT_NONFATAL_SIGNAL,
 };
 
-/* wait for the next trap from the inferior.
-
-returns the wait_trap_result corresponding to the event.
-
-if the exit is WAIT_EXITED, the exit status will be placed in *exit_status_out
-if it is non-NULL. */
-static enum wait_trap_result wait_for_next_trap(pid_t pid, pid_t *pid_out, int *exit_status_out) {
+/**
+ * Wait for the next trap from the inferior.
+ *
+ * Returns the `wait_trap_result` corresponding to the event.
+ *
+ * If the exit is WAIT_EXITED, the exit status will be placed in `*exit_status_out` if not `NULL`.
+ *
+ * If `WAIT_NONFATAL_SIGNAL` is returned, the signal received is written to `*signal_out` if not `NULL`.
+ */
+static enum wait_trap_result wait_for_next_trap(pid_t pid, pid_t *pid_out, int *exit_status_out, int *signal_out) {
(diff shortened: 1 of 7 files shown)


misc/test_runner/test_runner.c
```diff
@@ -118,12 +124,14 @@ int main() {
...
-    if WIFSIGNALED (stat) {
+    if (WIFSIGNALED(stat) && WTERMSIG(stat) != test_info->signal) {
       fprintf(stderr, "forked test child was terminated by signal %d\n", WTERMSIG(stat));
       return 1;
     }
     int exit_status = WEXITSTATUS(stat);
-    if (exit_status != test_info->exit_code) {
+    if (exit_status != test_info->exit_code ||
+        // Sometimes the signal doesn't show up with `WIFSIGNALED`, so we check the exit status as well.
+        (exit_status > 128 && exit_status - 128 == test_info->signal)) {
```
fw-immunant, line 134: This seems wrong. The first half checks if the exit status is not what's expected (which is a problem). We or this against the exit status indicating we received the expected signal (which is not a problem).

runtime/tracer/track_memory_map.c
```diff
@@ -471,15 +471,24 @@ enum wait_trap_result {
...
+ *
+ * Returns the `wait_trap_result` corresponding to the event.
+ *
+ * If the exit is `WAIT_EXITED`, the exit status will be placed
+ * in `*exit_status_out` if it is non-NULL.
+ *
+ * If the exit is `WAIT_PASSTHROUGH_SIGNAL`, the signal will be placed
+ * in `*passthrough_signal` if it is non-NULL.
+ */
+static enum wait_trap_result wait_for_next_trap(pid_t pid, pid_t *pid_out, int *exit_status_out, int *passthrough_signal) {
```
fw-immunant, line 491: This should be `passthrough_signal_out` for symmetry, since it's an out-param.

runtime/tracer/track_memory_map.c
```diff
@@ -817,7 +836,7 @@ bool track_memory_map(pid_t pid, int *exit_status_out, enum trace_mode mode) {
     case WAIT_SIGNALED: {
       fprintf(stderr, "process received fatal signal (syscall entry)\n");
       enum control_flow cf = handle_process_exit(&maps, waited_pid);
-      return false;
+      return true;
```
fw-immunant, line 839: Making this return `true` without setting `*exit_status_out` violates the doc comment for `track_memory_map`. This needs changes in `wait_for_next_trap`, which currently only writes through its copy of the `exit_status_out` pointer when returning `WAIT_EXITED`

fw-immunant (changes requested):
See inline comments.

runtime/tracer/track_memory_map.c
kkysen: I changed things to always return the `wait` status, which can then be inspected more thoroughly with all of the `W*` macros.

fw-immunant: This should still not `return true;`; doing so unconditionally exits even if other processes are still around and being traced. I actually don't understand why this case is being changed at all, given `WAIT_PASSTHROUGH_SIGNAL` is being introduced here as distinct from `WAIT_SIGNALED`. The latter used to mean "an *unexpected* signal was received", so it was handled by exiting from the entire function with failure. What situation is now being handled by this block?

fw-immunant (changes requested):
See individual comments.

fw-immunant (commented):
­

fw-immunant (approved):
It turns out that reviewing with "Comment" after reviewing with "Request changes" doesn't reset the "Request changes" status, so reviewing with "Approve". I believe I've addressed my own comments here but this should still get review from someone else.

kkysen (commented):
@fw-immunant, is there anything else I should look at here now?

runtime/tracer/track_memory_map.c
```diff
@@ -849,6 +871,12 @@ bool track_memory_map(pid_t pid, int *exit_status_out, enum trace_mode mode) {
       }
       continue;
       break;
+    case WAIT_NONFATAL_SIGNAL:
+      if (ptrace(continue_request, waited_pid, 0, sig) < 0) {
+        perror("could not PTRACE_SYSCALL...");
+      }
+      continue;
```
ayrtonm, line 878: Should this just be `break;`?

runtime/tracer/track_memory_map.c
```diff
@@ -829,6 +850,7 @@ bool track_memory_map(pid_t pid, int *exit_status_out, enum trace_mode mode) {
         perror("could not ptrace(continue_request)...");
       }
       continue;
+      break;
```
ayrtonm, line 853: Don't we need to remove the `continue;` on the previous line?

runtime/tracer/track_memory_map.c
```diff
@@ -924,14 +952,16 @@ bool track_memory_map(pid_t pid, int *exit_status_out, enum trace_mode mode) {
     }
     case WAIT_SIGNALED: {
       fprintf(stderr, "process received fatal signal (syscall entry)\n");
-      enum control_flow cf = handle_thread_exit(&maps, waited_pid);
-      return false;
+      propagate(handle_thread_exit(&maps, waited_pid));
+
+      // this process is gone, so wait for a new one
+      continue;
```
ayrtonm, line 958: If this works the same as the `continue; break;` above we should stick to one or the other.

ayrtonm (approved):
Took a cursory look and LGTM modulo consistency with use of `continue; break;` vs just `continue;` in switch case statements.

PR #490, opened 2025-01-06
runtime: condition x86_64 `ASSERT_PKRU` asm on `__x86_64__`

kkysen:
This leaves `ASSERT_PKRU` to do nothing on other architectures (aarch64). Since the asm is x86_64 asm, it doesn't work on aarch64 with `-DIA2_DEBUG=1` without this.

This came up when trying to rewrite and compartmentalize `dav1d` on aarch64.


runtime/libia2/include/ia2.h
```diff
@@ -107,6 +107,9 @@
   "movq %r11, %rdx\n"                           \
   "movq %r10, %rcx\n"
 /* clang-format on */
+#else
+// TODO aarch64
```
ayrtonm, line 111: I think `ASSERT_PKRU` is the only thing `IA2_DEBUG` toggles so we should be able to `#error "unimplemented"` here. If not, then this is fine too.

ayrtonm (approved):

runtime/libia2/include/ia2.h
kkysen: I'd prefer not to `#error`, since then `-DIA2_DEBUG=1` only works conditionally, but I could do a `#warning` if you prefer.

ayrtonm: That's fine, just leave it as a comment then

ayrtonm: Sorry in my last comment I didn't realize `#warning` would only warn in that specific config. Either `#warning` or comment is good with me.

PR #492, opened 2025-01-06
docs: switch back to suggesting `partition-alloc-padding`, as `partition-alloc` is not a target

kkysen:
@ayrtonm, I had changed this back after you said https://github.com/immunant/IA2-Phase2/pull/454#discussion_r1828322744, but when I re-encountered it in building `dav1d` for aarch64, I realized the `partition-alloc` target does not exist. There is only:
* `libpartition-alloc.so`
* `partition-alloc_unpadded`
* `partition-alloc-padding`

So I'm just switching these back to `partition-alloc-padding`.


fw-immunant (approved):
LGTM. Our documented CMake targets must be existing ones, of course.

PR #495, opened 2025-01-07
ARM: Align stack pointer to support SCTLR_EL1.SA0

### LINKED ISSUE #469: Enforce MTE protection for writeable file-backed mappings
See [this comment](https://github.com/immunant/IA2-Phase2/issues/314#issuecomment-2019145083) on #314 for details

ayrtonm:
It seems that the pixel's kernel sets `SCTLR_EL1.SA0` (unconditionally?) so stack accesses at EL0 must be aligned to 16 bytes. This fixes a few accesses in our handwritten asm to enable running the minimal test on MTE-enabled hardware. There are likely some stack accesses in the callgates that also need to be fixed and I'd like to fix the TODO in `allocate_stack` before merging this (@fw-immunant any ideas there?). I checked that when the minimal test binary is run from a tmpfs MTE protection for its writeable file-backed mappings are enforced (see #469).

### DIFF
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -271,12 +271,12 @@ asm(".macro movz_shifted_tag_x18 tag\n"
         "mov x9, sp\n"                                                         \
         /* switch to newly allocated stack */                                  \
         "mov sp, %0\n"                                                         \
-        /* push old stack pointer to new stack */                              \
-        "str x9, [sp], #-8\n"                                                  \
+        /* push sp in x9 and a dummy reg to keep new stack 16-byte aligned */  \
+        "stp x9, x10, [sp, #-16]!\n"                                           \
         /* initialize TLS */                                                   \
         "bl init_tls_" #i "\n"                                                 \
         /* pop old stack pointer from new stack */                             \
-        "ldr x9, [sp, #8]!\n"                                                  \
+        "ldp x9, x10, [sp], #16\n"                                             \
         /* save pointer to new stack */                                        \
         "mov x10, sp\n"                                                        \
         /* switch to old stack */                                              \


fw-immunant:
It turns out we must do this regardless of the kernel SCTLR flags because it's required by the AArch64 ABI and assumed by LLVM when performing codegen:
> I thought it was an LLVM codegen bug but it turns out LLVM was relying on the "sp is always 16B-aligned" guarantee when emitting instructions so it knew it was safe to AND an sp-derived pointer by #0xfffffffffffff0 (note trailing 0) while tagging it

(from [here](https://immunant.slack.com/archives/C010R2PU3NF/p1734564973523199))

I had locally made changes equivalent to the `allocate_stack` change here, so r+ on that account. But I don't understand the second commit here--can it have a comment explaining why a dummy register should be pushed and popped? What failures/problems occur without this commit?

ayrtonm:
> I had locally made changes equivalent to the allocate_stack change here, so r+ on that account.

Do some accesses/calls in the call gates assume the initial stack pointer mod 16 is 8?

fw-immunant:
> > I had locally made changes equivalent to the allocate_stack change here, so r+ on that account.
> 
> Do some accesses/calls in the call gates assume the initial stack pointer mod 16 is 8?

On x86_64, I'm pretty sure call gates don't rely on this because they don't use aligned mov instructions, but the more relevant case (where I'm seeing errors in the `structs` test) is AArch64, and I'm less familiar with that code. We do some padding based on count of registers, and that might be at fault for the issues I see on my branch. But I don't think that's a concern of this PR, is it?

fw-immunant (changes requested):
Please document why we should push a dummy register as well as the old stack pointer. Otherwise looks good.

fw-immunant:
> Please document why we should push a dummy register as well as the old stack pointer. Otherwise looks good.

We were going through this in Slack, but for posterity, the first section of this article does a good job of explaining why this necessary: https://community.arm.com/arm-community-blogs/b/architectures-and-processors-blog/posts/using-the-stack-in-aarch64-implementing-push-and-pop

PR #496, opened 2025-01-13
Support shared heap allocations on ARM

fw-immunant:
~Needs some cleanup and still debugging structs test.~

### DIFF
--- a/external/chromium/src/base/allocator/partition_allocator/thread_isolation/thread_isolation.cc
+++ b/external/chromium/src/base/allocator/partition_allocator/thread_isolation/thread_isolation.cc
@@ -37,7 +37,7 @@ void WriteProtectThreadIsolatedMemory(ThreadIsolationOption thread_isolation,
 #elif BUILDFLAG(ENABLE_MTE_ISOLATION)
   partition_alloc::internal::TagMemoryWithMte(
       thread_isolation.enabled ? thread_isolation.pkey : kDefaultPkey, address,
-      size);
+      (size+0xfff)&~0xfff);
 #else
 #error unexpected thread isolation mode
 #endif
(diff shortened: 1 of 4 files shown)


fw-immunant:
CI is failing because QEMU [no longer propagates MTE exceptions to userspace in our patched QEMU](https://github.com/qemu/qemu/commit/d5169c7608dfce89dd1f4df76f82832e1241a790), preventing us from handling segfaults with `CHECK_VIOLATION`.

fw-immunant:
The AArch64 structs test is also failing locally for me on `main`, so it wasn't regressed by these changes.

fw-immunant:
This fixes the `read_config` test.

external/chromium/src/base/allocator/partition_allocator/thread_isolation/thread_isolation.cc
```diff
@@ -37,7 +37,7 @@ void WriteProtectThreadIsolatedMemory(ThreadIsolationOption thread_isolation,
 #elif BUILDFLAG(ENABLE_MTE_ISOLATION)
   partition_alloc::internal::TagMemoryWithMte(
       thread_isolation.enabled ? thread_isolation.pkey : kDefaultPkey, address,
-      size);
+      (size+0xfff)&~0xfff);
```
ayrtonm, line 40: Can we use PAGE_SIZE if it's defined?

ayrtonm (approved):
One minor nit but otherwise LGTM

external/chromium/src/base/allocator/partition_allocator/thread_isolation/thread_isolation.cc
fw-immunant: PAGE_SIZE is not defined here, but it would be cleaner to use it and a ROUND_UP macro. I don't know if it's worth the trouble to redefine them or factor out an include we could use for this one occurrence.

PR #498, opened 2025-01-21
rewriter: wrap callgate header in `#ifndef __ASSEMBLER__` so it doesn't get interpreted as asm

kkysen:
In `meson`, `*.S` asm files are pre-processed with the C preprocessor, and thus C flags are used to compile `*.S` asm files (unlike `*.s` asm files). To fix this on the `meson` side, we'd have to manually add args to each target instead of using `add_project_arguments(args, language: 'c')`, so this way is simpler.


ayrtonm (approved):
looks good to me

Issue #412, opened 2024-10-03
Assignments to function pointer fields get rewritten incorrectly


randomPoison:
If we have some pre-rewriter code that does `foo.bar = 0;` where `foo.bar` is a function pointer, the rewritten code is `foo.bar = (fn_struct)0;`, where `fn_struct` is the rewritten function pointer struct type. This doesn't compile because we can't cast 0 to a struct.

The way I've manually fixed this before is to change it to `foo.bar.ptr = 0`, i.e. assign directly to the hidden `ptr` field, but this feels less than idea to me since (I think) we want to keep that inner ptr field hidden. We might have to come up with a better way to translate these statements.

rinon:
Can you add this to one of the tests (or a new test) so we have an example?

randomPoison:
Is it possible/reasonable to add a test without a fix? I assumed we'd add a test case for this when we work on the fix.

ayrtonm:
> Is it possible/reasonable to add a test without a fix?

Sure, just push a test to a new branch and one of us can open a PR on top of that commit when we go to fix it.

ayrtonm:
I got an ast matcher for this working but should probably handle the case where there's an explicit cast before the 0 literal like in #411.

Issue #411, opened 2024-10-03
Rewriter generates incorrect `IA2_ADDR` invocations for null checks


randomPoison:
If we have some pre-rewriter code that does `fn_ptr == 0`, this gets rewritten to `IA2_ADDR(fn_ptr) == IA2_ADDR((fn_struct)0)`. The latter `IA2_ADDR` is incorrect, since the "address" being taken is a literal, rather than a function ptr struct. It would be better if the rewriter could detect this case and not emit the incorrect `IA2_ADDR` expression.

rinon:
Can you add a test for this, please?

ayrtonm:
I'm handling this together with #412 but I can't reproduce this part. In my new tests cases 3 and 4 do get rewritten as expected.
```
diff --git a/tests/rewrite_fn_ptr_eq/main.c b/tests/rewrite_fn_ptr_eq/main.c
index f8fe55e22..8a5c7f2b0 100644
--- a/tests/rewrite_fn_ptr_eq/main.c
+++ b/tests/rewrite_fn_ptr_eq/main.c
@@ -78,4 +78,19 @@ Test(rewrite_fn_ptr_eq, main) {
     if (y || !fn) { }
     // REWRITER: if (x && IA2_ADDR(fn) && y) { }
     if (x && fn && y) { }
+
+    // REWRITER: fn = (typeof(fn)) { NULL };
+    fn = NULL;
+
+    // REWRITER: fn = (typeof(fn)) { 0 };
+    fn = 0;
+
+    // REWRITER: if (IA2_ADDR(fn) == 0) { }
+    if (fn == 0) { }
+
+    // REWRITER: if (IA2_ADDR(mod.fn) == 0) { }
+    if (mod.fn == 0) { }
+
+    // REWRITER: res = 0;
+    res = 0;
 }
```

randomPoison:
Maybe putting a cast before the 0 matters? An example case in zlib is

```c
if (strm->zalloc == (alloc_func)0) {}
```

which gets rewritten as 

```c
if (IA2_ADDR(strm->zalloc) == IA2_ADDR((alloc_func)0)) {}
```

ayrtonm:
yeah the explicit cast is the issue.

PR #502, opened 2025-01-27
Fix translation of literal 0 in function pointer assignment/comparison

randomPoison:
This PR builds on https://github.com/immunant/IA2-Phase2/pull/438 to fix cases where the rewriter was generating invalid code when assigning 0 to a function pointer or when comparing a function pointer against 0. This fixes https://github.com/immunant/IA2-Phase2/issues/412.

@ayrtonm This doesn't fully address #411 (which you were trying to do in the original PR), but I'm having a harder time getting that working. Does it make sense to you to land this portion of the changes on its own?

### DIFF
--- a/tests/rewrite_fn_ptr_eq/main.c
+++ b/tests/rewrite_fn_ptr_eq/main.c
@@ -84,6 +84,16 @@ Test(rewrite_fn_ptr_eq, main) {
     // REWRITER: fn = (typeof(fn)) { NULL };
     fn = NULL;
 
+    // Do we handle custom NULL defines?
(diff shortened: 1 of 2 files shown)


randomPoison:
Woops, tests were failing because I didn't realize I needed `ninja check` in order to fully run the test suite. I've updated the tests to reflect the changed rewriter logic, mainly changing `{ NULL }` into `{ 0 }`, since it seems reasonable to normalize the output to `0` instead of trying to figure out which null macro to use in the output. Let me know if any of that doesn't look right.

PR #503, opened 2025-01-27
ci: attempt to fix rtlib caching

fw-immunant:
Sadly I think the CI run for this will likely not populate the cache because it's a PR, but there might be enough output to see if this fixes the problem.


fw-immunant:
> [`Cache saved with key: Linux-rtlibs`](https://github.com/immunant/IA2-Phase2/actions/runs/12997866674/job/36249755463?pr=503#step:27:5)

PR #504, opened 2025-02-03
Add means to run one test without forking

fw-immunant:
This is handy for debugging some subtler interactions of our test runner and process state.


ayrtonm:
The change looks fine, but it'd be good to note that `.exit_code` and `.signal` won't be checked.

fw-immunant:
> The change looks fine, but it'd be good to note that `.exit_code` and `.signal` won't be checked.

That exit status isn't checked is mentioned in comments and stderr output.

PR #505, opened 2025-02-04
Fix structs test on ARM

fw-immunant:
It turns out we were not aligning our stacks properly in ARM call gates since 13c495ddec8658101c2453362e76eba50518e1c3.

### DIFF
--- a/tools/rewriter/GenCallAsm.cpp
+++ b/tools/rewriter/GenCallAsm.cpp
@@ -525,7 +525,9 @@ static void x86_emit_intermediate_pkru(AsmWriter &aw, uint32_t caller_pkey, uint
 
 static void emit_copy_args(AsmWriter &aw, const std::vector<ArgLocation> &args,
                            const std::optional<std::vector<ArgLocation>> &wrapper_args,
-                           size_t stack_return_size, size_t stack_return_padding, int stack_alignment,
+                           size_t stack_return_size, size_t stack_return_padding,
+                           size_t indirect_arg_size, size_t indirect_arg_padding,
+                           int stack_alignment,
                            size_t stack_arg_size, size_t stack_arg_padding, size_t wrapper_stack_arg_size,
                            uint32_t caller_pkey, Arch arch) {
   if (arch == Arch::X86) {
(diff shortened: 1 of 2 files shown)


runtime/libia2/init.c
```diff
@@ -32,8 +32,12 @@ char *allocate_stack(int i) {
   /* Tag the allocated stack pointer so it is accessed with the right pkey */
   stack = (char *)((uint64_t)stack | (uint64_t)i << 56);
 #endif
+#ifdef __aarch64__
+  return stack + STACK_SIZE - 16;
```
ayrtonm, line 36: Is there a good reason to not also do this on x86?

tools/rewriter/GenCallAsm.cpp
```diff
@@ -974,6 +970,21 @@ std::string emit_asm_wrapper(AbiSignature sig,
   llvm::errs() << "Generating wrapper for " << sig_string(sig, target_name) << "\n";
   auto rets = allocate_return_locations(sig, arch);
 
+  size_t indirect_arg_size = 0;
+  for (auto &arg : args) {
+    if (arg.is_indirect()) { // Only ever true on AArch64 for now.
```
ayrtonm, line 975: Maybe let's assert that arch is aarch64 in this branch then?

tools/rewriter/GenCallAsm.cpp
```diff
@@ -1026,8 +1037,11 @@ std::string emit_asm_wrapper(AbiSignature sig,
     | top |Top of the stack (stack grows down on AArch64). This address is
     |     |aligned to 16 bytes.
     +-----+
-    |ind  |Space for arguments passed indirectly (i.e. in memory). These arguments
-    |args |will point to this region.
+    |ind  |Space for arguments passed indirectly (i.e. in memory). Indirect arguments
+    |args |are passed as pointers into this region; for access in another compartment
+    |     |we must copy its contents and translate the pointers.
+    +-----+
+    |align|Space to align the stack back to 16 bytes after indirect arguments.
```
ayrtonm, line 1044: Was there a good reason to not merge this with the align section between ret ptr and stack args? We can leave it as-is if it was just for implementation simplicity.

fw-immunant: We can't directly merge this since they're not adjacent. We could have reordered things but doing so was going to be a more complex change and I wasn't sure at the time that this was necessarily going to fix things, so I went for the less invasive change. We could file a bug to reorder these as doing so would probably save space sometimes, but I'd be surprised if there was any performance difference.

tools/rewriter/GenCallAsm.cpp
fw-immunant: Done.

runtime/libia2/init.c
fw-immunant: This merged as part of #496: https://github.com/immunant/IA2-Phase2/pull/496/commits/5a6fc059c6ea1ed21775280d91c1cb2fc30a05be

ayrtonm (approved):

PR #506, opened 2025-02-04
ci: use LLVM patched to apply x18 tagging to atomic operations

fw-immunant:


fw-immunant:
I *think* this will properly avoid using the cached clang since this changes which ref it depents on. We'll see.

fw-immunant:
Superseded by #508, which includes this commit.

PR #510, opened 2025-02-18
rewriter: switch from hardcoded `--enable-dav1d_get_picture-post-condition` to `IA2_{PRE,POST}_CONDITION()` attribute annotation macros

kkysen:
This is much more robust and flexible and is much closer to the end state for how this API will look.  It also makes the tests much simpler by encoding the options in the source code to be detected, rather than needing to specify them in the CLI or hardcoded like before.

Will the full gamut of the attribute API is implemented here, we still only support a single post-condition function, and no pre-condition functions yet.  Those will come next.

I also tested this more thoroughly in another branch with mostly complete multiple pre- and post-condition functions, but that implementation is more complex, so I wanted to put this in a separate PR.

### DIFF
--- a/tools/rewriter/GenCallAsm.cpp
+++ b/tools/rewriter/GenCallAsm.cpp
@@ -972,7 +972,7 @@ std::string emit_asm_wrapper(AbiSignature sig,
 
   size_t indirect_arg_size = 0;
   for (auto &arg : args) {
-    if (arg.is_indirect()) { // Only ever true on AArch64 for now.
+    if (arg.is_indirect()) {         // Only ever true on AArch64 for now.
       assert(arch == Arch::Aarch64); // Let's find out if this ever gets encountered on x86_64.
       size_t align = std::max(arg.align(), (size_t)8);
       if (stack_arg_size % align != 0) {
--- a/tools/rewriter/GenCallAsm.cpp
+++ b/tools/rewriter/GenCallAsm.cpp
@@ -501,7 +501,7 @@ static void emit_prologue(AsmWriter &aw, uint32_t caller_pkey, uint32_t target_p
     }
 
     if (save_param_regs) {
-      llvm::report_fatal_error("--enable-dav1d_get_picture-post-condition is not yet supported on aarch64");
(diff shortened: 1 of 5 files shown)


tools/rewriter/SourceRewriter.cpp
```diff
@@ -171,6 +175,35 @@ static std::string append_name_if_nonempty(const std::string &new_type,
   return new_type + (name.empty() ? "" : " ") + name;
 };
 
+/// Collects in a multimap (`funcs`) all of the function names
+/// with an annotation starting with `prefix`.
+/// The annotation minus the prefix is the key,
+/// and the function name is the value.
```
ayrtonm, line 181: Maybe mention that this is used for the pre/post-condition annotations.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -1102,12 +1102,15 @@ std::string emit_asm_wrapper(AbiSignature sig,
...
   }
 
-  // For now, we hardcode the existence and name of the post-condition function.
-  // The name is `${target_name}_post_condition`,
-  // and we only do this for `dav1d_get_picture`.
   std::optional<std::string> target_post_condition_name = std::nullopt;
-  if (enable_dav1d_get_picture_post_condition && target_name && *target_name == "dav1d_get_picture") {
-    target_post_condition_name = *target_name + "_post_condition";
+  if (target_name && arch == Arch::X86) {
+    // TODO implement on aarch64
```
ayrtonm, line 1107: We should make it a hard error on ARM to have any pre/post-condition functions since we don't support them yet.

ayrtonm (approved):
A few minor nits but generally looks good.

tools/rewriter/GenCallAsm.cpp
kkysen: If I do that, how do I mark the tests as x86_64 only?  Otherwise the tests will just fail.

ayrtonm: It's fine then, you can leave it as is.

PR #511, opened 2025-02-18
rewriter: implement support for a single pre-condition function

kkysen:


tools/rewriter/GenCallAsm.cpp
```diff
@@ -1160,10 +1178,27 @@ std::string emit_asm_wrapper(AbiSignature sig,
...
-    emit_post_condition_fn_call(aw, arch, *target_post_condition_name);
+  // Call the post-condition function for this target function.
+  // The calls happens in the caller's compartment.
+  if (!post_conditions.empty()) {
+    // Args are callee saved, so we need to restore them.
+    emit_restore_args(aw, arch);
+
+    add_comment_line(aw, "Align stack");
+    if (arch == Arch::X86) {
+      add_asm_line(aw, "subq $8, %rsp");
```
ayrtonm, line 1189: Doesn't this and the `addq $8, %rsp` below depend on the number of arguments?

tests/pre_post_condition/include/src.h
```diff
@@ -0,0 +1,3 @@
+#pragma once
+
+void f1(int a, int *b);
```
ayrtonm, line 3: We should have a few tests with a different number of arguments. They don't have to be extensive, but in particular I think a function with an odd number of arguments that calls `printf` (which often uses x86 simd instructions that assume rsp is aligned to 16 bytes before every function call) is likely fail because this PR currently misaligns the stack.

ayrtonm (changes requested):
Seems like we'll misalign the stack pointer in some cases.

tools/rewriter/GenCallAsm.cpp
fw-immunant: I agree, it isn't clear to me what the alignment of the stack is going to be when we hit this code. That said, this PR isn't changing the effective control flow here from what I can tell--it's just inlining this from `emit_post_condition_fn_call`.

fw-immunant (approved):
LGTM. One inline comment, but that code is only moved in this PR and seems to run in the same cases as previously, so it's not a blocker.

tools/rewriter/GenCallAsm.cpp
kkysen: From offline discussions, it also seems like this shouldn't depend on the number of arguments, since we always assume 6 args.  There are still some issues with stack alignment, though, but I'll fix those in a separate PR instead of at the base of this PR stack.

PR #513, opened 2025-02-21
Always test optimized builds on AArch64

fw-immunant:
The persistent weird failures we're seeing in CI couldn't reproduce locally because I was doing an optimized build locally. We should test both configurations in CI, but for now let's get CI green by always testing optimized builds.


randomPoison (approved):

fw-immunant:
This gets the simple1 test to pass in CI.

PR #514, opened 2025-02-24
rewriter: wrap `AbiSignature` in `FnSignature` that also stores API info (arg/return types)

kkysen:
This wraps `AbiSignature` in a more generic `FnSignature` that also contains API info as well as the existing ABI info.  The API info is stored in `ApiSignature`, which contains `Param`s for the args and return.  Each `Param` contains the parameter name (if a parameter), the type name, the canonical type name, and a type ID based on the canonical name.  For now, the type ID is counted by interning the canonical types, so it's unique within a single `ia2-rewriter` invocation.

Next, we can use this API info to both define C functions (if we want to) and call into the `type-registry` library (#493) with type IDs for the arguments and return type.

### DIFF
--- a/tools/rewriter/DetermineAbi.cpp
+++ b/tools/rewriter/DetermineAbi.cpp
@@ -177,15 +177,7 @@ AbiSignature determineAbiSignature(const clang::CodeGen::CGFunctionInfo &info,
   return sig;
 }
 
-FnSignature determineFnSignature(const clang::CodeGen::CGFunctionInfo &info,
-                                 const clang::ASTContext &astContext, Arch arch) {
-  return (FnSignature){
-      .abi = determineAbiSignature(info, astContext, arch),
-      .variadic = info.isVariadic(),
-  };
-}
-
-FnSignature determineFnSignatureForDecl(const clang::FunctionDecl &fnDecl, Arch arch) {
+AbiSignature determineAbiSignatureForDecl(const clang::FunctionDecl &fnDecl, Arch arch) {
   clang::ASTContext &astContext = fnDecl.getASTContext();
 
   // set up context for codegen so we can ask about function ABI
@@ -216,7 +208,7 @@ FnSignature determineFnSignatureForDecl(const clang::FunctionDecl &fnDecl, Arch 
 
   auto name = fnDecl.getNameInfo().getAsString();
   const auto &info = cgFunctionInfo(cgm, fnDecl);
-  DEBUG(llvm::dbgs() << "determineFnSignatureForDecl: " << name << "\n");
+  DEBUG(llvm::dbgs() << "determineAbiSignatureForDecl: " << name << "\n");
 
   const auto &convention = info.getEffectiveCallingConvention();
 
@@ -228,12 +220,12 @@ FnSignature determineFnSignatureForDecl(const clang::FunctionDecl &fnDecl, Arch 
            info.getASTCallingConvention());
     abort();
   }
-  return determineFnSignature(info, astContext, arch);
+  return determineAbiSignature(info, astContext, arch);
 }
 
-FnSignature determineFnSignatureForProtoType(const clang::FunctionProtoType &fpt,
-                                             clang::ASTContext &astContext, Arch arch) {
-  // FIXME: This is copied verbatim from determineFnSignatureForDecl and could be
+AbiSignature determineAbiSignatureForProtoType(const clang::FunctionProtoType &fpt,
+                                               clang::ASTContext &astContext, Arch arch) {
+  // FIXME: This is copied verbatim from determineAbiSignatureForDecl and could be
   // factored out. This depends on what we do with PR #78 so I'm leaving it as
   // is for now.
 #if LLVM_VERSION_MAJOR >= 15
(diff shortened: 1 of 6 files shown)


randomPoison (commented):
There seem to be a lot of unrelated formatting changes in here. Are you making these changes manually or is this formatting style being enforced by clang tidy or something? Either way it'd be better to split the formatting changes into their own PR, rather than lumping them in here with other functional changes.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -1212,5 +1212,12 @@ std::string emit_asm_wrapper(AbiSignature sig,
...
 
-  return aw.ss.str();
+  std::string wrapper = "asm("s;
+  if (as_macro) {
+    wrapper += "\\";
+  }
+  wrapper += "\n";
+  wrapper += aw.ss.str();
+  wrapper += ");\n";
+  return wrapper;
```
randomPoison, line 1222: Why is this necessary now? It's not clear to me why your changes here require that we now manually wrap the generated asm in `asm()` when we previously could just return `aw.ss.str()` directly, but I'm also not familiar with this code.

tools/rewriter/DetermineAbi.cpp
```diff
@@ -249,5 +253,73 @@ AbiSignature determineAbiForProtoType(const clang::FunctionProtoType &fpt,
       cgm,
       fpt.getCanonicalTypeUnqualified().castAs<clang::FunctionProtoType>());
 
-  return determineAbi(info, astContext, arch);
+  return determineAbiSignature(info, astContext, arch);
+}
+
+// Not thread safe.
+std::unordered_map<std::string, uint32_t> type_ids;
```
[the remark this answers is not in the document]
kkysen: We discussed this in the meeting, but we use globals elsewhere; necessary evil I guess and clang shouldn't run things multithreaded here.

tools/rewriter/GenCallAsm.cpp
kkysen: I just moved it from all of the many call sites.  Might not be necessary now after I changed how I was doing things, but it should be a non-functional change.

kkysen:
> There seem to be a lot of unrelated formatting changes in here. Are you making these changes manually or is this formatting style being enforced by clang tidy or something?  Either way it'd be better to split the formatting changes into their own PR, rather than lumping them in here with other functional changes.

There are a lot of files that aren't formatted (nothing in CI checks this), but it's convenient to autoformat when I'm editing, so I formatted these files that I'm editing here.  I did split them into separate commits, but I wasn't sure if a whole PR is worth it.  But I can do that; it's pretty trivial.

randomPoison:
> I did split them into separate commits, but I wasn't sure if a whole PR is worth it. But I can do that; it's pretty trivial.

Nah, don't worry about it here. I would just try to do separate PRs for any non-trivial formatting in the future.

tools/rewriter/GenCallAsm.cpp
randomPoison: Ah okay that's fine then, I just hadn't noticed the places where you had removed it at the call sites.

randomPoison (approved):
Looks good to me, though I'm not familiar enough with the rewriter to know if this is the best solution. I defer to @ayrtonm and @fw-immunant to judge that.

tools/rewriter/DetermineAbi.cpp
```diff
@@ -261,5 +253,72 @@ FnSignature determineFnSignatureForProtoType(const clang::FunctionProtoType &fpt
...
+
+uint32_t get_type_id(clang::QualType type) {
+  auto canonical_name = type.getCanonicalType().getAsString();
+  // Constructs the default value, 0, if the key doesn't exist yet.
+  auto type_id = type_ids[canonical_name];
+  if (type_id == 0) {
+    type_id = type_ids.size() + 1;
+    type_ids[canonical_name] = type_id;
+  }
+  return type_id;
```
fw-immunant, line 270: AFAIK this will collide for identically-named structs defined in different files. Canonical types are documented as being suitable for pointer-based comparison, so it might be reasonable to simply use `.getCanonicalType().getAsOpaquePtr()` here.

tools/rewriter/DetermineAbi.cpp
```diff
@@ -261,5 +253,72 @@ FnSignature determineFnSignatureForProtoType(const clang::FunctionProtoType &fpt
...
+}
+
+Param determineParam(std::string name, clang::QualType type) {
+  return (Param) {
+      .name = name,
+      .type_name = type.getAsString(),
+      .canonical_type_name = type.getCanonicalType().getAsString(),
+      .type_id = get_type_id(type),
+  };
+}
```
fw-immunant, line 280: Nit, but I'm not a fan of `determineParam` as the name here; it doesn't seem like this function is determining a parameter, it's just obtaining the type info for a parameter.

fw-immunant (approved):
A couple non-critical nits, nothing blocking.

tools/rewriter/DetermineAbi.cpp
kkysen: Their lifetimes don't last long enough, though.

tools/rewriter/DetermineAbi.cpp
kkysen: I named it `determine*` since all of the other functions were `determine*` like `determineAbiForDecl`, etc.  If you have a better name I can use that, but I'm not sure what to name it otherwise.

tools/rewriter/DetermineAbi.cpp
kkysen: That's what I had originally tried.

fw-immunant: That's surprising but unfortunate; in any case if we stick to hashing names it would be good to note with a TODO comment that there could be collisions.

tools/rewriter/DetermineAbi.cpp
fw-immunant: This effectively seems to be the constructor for `Param`, so it could just be `createParam`.

tools/rewriter/DetermineAbi.cpp
kkysen: Sure, and I'll open an issue for this.

tools/rewriter/DetermineAbi.cpp
kkysen: Actually I just realized #522 deletes (changes and inlines) `determineParam` so I'm just gonna leave it as is then.

tools/rewriter/DetermineAbi.cpp
kkysen: Moved to #553.

PR #517, opened 2025-03-03
add `#pragma once` to headers w/ no include guards

kkysen:
A bunch of headers are missing include guards (which are mostly `#pragma once`s), so this adds them.  I ran into this with `tools/rewriter/TypeOps.h` but there are more headers missing so this fixes them all.


fw-immunant (approved):
NFC, looks fine.

PR #518, opened 2025-03-03
format: run `rg --files --type c --type cpp --glob '!external/' | xags clang-format -i`

kkysen:
A lot of files aren't formatted correctly, so whenever I edit files and format, it adds a lot of extra formatting fixes, which makes PRs much harder to review with all of this extraneous info.

We should probably add this to CI/pre-commit hook, but this should be enough for now.

This is a more complete fix to #516, which was getting in the way of reviewing #514.


fw-immunant:
This looks to have broken `lit` testing for the rewriter.

kkysen:
> This looks to have broken `lit` testing for the rewriter.

Oh weird, it worked fine locally. I'll fix it later at some point, not too important if it's not trivial.

fw-immunant:
I'd like to wait until other PRs land before we merge this, since it'll create a *lot* of conflicts. There are also some formatting decisions that clang-format makes here that really destroy readability, mostly around macros. Can we ask clang-format not to alter macro definitions? E.g. see `ia2_internal.h` where it puts comments and newlines on the wrong lines.

kkysen:
> I'd like to wait until other PRs land before we merge this, since it'll create a _lot_ of conflicts. 

It makes it a bit easier for me actually, as most of those PRs are mine, and they've been having conflicts on formatting since not everything is formatted, especially when I try to move them around and re-order them.  Having everything consistently formatted should make that easier.

> There are also some formatting decisions that clang-format makes here that really destroy readability, mostly around macros. Can we ask clang-format not to alter macro definitions? E.g. see `ia2_internal.h` where it puts comments and newlines on the wrong lines.

I disabled `clang-format` there.  It was already disabled for the `__x86_64__` macro for the same reason, just not the `aarch64` one.  For the other macros, I checked and it's mostly just lining up the `\`s, not moving comments.

PR #520, opened 2025-03-03
type-registry: switch from `cdylib` to `staticlib` and add it as a dependency to `libia2`

kkysen:


kkysen:
https://github.com/immunant/IA2-Phase2/actions/runs/13639316646/job/38126137570?pr=520#step:3:165

```shell
FAILED: runtime/type-registry/debug/libtype_registry.a /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/runtime/type-registry/debug/libtype_registry.a 
cd /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/runtime/type-registry && CARGO_TARGET_DIR=/mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/runtime/type-registry cargo build --manifest-path /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/runtime/type-registry/Cargo.toml --profile dev
error: rustc 1.83.0 is not supported by the following package:
  type-registry@0.1.0 requires rustc 1.84
```

@fw-immunant, do you know where the rustc 1.83 version is coming from here?  I don't see it in `ci.yml`.

fw-immunant:
> https://github.com/immunant/IA2-Phase2/actions/runs/13639316646/job/38126137570?pr=520#step:3:165
> 
> ```shell
> FAILED: runtime/type-registry/debug/libtype_registry.a /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/runtime/type-registry/debug/libtype_registry.a 
> cd /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/runtime/type-registry && CARGO_TARGET_DIR=/mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/build/runtime/type-registry cargo build --manifest-path /mnt/ssd2/home/perl/actions-runner/_work/IA2-Phase2/IA2-Phase2/runtime/type-registry/Cargo.toml --profile dev
> error: rustc 1.83.0 is not supported by the following package:
>   type-registry@0.1.0 requires rustc 1.84
> ```
> 
> @fw-immunant, do you know where the rustc 1.83 version is coming from here? I don't see it in `ci.yml`.

It must come from the runner environment on donna somewhere? cc @thedataking

kkysen:
> It must come from the runner environment on donna somewhere?

Is this from https://github.com/rust-lang/rustup/issues/4211, or is this something else?

fw-immunant:
See if adding a `rustup toolchain install` step to CI (and maybe a `rust-toolchain.toml` to the repo) fixes things, I guess.

fw-immunant:
It seems like the consumer (libia2) doesn't seem to include `CARGO_ARCH_SUFFIX` in the path to the `.a`.

fw-immunant:
I think `target_link_libraries` should just be able to refer to the `type-library` target instead of its internal `.a`.

fw-immunant:
That said, it also seems like libia2 shouldn't actually need to depend on the type registry library, should it? Generated call-gates may, if the rewriter inserts type-registry based validation, but it seems like it'll just be dead code otherwise. Presumably just tests that depend on the type registry should have cmake dependencies declared on it.

kkysen:
> I think `target_link_libraries` should just be able to refer to the `type-library` target instead of its internal `.a`.

I tried that first, using `type-registry`, but it wasn't working.

> That said, it also seems like libia2 shouldn't actually need to depend on the type registry library, should it? Generated call-gates may, if the rewriter inserts type-registry based validation, but it seems like it'll just be dead code otherwise. Presumably just tests that depend on the type registry should have cmake dependencies declared on it.

Wouldn't they just get dead-code eliminated if not used?  I thought it'd be simpler that way.

fw-immunant:
[Amazing](https://cmake.org/cmake/help/latest/command/add_library.html#id5): if you use `IMPORTED` with `add_library`, the target is only visible in the current directory and below, unless you add the `GLOBAL` keyword to the `add_library` call.

fw-immunant:
> > That said, it also seems like libia2 shouldn't actually need to depend on the type registry library, should it? Generated call-gates may, if the rewriter inserts type-registry based validation, but it seems like it'll just be dead code otherwise. Presumably just tests that depend on the type registry should have cmake dependencies declared on it.
> 
> Wouldn't they just get dead-code eliminated if not used? I thought it'd be simpler that way.

It's more a question of semantic organization than functional correctness. The type-registry is an example of one kind of data validation that a project using IA2 might want to do; it isn't part of what IA2 itself needs to do for all consumers.

fw-immunant:
ARM CI seems to fail:
>    Compiling type-registry v0.1.0 (/home/runner/work/IA2-Phase2/IA2-Phase2/runtime/type-registry)
> error[E0463]: can't find crate for `std`
>   |
>   = note: the `aarch64-unknown-linux-gnu` target may not be installed
>   = help: consider downloading the target with `rustup target add aarch64-unknown-linux-gnu`
> 
> For more information about this error, try `rustc --explain E0463`.
> error: could not compile `type-registry` (lib) due to 1 previous error

fw-immunant (approved):
LGTM. Per discussion the plan is to move the dependency from libia2 to the tests that use it so as to avoid growing the tcb for IA2 users that don't use the type registry for data validation.

PR #522, opened 2025-03-09
rewriter: thread through `Context` type and move `TypeId`/`TypeInfo` to `TypeInfoInterner`

kkysen:
Some refactoring on top of #514 that sets things up nicely for the type registry calls.


fw-immunant (changes requested):
The `Context` is a bit confusing to me. What is this the context *of*, i.e. to what it it scoped? It seems we only create one `Context`, in `main`. There ends up being a lot of collateral churn and a confusing parameter passed down through a lot of functions that don't need any additional information to compute their results (e.g. all the ABI-determination code). As ABI determination is already hairy, I want to keep that code as simple as possible. Could the API tracking logic be kept more separate, and maybe just access the `TypeInfoInterner` through a global? I'm not sure what work the notion of a `Context` is actually doing here.

fw-immunant:
I guess what I'm reaching for is that `Context` seems to really be about describing the types of things, and it would be nicest to separate that from code that just cares about ABI. For example, the definitions of `Param` and so on in `CAbi.h` don't really relate to the C ABI at all.

kkysen:
`Context` is supposed to be all of the various globals we have scattered everywhere that need to be passed mostly from `main` to various rewriter functions.  It's meant to simplify the parameter passing by making globals explicit, and there were also some situations where just adding yet more globals wasn't working, although it was many months ago so I forget exactly where that was.

fw-immunant (approved):
Approving so we can get this merged; I'll follow up with some refactoring when there aren't so many stacked PRs in flight.

PR #521, opened 2025-03-07
tests: add more `type_confusion` tests for `Dav1dContext`

kkysen:

PR #524, opened 2025-03-09
rewriter: add `IA2_{CON,DE}STRUCTOR` attribute macros, emit `ia2_type_registry_*` calls, and enable `type_confusion` tests

kkysen:
Note that this PR includes #521 (`git cherry-pick`ed) since the branches are orthogonal (#521 points directly to `main`), so once #521 is merged, those commits can be rebased out of this PR.


tools/rewriter/GenCallAsm.cpp
```diff
@@ -915,6 +915,24 @@ static void emit_restore_args(AsmWriter &aw, Arch arch, bool pop) {
...
+
+static void emit_revert_align_stack(AsmWriter &aw, Arch arch) {
+  add_comment_line(aw, "Revert align stack");
+  if (arch == Arch::X86) {
+    add_asm_line(aw, "addq $8, %rsp");
+  } else if (arch == Arch::Aarch64) {
+    // TODO
+  }
+}
+
```
fw-immunant: These should have some documentation or clearer names to indicate that they don't *always* align the stack, they just adjust it by 8. The calling context needs to know the current stack misalignment to know whether calling `emit_align_stack` will align the stack or not.

kkysen: Fixed in ba28321.  How does that look?  I'm not really sure what a great name would be, though.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -1129,6 +1151,60 @@ std::string emit_asm_wrapper(
...
+      // `ia2_type_registry_*` arg 0 is a ptr of type `T*` from arg i.
+      add_asm_line(aw, llvm::formatv("movq {0}(%rsp), %{1}", 8 + 8 * (x86_int_param_reg_order.size() - 1 - i), x86_int_param_reg_order[0]));
+      // `ia2_type_registry_*` arg 1 is the `TypeId`.
+      add_asm_line(aw, llvm::formatv("movq ${0}, %{1}", type.id, x86_int_param_reg_order[1]));
+      emit_direct_call(aw, arch, ia2_type_registry_fn_name);
+    }
+    emit_revert_align_stack_8_to_16(aw, arch);
+    emit_restore_args(aw, arch, /* pop */ true);
+  }
+
```
fw-immunant, line 1207: Just because this function is already so long, it would be nice to factor this out into a separate `emit_type_registry_callback` (or similar) method. Doesn't have to be done in this PR.

fw-immunant (approved):
Code looks OK (one comment inline), but tests seem to be failing; fix them before merging.

tools/rewriter/GenCallAsm.cpp
```diff
@@ -954,6 +976,62 @@ static void emit_return(AsmWriter &aw) {
...
+    emit_align_stack_8_to_16(aw, arch);
+    for (const auto i : arg_indices_to_check) {
+      const auto &arg = sig.api.args[i];
+      const auto &type = ctx.types.get(arg.type);
+      add_comment_line(aw, llvm::formatv("{0}ing type {1}, type ID {2}, arg #{3}", ia2_type_registry_fn_kind, type.canonical_name, type.id, i));
+      // `ia2_type_registry_*` arg 0 is a ptr of type `T*` from arg i.
+      add_asm_line(aw, llvm::formatv("movq {0}(%rsp), %{1}", 8 + 8 * (x86_int_param_reg_order.size() - 1 - i), x86_int_param_reg_order[0]));
+      // `ia2_type_registry_*` arg 1 is the `TypeId`.
+      add_asm_line(aw, llvm::formatv("movq ${0}, %{1}", type.id, x86_int_param_reg_order[1]));
+      emit_direct_call(aw, arch, ia2_type_registry_fn_name);
```
kkysen, line 1028: I think the segfault I'm seeing is a compartment violation in the type registry library.  Do I need to declare the globals in the type registry as `IA2_SHARED_DATA`?

kkysen: So it seems #578 caused this.  At least, previously (and with #578 reverted) the tests passed locally and in most CI runs (I think one or two combinations still failed), but now it fails in all of them.

fw-immunant: Logically speaking, we shouldn't be linking libia2 into call gates, so I'd rather we not revert #578 wholesale. But the type registry itself may need to be linked into call-gates that use it, because logically it is part of the compartmentalized library (we want to avoid other compartments corrupting this one's idea of which objects exist and what types they have). This seems related to #526--essentially, validation of inputs to a compartment is up to an individual compartment rather than IA2 itself, and some compartments might use the type registry while others might have other means of data validation.

fw-immunant:
CI is failing with the tracer because donna has `0xfffffff0` in `/sys/kernel/debug/x86/init_pkru`. When we reboot or set it back to `0x55555554` CI here should pass.

runtime/type-registry/src/lib.rs
```diff
@@ -1,16 +1,134 @@
-use std::collections::HashMap;
-use std::fmt;
-use std::fmt::Debug;
-use std::fmt::Display;
-use std::fmt::Formatter;
-use std::ptr;
-use std::sync::LazyLock;
-use std::sync::RwLock;
+#![no_std]
+/* use the C allocator; don't use libstd */
```
kkysen, line 2: Why do we need to avoid `std`?  It doesn't seem smart to instead write our own `Mutex`; there's a lot we can do incorrectly.

runtime/type-registry/src/lib.rs
```diff
@@ -44,7 +162,9 @@ impl Debug for TypeId {
     }
 }
 
-static GLOBAL_TYPE_REGISTRY: LazyLock<TypeRegistry> = LazyLock::new(TypeRegistry::default);
+static GLOBAL_TYPE_REGISTRY: TypeRegistry = TypeRegistry {
+    map: Mutex::new(BTreeMap::new()),
```
kkysen, line 166: We can build a `const`-compatible `HashMap` now (`HashMap::with_hasher(BuildHasherDefault::new())`).

runtime/type-registry/src/lib.rs
```diff
@@ -1,16 +1,134 @@
...
+
+struct StdErrWriter;
+impl alloc::fmt::Write for StdErrWriter {
+    fn write_str(&mut self, s: &str) -> Result<(), core::fmt::Error> {
+        unsafe extern "C" {
+            unsafe fn write(fd: i32, buf: *const u8, len: usize);
+        }
+        unsafe {
+            write(2, s.as_ptr(), s.len());
+            write(2, "\n".as_ptr(), 1);
```
kkysen, line 17: Why are we writing a newline here?

runtime/type-registry/src/lib.rs
kkysen: If we're going to write our own `Mutex`, we should use `no_std` [`lock_api`](https://docs.rs/lock_api) to write it.

tests/type_confusion/CMakeLists.txt
```diff
@@ -12,3 +12,6 @@ define_test(
     CRITERION_TEST
     WITHOUT_SANDBOX # TODO remove once tracer bug (#488) is fixed
 )
+
+# This test uses the type registry, so link it into the call gates
+target_link_libraries(type_confusion_call_gates PRIVATE type-registry)
```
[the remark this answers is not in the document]
kkysen: That branch died on donna 😢.

runtime/type-registry/src/lib.rs
fw-immunant: Good catch, the newline should be in the caller (where we can just use `writeln!` instead of `write!`). Fixed.

PR #529, opened 2025-03-11
tests/type_confusion: add `dav1d_{alloc,free}` instead of `IA2_SHARED_DATA` and make `Dav1dContext` opaque

kkysen:
`donna` is down, so I'm testing this in CI.

@fw-immunant, is this what you meant?

### DIFF
--- a/tests/type_confusion/dav1d.c
+++ b/tests/type_confusion/dav1d.c
@@ -8,6 +8,15 @@ RUN: cat dav1d_call_gates_1.ld | FileCheck --check-prefix=LINKARGS %s
 #define IA2_COMPARTMENT 2
 #include <ia2_compartment_init.inc>
 
+/// Allocate memory in dav1d's compartment 2.
+void* dav1d_alloc(const size_t size) {
+  return malloc(size);
+}
+
+void dav1d_free(void* const memory) {
+  free(memory);
+}
+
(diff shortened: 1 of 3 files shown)


fw-immunant (changes requested):
I was suggesting individual wrappers for the dav1d initializer-type functions that would return new, allocated-and-initialized objects and be marked as the constructor for the given type. That said, this also achieves the same memory ownership for the allocated objects, which is the main point: the caller can create objects but can only modify them via the dav1d compartment's API, not by poking at their bytes directly.

kkysen:
> I was suggesting individual wrappers for the dav1d initializer-type functions that would return new, allocated-and-initialized objects and be marked as the constructor for the given type.

Correct me if I'm wrong, but I thought we had settled on a more generic `dav1d_alloc`/`dav1d_free` being simpler than individual wrappers for each type.  Although this was a while ago so I don't remember too much, and I might be remembering wrong.

> That said, this also achieves the same memory ownership for the allocated objects, which is the main point: the caller can create objects but can only modify them via the dav1d compartment's API, not by poking at their bytes directly.

Is this way good then?  Or what changes did you want me to still make?

fw-immunant:
> Is this way good then? Or what changes did you want me to still make?

Given discussions we've had, I think we should probably just merge this approach when tests pass; what I really wanted was going to need something like #485 but that was mostly a matter of aesthetics.

fw-immunant (approved):

Issue #482, opened 2024-12-13
Make `--output-prefix` optional (or remove it entirely)


randomPoison:
Currently the `--output-prefix` flag on the rewriter is required to be specified. I'm not sure why we need that prefix to be specified by the user, it seems like we could use a default prefix of `ia2_` or `wrapper_` (or something similar). If there's a reason why we need to give the user an option to control that prefix, then I think we should make `--output-prefix` optional, though I think the better solution would be to remove it entirely if that wouldn't cause any problems.

kkysen:
I agree.  Having this option doesn't seem very useful, and it also makes referring to these files in documentation, etc. more confusing.  It should also go in `--output-directory` (I forgot if it already does, since I just pass an absolute path).

randomPoison:
It currently puts those files in whatever directory you run the rewriter in. I agree they should go in the output directory, but when I brought that up the other day I think @ayrtonm said that doesn't make sense for some reason (I don't remember what exactly).

PR #531, opened 2025-03-13
Use CLI11 instead of llvm::cl to parse command-line arguments

### LINKED ISSUE #328: Rewriter command-line interface is inscrutable
Some of this we can't easily solve because it comes from libClangTooling's built-in behavior, but I wanted to at least dump my notes as partial documentation:

The `--help` usage message suggests:
```
USAGE: ia2-rewriter [options] <source0> [... <sourceN>]
```

but it also accepts, after sources, a `--` followed by arbitrary compiler flags, where if `--` is present, the built-in flags (e.g. where the compiler would learn where to find `stddef.h`) seem to be ignored and `-I /usr/lib/clang/${CLANG_MAJOR}/include` must be passed to find built-in headers. We do this in our ffmpeg/libusb tests with `-I %resource_dir`.

In some of my tests running the rewriter in this way clears the content of all files in `runtime/libia2/include`, which is highly unexpected.

This mode also overlaps with the possible use cases for `--extra-arg`, and it isn't clear to a user which to prefer or how they interact.

In addition, the `-p` flag is documented as `-p <string>                 - Build path`: this does not make it all clear that this is a path to a directory within in which `compile_commands.json` will be looked for.

fw-immunant:
* Fixes #328.

This fixes the weird interaction of `--extra-args`, actual extra arguments (after `--`), and the per-file compiler arguments loaded from `compile_commands.json`. Now only the arguments from `compile_commands.json` are used (adding any arguments passed in `--extra-args`, as we rely on this functionality), and we work around Clang's attempts to "synthesize" new arguments for unknown files.

Separated out from library-only mode work, which adds some of its own command-line options and was infeasible atop our existing pile of buggy complexity.


fw-immunant:
Fixes #328.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -1117,31 +1072,120 @@ std::optional<Pkey> pkey_from_commands(std::function<std::optional<std::vector<C
...
+    ->option_text("<DIR> (REQUIRED)")
+    ->transform(ValidateDirectory)
+    ->required();
+  app.add_option("--root-directory", RootDirectory, "Root directory for input files")
+    ->option_text("<DIR> (REQUIRED)")
+    ->transform(ValidateDirectory)
+    ->required();
+  app.add_option("--output-prefix", OutputPrefix, "Path prefix for output files")
+    ->option_text("<FILENAME-PREFIX> (REQUIRED)")
+    ->required();
```
randomPoison, line 1157: Would it make sense to have a default value for the output prefix? This is just the name used for the various generated `.ld` and `.objcopy` files, right? Seems like we could default that to "wrapper" or "ia2_" or something so that users don't need to specify the flag unless they want to customize the name.

randomPoison (approved):
LGTM, just one small question that doesn't need to block merge.

tools/rewriter/SourceRewriter.cpp
fw-immunant: I think so, that's the gist of #482, but for this PR I wanted to minimize semantic changes.

PR #532, opened 2025-03-13
Try to speed up Rust setup in CI

fw-immunant:


.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         ia2-tracer: [ON, OFF]
     steps:
       - name: Setup Rust
-        uses: actions-rust-lang/setup-rust-toolchain@v1
+        run: rustup toolchain install stable-x86_64-unknown-linux-gnu
```
kkysen, line 24: Does it make sense to do this same thing for the aarch64 one?

fw-immunant: It doesn't really matter as that test can run in parallel with the ones on our runner.

kkysen: Yeah but it's helpful to be consistent.

PR #533, opened 2025-03-14
Minor rewriter fixes and cleanups

fw-immunant:
Split off of library-only mode, sending PR to keep that branch closer to `main`. The only real semantic change here is a fix for a bug in type alias handling.

### DIFF
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -513,16 +513,14 @@ public:
 
     std::string new_expr =
         "IA2_CALL("s + old_callee.str() + ", " + mangled_ty;
-    
+
     for (auto const &arg : fn_ptr_call->arguments()) {
-      new_expr += ", " + clang::Lexer::getSourceText(
-          clang::CharSourceRange::getTokenRange(arg->getSourceRange()), sm,
-          ctxt.getLangOpts())
-                     .str();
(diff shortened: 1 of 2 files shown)


tools/rewriter/SourceRewriter.cpp
```diff
@@ -513,16 +513,14 @@ class FnPtrCall : public RefactoringCallback {
 
     std::string new_expr =
         "IA2_CALL("s + old_callee.str() + ", " + mangled_ty;
-    
+
```
kkysen, line 516: It'd be nice to merge in #518 first so we don't have these same formatting changes in a bunch of different PRs (I've seen this one, too, in a bunch of mine).

tools/rewriter/SourceRewriter.cpp
```diff
@@ -597,7 +595,6 @@ class FnPtrExpr : public RefactoringCallback {
     auto *fn_ptr_expr = result.Nodes.getNodeAs<clang::DeclRefExpr>("fnPtrExpr");
     assert(fn_ptr_expr != nullptr);
 
-    auto annotation = fn_ptr_expr->getDecl()->getAttr<clang::AnnotateAttr>();
```
kkysen, line 600: This is dead code, right?  I've seen a warning for it for a while

kkysen (approved):

PR #534, opened 2025-04-07
Build changes for ARM64

ayrtonm:


fw-immunant:
The x86 switch of PartitionAlloc to libc++ breaks x86 CI. It might just need to be installed.

fw-immunant:
Should this be changing our submodule to point at https://github.com/immunant/IA2-glibc? Testing this locally I see the glibc configure failing at its gcc version check step:
`checking if llvm-project/build/bin/clang-19 -target aarch64-linux-gnu -ffixed-x18 -march=armv8.5-a+memtag --rtlib=compiler-rt -O3 -B/usr/aarch64-linux-gnu/bin/ is sufficient to build libc... no`

runtime/libia2/CMakeLists.txt
```diff
@@ -74,7 +74,7 @@ if (LIBIA2_AARCH64)
                 # This is only for the heap tunable
                 "--enable-memory-tagging"
             BUILD_COMMAND make -j8
-            INSTALL_COMMAND make install
+            INSTALL_COMMAND cmake -E echo "skip install"
```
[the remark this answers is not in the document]
fw-immunant: From what I can tell, this doesn't get skipped before this PR but also doesn't need to run. I'll add a comment to that effect.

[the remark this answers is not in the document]
fw-immunant: Ok, I now fully understand this: because of our `--prefix` in the `configure` arguments, this was installing to `${CMAKE_BINARY_DIR}/external/glibc/sysroot` (not system dirs). We were then copying libc out of that sysroot. But the install command was technically failing because the built ldconfig isn't compatible with the host architecture. So I'll instead grab the libc.so and ld-linux.so out of the build directories rather than the sysroot, and skip the `make install` step.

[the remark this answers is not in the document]
fw-immunant: Saying "fully understand" was setting myself up for failure. The above does, AFAICT, capure the extent of the *build-time* interactions with glibc's `make install`, but I failed to notice that the install step also placed `libc.so.6` and `ld-linux-aarch64.so.1` at the location required for dynamically finding them when running our binaries under QEMU. I've reinstated the install step and CI is now passing.

.github/workflows/ci.yml
```diff
@@ -203,11 +208,13 @@ jobs:
             -DCMAKE_OBJCOPY=/usr/bin/llvm-objcopy-15 \
             -DCMAKE_C_FLAGS="$cross_flags" \
             -DCMAKE_CXX_FLAGS="$cross_flags" \
-            -DCMAKE_EXE_LINKER_FLAGS="$cross_link_flags -lm -Wl,-rpath-link,${rtlibs_dir}/lib" \
-            -DCMAKE_SHARED_LINKER_FLAGS="$cross_link_flags -Wl,-rpath,${rtlibs_dir}/lib" \
```
ayrtonm, line 207: Was switching -Wl,-rpath and -Wl,-rpath-link in these two lines intentional?

cmake/qemu-command.cmake
```diff
@@ -12,7 +12,7 @@ else()
   endif()
   set(QEMU_COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR}
     "-one-insn-per-tb"
-    "-L" "${CMAKE_BINARY_DIR}/external/glibc/sysroot/usr/"
```
ayrtonm, line 15: Does the sysroot not include the headers anymore? Why are we removing the usr suffix?

runtime/libia2/CMakeLists.txt
```diff
@@ -55,15 +55,18 @@ target_link_options(libia2 INTERFACE
...
 if (LIBIA2_AARCH64)
+    set(SYSROOT_BIN_DIR ${CMAKE_BINARY_DIR}/external/glibc/sysroot)
+    set(LIBC_PATH ${SYSROOT_BIN_DIR}/lib/libc.so.6)
+    set(LD_LINUX_PATH ${SYSROOT_BIN_DIR}/lib/ld-linux-aarch64.so.1)
     if (LIBIA2_REBUILD_GLIBC)
         ExternalProject_Add(glibc
             SOURCE_DIR ${CMAKE_SOURCE_DIR}/external/glibc
             BINARY_DIR ${CMAKE_BINARY_DIR}/external/glibc
             CONFIGURE_COMMAND ${CMAKE_COMMAND} -E env
                 # ffixed-x18 has to be set here instead of CFLAGS for it to apply to all files
```
ayrtonm, line 66: nit: s/ffixed-x18/compiler flags in the comment

fw-immunant: Will fix in a follow-up.

cmake/qemu-command.cmake
ayrtonm: That seems fine then

PR #535, opened 2025-04-10
runtime: fix zeroed PKRU in init_pkey

fw-immunant:
we meant to zero the compartment; zeroing the PKRU gives access to *everything*


fw-immunant:
Seems like this breaks the tracer. I'll look into it when I have a chance.

ayrtonm:
I don't think we have an issue explicitly for moving code out of headers but maybe we should do that and try to untangle the initialization stuff before spending time debugging this.

ayrtonm:
`init_pkey` was replaced by calls to `dl_iterate_phdr(protect_pages, ...)` without `wrpkru`s so I think we can close this if we choose to go with #561.

ayrtonm:
Superseded by #561. That PR does these initial `pkey_mprotect`s without changing the PKRU which removes the need for the wrpkru in the first place.

PR #537, opened 2025-04-10
libia2 fixes for C++ compat

fw-immunant:
Developed working on FF compartmentalization. We don't handle C++ constructs crossing the compartment boundary but might need to include our headers from C++ sources.

### DIFF
--- a/runtime/libia2/include/ia2_compartment_init.inc
+++ b/runtime/libia2/include/ia2_compartment_init.inc
@@ -44,7 +44,7 @@ extern int ia2_n_pkeys_to_alloc;
 extern char __start_ia2_shared_data __attribute__((visibility("hidden"))),
     __stop_ia2_shared_data __attribute__((visibility("hidden")));
 
-void ia2_set_up_tags(int *n_to_alloc);
+IA2_EXTERN_C void ia2_set_up_tags(int *n_to_alloc);
 __attribute__((constructor)) static void COMPARTMENT_IDENT(init_pkey)() {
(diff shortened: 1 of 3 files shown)


runtime/libia2/include/ia2.h
```diff
@@ -131,8 +131,11 @@
 #define IA2_AS_PTR(opaque) (opaque).ptr
 
 /// Get an IA2 opaque function pointer for the wrapped version of `func`
-#define IA2_FN(func)                                                           \
-  (typeof(__ia2_##func)) { (void *)&__ia2_##func }
+#ifdef __cplusplus
+#define IA2_FN(func) reinterpret_cast<typeof &func>((void *)&((__ia2_##func).ptr))
```
ayrtonm, line 135: I don't think `IA2_FN` should do this just because it's C++. In C it expands to a struct while in C++ it's expanding to a function pointer. It made sense to do this for Firefox because of the way we're running the rewriter on a subset of the sources, but I think this macro should expand to the same thing in both languages. We could just throw out the opaque function pointers and have both expand to the same thing but that should be a separate change.

ayrtonm (commented):
LGTM for everything except the `IA2_FN` change.

runtime/libia2/include/ia2.h
fw-immunant: That's a good point, this is more of a library-only-mode change (or FF-specific) that has to be expressed differently in C++ than in C.

fw-immunant: Plus that change wasn't even described by the commit message. Removed.

PR #539, opened 2025-04-12
Forward-port benchmarks

fw-immunant:
@rinon wrote some nice microbenchmarks and a larger video-decoding macrobenchmark. There's no reason not to merge these to keep them from bitrotting. This PR brings them into the build but doesn't run benchmarks as part of CI; we can decide if we want to do that later.


fw-immunant:
I'm going to merge this as the only code I touched is a couple build/CI tweaks and the rest of the code written by Stephen looks good to me.

PR #555, opened 2025-05-12
permissive_mode: label (main thread) ia2 stacks

kkysen:
* Part of #405.
* Split out from #421.

### DIFF
--- a/runtime/libia2/include/ia2.h
+++ b/runtime/libia2/include/ia2.h
@@ -161,6 +161,8 @@
 #define IA2_CAST(func, ty) (ty) { (void *)IA2_FN_ADDR(func) }
 #endif // !IA2_ENABLE
 
+#define IA2_MAX_COMPARTMENTS 16
+
 /// Convert a compartment pkey to a PKRU register value
 #define PKRU(pkey) (~((3U << (2 * pkey)) | 3))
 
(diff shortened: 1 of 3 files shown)


runtime/libia2/include/ia2.h
```diff
@@ -161,6 +161,8 @@
 #define IA2_CAST(func, ty) (ty) { (void *)IA2_FN_ADDR(func) }
 #endif // !IA2_ENABLE
 
+#define MAX_PKEYS 16
```
ayrtonm, line 164: We should namespace things in this header (i.e. IA2_MAX_PKEYS). Also that's x86-specific so maybe `IA2_MAX_COMPARTMENTS` is better.

ayrtonm (approved):
minor nit, but otherwise looks good

runtime/libia2/include/permissive_mode.h
```diff
@@ -465,23 +465,78 @@ __attribute__((constructor)) void permissive_mode_init(void) {
...
+    // Parse `/proc/self/maps` line.
+    uintptr_t start_addr = 0;
+    uintptr_t end_addr = 0;
+    char perms[4] = {0};
+    uintptr_t offset = 0;
+    unsigned int dev_major = 0;
+    unsigned int dev_minor = 0;
+    ino_t inode = 0;
+    int path_index = 0;
+    sscanf(line, "%lx-%lx %4c %lx %x:%x %lu %n", &start_addr, &end_addr, perms, &offset, &dev_major, &dev_minor, &inode, &path_index);
```
fw-immunant, line 506: We should warn if this fails rather than printing bogus info.

kkysen: Fixed in #579.

PR #556, opened 2025-05-12
permissive_mode: label the (main thread) TLS regions

kkysen:
* Part of #405.
* Split out from #421.

### DIFF
--- a/runtime/libia2/include/permissive_mode.h
+++ b/runtime/libia2/include/permissive_mode.h
@@ -472,6 +472,10 @@ void __real_free(void *ptr);
 
 extern uintptr_t ia2_stack_addrs[IA2_MAX_COMPARTMENTS];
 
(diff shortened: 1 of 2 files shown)


runtime/libia2/include/permissive_mode.h
```diff
@@ -514,11 +518,20 @@ void log_memory_map(void) {
       fprintf(log, "%s", path);
     } else {
       // No path, try to identify it.
+      if (start_addr == ia2_tls_addr_compartment1_first) {
+        fprintf(log, "[tls:tid ?:compartment 1, first]");
+      } else if (start_addr == ia2_tls_addr_compartment1_second) {
+        fprintf(log, "[tls:tid ?:compartment 1, second]");
```
ayrtonm, line 524: I think it's fine to leave out the `first`/`second` from the labels and just have both parts the same.

ayrtonm (approved):

Issue #546, opened 2025-05-02
Deallocate compartment stacks when a thread terminates


ayrtonm:
This didn't explicitly cause issues in Firefox but since it spins up many threads, it likely closes some as well. When a thread terminates we currently leak the memory mmap'ed for that thread's stacks (one stack per compartment). We should clean up this memory instead.

fw-immunant:
It looks like we have a few options for hooking pthread termination here:
- manually wrapping `pthread_exit` (doesn't catch canceled threads or `return`)
- `pthread_cleanup_push` (which runs on `pthread_cancel` and `pthread_exit` but not `return` from the thread body
- adding a dtor via `pthread_key_create()` (which runs on `pthread_cancel`, `pthread_exit`, and `return` from the thread body)

fw-immunant:
See also #303 which discusses additional handling we may need for `pthread_key_create` destructors. I don't know whether that's an argument for (based on coupling) or against (to avoid interacting with another component having potentially conflicting requirements) using `pthread_key_create` here.

Issue #249, opened 2023-07-26
Thread-local shared data


rinon:
I don't think we currently support the `IA2_SHARED` annotation on thread-local data. We just need some way to make sure this data is allocated to compartment 0, right?

We currently need thread-local shared data for at least signal alternate stacks, although for that we need to be careful that the _default_ pkru value can read/write the stack, since the kernel resets the PKRU to a default value before invoking the signal handler.

fw-immunant:
> we need to be careful that the _default_ pkru value can read/write the stack, since the kernel resets the PKRU to a default value before invoking the signal handler.

I.e., 0x55555554?

rinon:
Yes, `0x55555554` although I believe there is some kernel support for changing the default? https://elixir.bootlin.com/linux/latest/source/arch/x86/mm/pkeys.c#L168

rinon:
init_pkru requires kernel 4.9 fwiw

fw-immunant:
We discussed this a fair bit in a call. Takeaways:
- On the thread-local shared data side:
  - `ia2_stackptr_0` is basically our prototype for shared TLS data. Presently, we `dlsym()` it and avoid protecting its page in `protect_tls_data`. This [gives us a dependency on calling `dlsym()` in protect_tls_data`](https://github.com/immunant/IA2-Phase2/pull/229/files#r1273823909), which we don't want. It also doesn't scale to other shared thread-locals.
  - We can't simply put shared TLS data in a different ELF section, because we really need glibc to do its setup magic which it only does on the TLS sections it knows about.
  - I considered some kind of macro scheme using asm symbol names (`type foo __asm__("ia2_shared_tls_foo");`) so that we could at least detect shared TLS data by name by iterating+inspecting symbols. But I think this would break cross-library references, and iterating+inspecting symbols is likely to be very slow, and probably still has a libdl dependency.
  - If we can opt into the general dynamic TLS model (see tls.pdf) per-variable, that may offer a solution. This model involves calling `__tls_get_addr`, which is "basically" implemented as `__tls_get_addr(module, offset) = dtv[thread_id][module] + offset`. If we can guarantee that no non-shared TLS variables use the general dynamic TLS model, and also opt shared TLS variables into it, we can then overwrite `dtv[thread_id][module]` to point at unprotected, shared storage that we allocate at thread startup.

- On the signal-handling side:
[...]

PR #557, opened 2025-05-12
permissive_mode: label stacks and TLS regions in all threads

### LINKED ISSUE #405: Add mmap region descriptions to permissive mode logs

kkysen:
* Fixes #405.

<details>
  <summary>An example <code>mpk_log_*</code>:</summary>

```
  start addr-end addr     perms offset  path
1987fffff000-198c00000000 ---p 00000000
3a0c00000000-3a1000000000 ---p 00000000
3efe50000000-3efe50001000 ---p 00000000
3efe50001000-3efe50002000 rw-p 00000000
3efe50002000-3efe50004000 ---p 00000000
3efe50004000-3efe50008000 rw-p 00000000
3efe50008000-3efe60000000 ---p 00000000
5ea323a3f000-5ea323a49000 r--p 00000000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5ea323a49000-5ea323a7e000 r-xp 0000a000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5ea323a7e000-5ea323a93000 r--p 0003f000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5ea323a94000-5ea323a95000 r--p 00054000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5ea323a95000-5ea323a9a000 rw-p 00055000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5ea323a9a000-5ea323c20000 rw-p 0005a000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5ea323c20000-5ea324562000 rw-p 00000000
5ea330893000-5ea3308b4000 rw-p 00000000 [heap]
70ea42000000-70ea42400000 rw-p 00000000 [stack:tid 2761199:compartment 0]
70ea42400000-70ea42800000 rw-p 00000000 [stack:tid 2761199:compartment 1]
70ea42800000-70ea42c00000 rw-p 00000000 [stack:tid 2761198:compartment 0]
70ea42c00000-70ea43000000 rw-p 00000000 [stack:tid 2761198:compartment 1]
70ea43000000-70ea43001000 ---p 00000000
70ea43001000-70ea437fc000 rw-p 00000000
70ea437fc000-70ea437fe000 rw-p 00000000 [tls:tid 2761199:compartment 1]
70ea437fe000-70ea43801000 rw-p 00000000
70ea43a00000-70ea43e00000 rw-p 00000000 [stack:tid 2761197:compartment 0]
70ea43e00000-70ea43e01000 ---p 00000000
70ea43e01000-70ea445fc000 rw-p 00000000
70ea445fc000-70ea445fe000 rw-p 00000000 [tls:tid 2761198:compartment 1]
70ea445fe000-70ea44601000 rw-p 00000000
70ea44800000-70ea44c00000 rw-p 00000000 [stack:tid 2761197:compartment 1]
70ea44c00000-70ea45000000 rw-p 00000000 [stack:tid 2761196:compartment 0]
70ea45000000-70ea45400000 rw-p 00000000 [stack:tid 2761196:compartment 1]
70ea45400000-70ea45401000 ---p 00000000
70ea45401000-70ea45bfc000 rw-p 00000000
70ea45bfc000-70ea45bfe000 rw-p 00000000 [tls:tid 2761197:compartment 1]
70ea45bfe000-70ea45c01000 rw-p 00000000
70ea45e00000-70ea46200000 rw-p 00000000 [stack:tid 2761195:compartment 0]
70ea46200000-70ea46600000 rw-p 00000000 [stack:tid 2761195:compartment 1]
70ea46600000-70ea46601000 ---p 00000000
70ea46601000-70ea46dfc000 rw-p 00000000
70ea46dfc000-70ea46dfe000 rw-p 00000000 [tls:tid 2761196:compartment 1]
70ea46dfe000-70ea46e01000 rw-p 00000000
70ea47000000-70ea47400000 rw-p 00000000 [stack:tid 2761194:compartment 0]
70ea47400000-70ea47800000 rw-p 00000000 [stack:tid 2761194:compartment 1]
70ea47800000-70ea47801000 ---p 00000000
70ea47801000-70ea47ffc000 rw-p 00000000
70ea47ffc000-70ea47ffe000 rw-p 00000000 [tls:tid 2761195:compartment 1]
70ea47ffe000-70ea48001000 rw-p 00000000
70ea48200000-70ea48600000 rw-p 00000000 [stack:tid 2761193:compartment 0]
70ea48600000-70ea48a00000 rw-p 00000000 [stack:tid 2761193:compartment 1]
70ea48a00000-70ea48a01000 ---p 00000000
70ea48a01000-70ea491fc000 rw-p 00000000
70ea491fc000-70ea491fe000 rw-p 00000000 [tls:tid 2761194:compartment 1]
70ea491fe000-70ea49201000 rw-p 00000000
70ea49400000-70ea49c00000 rw-p 00000000 [stack:tid 2761192:compartment 0]
70ea49c00000-70ea4a000000 rw-p 00000000 [stack:tid 2761192:compartment 1]
70ea4a000000-70ea4a400000 rw-p 00000000 [stack:tid 2761190:compartment 0]
70ea4a400000-70ea4a401000 ---p 00000000
70ea4a401000-70ea4abfc000 rw-p 00000000
70ea4abfc000-70ea4abfe000 rw-p 00000000 [tls:tid 2761193:compartment 1]
70ea4abfe000-70ea4ac01000 rw-p 00000000
70ea4ae00000-70ea4b200000 rw-p 00000000 [stack:tid 2761191:compartment 1]
70ea4b200000-70ea4b201000 ---p 00000000
70ea4b201000-70ea4b9fc000 rw-p 00000000
70ea4b9fc000-70ea4b9fe000 rw-p 00000000 [tls:tid 2761192:compartment 1]
70ea4b9fe000-70ea4ba01000 rw-p 00000000
70ea4bc00000-70ea4c000000 rw-p 00000000 [stack:tid 2761190:compartment 1]
70ea4c000000-70ea4c021000 rw-p 00000000
70ea4c021000-70ea50000000 ---p 00000000
70ea50400000-70ea50401000 ---p 00000000
70ea50401000-70ea50bfc000 rw-p 00000000
70ea50bfc000-70ea50bfe000 rw-p 00000000 [tls:tid 2761191:compartment 1]
70ea50bfe000-70ea50c01000 rw-p 00000000
70ea50e00000-70ea50e01000 ---p 00000000
70ea50e01000-70ea515fc000 rw-p 00000000
70ea515fc000-70ea515fe000 rw-p 00000000 [tls:tid 2761190:compartment 1]
70ea515fe000-70ea51601000 rw-p 00000000
70ea51800000-70ea51c00000 rw-p 00000000 [stack:tid 2761184:compartment 0]
70ea51c00000-70ea52000000 rw-p 00000000 [stack:tid 2761184:compartment 1]
70ea52000000-70ea52001000 ---p 00000000
70ea52001000-70ea52801000 rw-p 00000000
70ea52a00000-70ea52b00000 rw-p 00000000
70ea52c00000-70ea52d00000 rw-p 00000000
70ea52e00000-70ea52e9a000 r--p 00000000 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30
70ea52e9a000-70ea52fab000 r-xp 0009a000 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30
70ea52fab000-70ea5301a000 r--p 001ab000 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30
70ea5301a000-70ea5301b000 ---p 0021a000 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30
70ea5301b000-70ea53026000 r--p 0021a000 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30
70ea53026000-70ea53029000 rw-p 00225000 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30
70ea53029000-70ea5302c000 rw-p 00000000
70ea53200000-70ea53228000 r--p 00000000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/libia2/libc.so.6
70ea53228000-70ea533bd000 r-xp 00028000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/libia2/libc.so.6
70ea533bd000-70ea53415000 r--p 001bd000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/libia2/libc.so.6
70ea53415000-70ea53416000 ---p 00215000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/libia2/libc.so.6
70ea53416000-70ea5341a000 r--p 00215000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/libia2/libc.so.6
70ea5341a000-70ea5341c000 rw-p 00219000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/libia2/libc.so.6
70ea5341c000-70ea53429000 rw-p 00000000
70ea53441000-70ea53464000 rw-p 00000000
70ea53464000-70ea53466000 rw-p 00000000 [tls:tid 2761184:compartment 1]
70ea53466000-70ea5346d000 rw-p 00000000
70ea5346d000-70ea53470000 r--p 00000000 /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
70ea53470000-70ea53487000 r-xp 00003000 /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
70ea53487000-70ea5348b000 r--p 0001a000 /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
70ea5348b000-70ea5348c000 r--p 0001d000 /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
70ea5348c000-70ea5348d000 rw-p 0001e000 /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
70ea5348d000-70ea5348f000 rw-p 00000000
70ea5348f000-70ea53492000 r--p 00000000 /usr/lib/x86_64-linux-gnu/libresolv.so.2
70ea53492000-70ea5349c000 r-xp 00003000 /usr/lib/x86_64-linux-gnu/libresolv.so.2
70ea5349c000-70ea5349f000 r--p 0000d000 /usr/lib/x86_64-linux-gnu/libresolv.so.2
70ea5349f000-70ea534a0000 r--p 0000f000 /usr/lib/x86_64-linux-gnu/libresolv.so.2
70ea534a0000-70ea534a1000 rw-p 00010000 /usr/lib/x86_64-linux-gnu/libresolv.so.2
70ea534a1000-70ea534a3000 rw-p 00000000
70ea534a3000-70ea534b1000 r--p 00000000 /usr/lib/x86_64-linux-gnu/libm.so.6
70ea534b1000-70ea5352d000 r-xp 0000e000 /usr/lib/x86_64-linux-gnu/libm.so.6
70ea5352d000-70ea53588000 r--p 0008a000 /usr/lib/x86_64-linux-gnu/libm.so.6
70ea53588000-70ea53589000 r--p 000e4000 /usr/lib/x86_64-linux-gnu/libm.so.6
70ea53589000-70ea5358a000 rw-p 000e5000 /usr/lib/x86_64-linux-gnu/libm.so.6
70ea53593000-70ea535a8000 rw-p 00000000
70ea535a8000-70ea535a9000 r--p 00000000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/libpermissive_mode_call_gates.so
70ea535a9000-70ea535aa000 r-xp 00001000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/libpermissive_mode_call_gates.so
70ea535aa000-70ea535ab000 r--p 00002000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/libpermissive_mode_call_gates.so
70ea535ab000-70ea535ac000 r--p 00002000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/libpermissive_mode_call_gates.so
70ea535ac000-70ea535ad000 rw-p 00003000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/libpermissive_mode_call_gates.so
70ea535ad000-70ea535ae000 rw-p 00000000
70ea535ae000-70ea53615000 r--p 00000000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/partition-alloc/libpartition-alloc.so
70ea53615000-70ea53675000 r-xp 00067000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/partition-alloc/libpartition-alloc.so
70ea53675000-70ea5369d000 r--p 000c7000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/partition-alloc/libpartition-alloc.so
70ea5369d000-70ea536a3000 r--p 000ee000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/partition-alloc/libpartition-alloc.so
70ea536a3000-70ea536aa000 rw-p 000f4000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/partition-alloc/libpartition-alloc.so
70ea536aa000-70ea536ac000 rw-p 000fb000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/partition-alloc/libpartition-alloc.so
70ea536ac000-70ea536f3000 rw-p 000fd000 /home/kkysen/work/rust/ia2/build/x86_64/runtime/partition-alloc/libpartition-alloc.so
70ea536f3000-70ea5370f000 rw-p 00000000
70ea5370f000-70ea53710000 rw-p 00000000
70ea53710000-70ea53725000 rw-p 00000000
70ea53725000-70ea53727000 r--p 00000000 /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
70ea53727000-70ea53751000 r-xp 00002000 /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
70ea53751000-70ea5375c000 r--p 0002c000 /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
70ea5375d000-70ea5375f000 r--p 00037000 /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
70ea5375f000-70ea53761000 rw-p 00039000 /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
7ffc6514a000-7ffc6516c000 rw-p 00000000 [stack]
7ffc65184000-7ffc65188000 r--p 00000000 [vvar]
7ffc65188000-7ffc6518a000 r-xp 00000000 [vdso]
ffffffffff600000-ffffffffff601000 --xp 00000000 [vsyscall]
```

</details>

### DIFF
--- a/runtime/libia2/ia2.c
+++ b/runtime/libia2/ia2.c
@@ -262,7 +262,7 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
 
   const int pkey = search_args->pkey;
 
-  struct ia2_thread_data *const thread_data = ia2_thread_data_get_current_thread();
+  struct ia2_thread_metadata *const thread_metadata = ia2_thread_metadata_get_current_thread();
 
   // Protect TLS segment.
   for (size_t i = 0; i < info->dlpi_phnum; i++) {
@@ -318,8 +318,8 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
           printf("ia2_mprotect_with_tag failed: %s\n", strerror(errno));
           exit(-1);
         }
-        if (thread_data) {
-          thread_data->tls_addr_compartment1_first = (uintptr_t)start_round_down;
+        if (thread_metadata) {
+          thread_metadata->tls_addr_compartment1_first = (uintptr_t)start_round_down;
         }
       }
       uint64_t after_untrusted_region_start = untrusted_stackptr_addr + 0x1000;
@@ -332,8 +332,8 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
           printf("ia2_mprotect_with_tag failed: %s\n", strerror(errno));
           exit(-1);
         }
-        if (thread_data) {
-          thread_data->tls_addr_compartment1_second = (uintptr_t)after_untrusted_region_start;
+        if (thread_metadata) {
+          thread_metadata->tls_addr_compartment1_second = (uintptr_t)after_untrusted_region_start;
         }
       }
     } else {
@@ -344,8 +344,8 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
         printf("ia2_mprotect_with_tag failed: %s\n", strerror(errno));
         exit(-1);
       }
-      if (thread_data) {
-        thread_data->tls_addrs[pkey] = (uintptr_t)start_round_down;
+      if (thread_metadata) {
+        thread_metadata->tls_addrs[pkey] = (uintptr_t)start_round_down;
       }
     }
   }
--- a/runtime/libia2/ia2.c
+++ b/runtime/libia2/ia2.c
@@ -5,8 +5,9 @@
 #include <stdio.h>
 #include <string.h>
 
-#include "ia2_internal.h"
 #include "ia2.h"
+#include "ia2_internal.h"
+#include "ia2_threads.h"
 
 #if defined(__x86_64__)
 
@@ -246,12 +247,6 @@ static int segment_flags_to_access_flags(Elf64_Word flags) {
       ((flags & PF_R) != 0 ? PROT_READ : 0);
 }
 
-/// The TLS region is split only for the first compartment,
-/// so we need two addresses for just that one.
-uintptr_t ia2_tls_addr_compartment1_first IA2_SHARED_DATA = 0;
-uintptr_t ia2_tls_addr_compartment1_second IA2_SHARED_DATA = 0;
-uintptr_t ia2_tls_addrs[IA2_MAX_COMPARTMENTS] IA2_SHARED_DATA = {0};
-
 int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
   if (!data || !info) {
     printf("Passed invalid args to dl_iterate_phdr callback\n");
@@ -266,6 +261,8 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
   }
 
   const int pkey = search_args->pkey;
+
+  struct ia2_thread_data *const thread_data = ia2_thread_data_get_current_thread();
 
   // Protect TLS segment.
   for (size_t i = 0; i < info->dlpi_phnum; i++) {
@@ -321,7 +318,9 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
           printf("ia2_mprotect_with_tag failed: %s\n", strerror(errno));
           exit(-1);
         }
-        ia2_tls_addr_compartment1_first = (uintptr_t)start_round_down;
+        if (thread_data) {
+          thread_data->tls_addr_compartment1_first = (uintptr_t)start_round_down;
+        }
       }
       uint64_t after_untrusted_region_start = untrusted_stackptr_addr + 0x1000;
       uint64_t after_untrusted_region_len = end - after_untrusted_region_start;
@@ -333,7 +332,9 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
           printf("ia2_mprotect_with_tag failed: %s\n", strerror(errno));
           exit(-1);
         }
-        ia2_tls_addr_compartment1_second = (uintptr_t)after_untrusted_region_start;
+        if (thread_data) {
+          thread_data->tls_addr_compartment1_second = (uintptr_t)after_untrusted_region_start;
+        }
       }
     } else {
       int mprotect_err =
@@ -343,7 +344,9 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
         printf("ia2_mprotect_with_tag failed: %s\n", strerror(errno));
         exit(-1);
       }
-      ia2_tls_addrs[pkey] = (uintptr_t)start_round_down;
+      if (thread_data) {
+        thread_data->tls_addrs[pkey] = (uintptr_t)start_round_down;
+      }
     }
   }
 
--- a/runtime/libia2/include/ia2_threads.h
+++ b/runtime/libia2/include/ia2_threads.h
@@ -15,8 +15,8 @@ struct ia2_thread_metadata {
   /// The TLS region is split only for the first compartment,
   /// so we need two addresses for just that one.
   ///
-  /// Compartment 1's TLS region is split because there is
-  /// a page of unprotected data for `ia2_stackptr_0` plus padding,
+  /// Compartment 1's TLS region is split because there is a page of
+  /// unprotected data for `ia2_stackptr_0` (in compartment 0), plus padding,
   /// as we don't have a general implementation of shared TLS yet,
   /// but `ia2_stackptr_0` is special-cased for now
   /// as it must be stored in TLS and unprotected.
--- a/runtime/libia2/include/ia2_threads.h
+++ b/runtime/libia2/include/ia2_threads.h
@@ -2,7 +2,7 @@
 
 #include "ia2.h"
 
-struct ia2_thread_data {
+struct ia2_thread_metadata {
   /// The addresses of each compartment's stack for this thread.
   ///
   /// This data is shared, so it should not be trusted for use as a pointer,
@@ -37,14 +37,14 @@ struct ia2_addr_location {
   int compartment;
 };
 
-/// Find the `struct ia2_thread_data*` for the current thread,
+/// Find the `struct ia2_thread_metadata*` for the current thread,
 /// adding (but not allocating) one if there isn't one yet.
 /// If there is no memory for more or an error, `NULL` is returned.
 /// This is a purely lookup and/or additive operation,
-/// so the lifetime of the returned `struct ia2_thread_data*` is infinite,
+/// so the lifetime of the returned `struct ia2_thread_metadata*` is infinite,
 /// and since it's thread-specific,
 /// it is thread-safe to read and write.
-struct ia2_thread_data *ia2_thread_data_get_current_thread(void);
+struct ia2_thread_metadata *ia2_thread_metadata_get_current_thread(void);
 
 /// Find the `ia2_addr_location` of `addr`.
 /// If it is not found or there is an error,
--- a/runtime/libia2/include/ia2_threads.h
+++ b/runtime/libia2/include/ia2_threads.h
@@ -26,7 +26,7 @@ struct ia2_addr_location {
   /// `NULL` if unknown.
   const char *name;
 
-  /// The thread ID of the thread this address is on.
+  /// The thread ID of the thread this address belongs to.
   ///
   /// `-1` if unknown.
   pid_t tid;
--- a/runtime/libia2/include/ia2_threads.h
+++ b/runtime/libia2/include/ia2_threads.h
@@ -1,0 +1,43 @@
+#pragma once
+
+#include "ia2.h"
+
+struct ia2_thread_data {
+  /// The addresses of each compartment's stack for this thread.
+  ///
+  /// This data is shared, so it should not be trusted for use as a pointer,
+  /// but it can be used best effort as a provenance-less address.
+  uintptr_t stack_addrs[IA2_MAX_COMPARTMENTS];
+};
+
+struct ia2_addr_location {
+  /// A descriptive name of what this address points to.
+  /// For example, "stack".
+  ///
+  /// `NULL` if unknown.
+  const char *name;
+
+  /// The thread ID of the thread this address is on.
+  ///
+  /// `-1` if unknown.
+  pid_t tid;
+
+  /// The compartment this address is in.
+  ///
+  /// `-1` if unknown.
+  int compartment;
+};
+
+/// Find the `struct ia2_thread_data*` for the current thread,
+/// adding (but not allocating) one if there isn't one yet.
+/// If there is no memory for more or an error, `NULL` is returned.
+/// This is a purely lookup and/or additive operation,
+/// so the lifetime of the returned `struct ia2_thread_data*` is infinite,
+/// and since it's thread-specific,
+/// it is thread-safe to read and write.
+struct ia2_thread_data *ia2_thread_data_get_current_thread(void);
+
+/// Find the `ia2_addr_location` of `addr`.
+/// If it is not found or there is an error,
+/// the fields are set to `NULL` or `-1` depending on the type.
+struct ia2_addr_location ia2_addr_location_find(uintptr_t addr);
--- a/runtime/libia2/include/ia2_threads.h
+++ b/runtime/libia2/include/ia2_threads.h
@@ -2,11 +2,10 @@
 
 #include "ia2.h"
 
+/// The data here is shared, so it should not be trusted for use as a pointer,
+/// but it can be used best effort for non-trusted purposes.
 struct ia2_thread_metadata {
   /// The addresses of each compartment's stack for this thread.
-  ///
-  /// This data is shared, so it should not be trusted for use as a pointer,
-  /// but it can be used best effort for non-trusted purposes.
   uintptr_t stack_addrs[IA2_MAX_COMPARTMENTS];
 
   /// The addresses of each compartment's TLS region for this thread,
--- a/runtime/libia2/include/ia2_threads.h
+++ b/runtime/libia2/include/ia2_threads.h
@@ -14,6 +14,12 @@ struct ia2_thread_metadata {
 
   /// The TLS region is split only for the first compartment,
   /// so we need two addresses for just that one.
+  ///
+  /// Compartment 1's TLS region is split because there is
+  /// a page of unprotected data for `ia2_stackptr_0` plus padding,
+  /// as we don't have a general implementation of shared TLS yet,
+  /// but `ia2_stackptr_0` is special-cased for now
+  /// as it must be stored in TLS and unprotected.
   uintptr_t tls_addr_compartment1_first;
   uintptr_t tls_addr_compartment1_second;
 };
--- a/runtime/libia2/include/ia2_threads.h
+++ b/runtime/libia2/include/ia2_threads.h
@@ -6,7 +6,7 @@ struct ia2_thread_metadata {
   /// The addresses of each compartment's stack for this thread.
   ///
   /// This data is shared, so it should not be trusted for use as a pointer,
-  /// but it can be used best effort as a provenance-less address.
+  /// but it can be used best effort for non-trusted purposes.
   uintptr_t stack_addrs[IA2_MAX_COMPARTMENTS];
 
   /// The addresses of each compartment's TLS region for this thread,
--- a/runtime/libia2/include/ia2_threads.h
+++ b/runtime/libia2/include/ia2_threads.h
@@ -8,6 +8,15 @@ struct ia2_thread_data {
   /// This data is shared, so it should not be trusted for use as a pointer,
   /// but it can be used best effort as a provenance-less address.
   uintptr_t stack_addrs[IA2_MAX_COMPARTMENTS];
+
+  /// The addresses of each compartment's TLS region for this thread,
+  /// except for compartment 1, which has split TLS regions (see below).
+  uintptr_t tls_addrs[IA2_MAX_COMPARTMENTS];
+
+  /// The TLS region is split only for the first compartment,
+  /// so we need two addresses for just that one.
+  uintptr_t tls_addr_compartment1_first;
+  uintptr_t tls_addr_compartment1_second;
 };
 
 struct ia2_addr_location {
--- a/runtime/libia2/include/permissive_mode.h
+++ b/runtime/libia2/include/permissive_mode.h
@@ -468,7 +468,7 @@ __attribute__((constructor)) void permissive_mode_init(void) {
 // `getline` calls `malloc` inside of `libc`,
 // but we wrap `malloc` with `__wrap_malloc`,
 // so we need to free what `getline` allocated with `__real_free`.
-void __real_free(void *ptr);
+typeof(IA2_IGNORE(free)) __real_free;
 
 extern uintptr_t ia2_stack_addrs[IA2_MAX_COMPARTMENTS];
 
--- a/runtime/libia2/include/permissive_mode.h
+++ b/runtime/libia2/include/permissive_mode.h
@@ -19,6 +19,7 @@
 #include <sys/ucontext.h>
 
 #include "ia2.h"
+#include "ia2_threads.h"
 
 /*
  * PKRU is defined to be bit 9 in the XSAVE state component bitmap (Intel SDM
@@ -470,8 +471,6 @@ __attribute__((constructor)) void permissive_mode_init(void) {
 // so we need to free what `getline` allocated with `__real_free`.
 typeof(IA2_IGNORE(free)) __real_free;
 
-extern uintptr_t ia2_stack_addrs[IA2_MAX_COMPARTMENTS];
-
 extern uintptr_t ia2_tls_addr_compartment1_first;
 extern uintptr_t ia2_tls_addr_compartment1_second;
 extern uintptr_t ia2_tls_addrs[IA2_MAX_COMPARTMENTS];
@@ -524,10 +523,6 @@ void log_memory_map(void) {
         fprintf(log, "[tls:tid ?:compartment 1]");
       }
       for (size_t pkey = 0; pkey < IA2_MAX_COMPARTMENTS; pkey++) {
-        if (start_addr == ia2_stack_addrs[pkey]) {
-          fprintf(log, "[stack:tid ?:compartment %zu]", pkey);
-          break;
-        }
         if (start_addr == ia2_tls_addrs[pkey]) {
           fprintf(log, "[tls:tid ?:compartment %zu]", pkey);
           break;
@@ -535,6 +530,10 @@ void log_memory_map(void) {
         if (partition_alloc_thread_isolated_pool_base_address && start_addr == (*partition_alloc_thread_isolated_pool_base_address)[pkey]) {
           fprintf(log, "[heap:compartment %zu]", pkey);
         }
+      }
+      const struct ia2_addr_location location = ia2_addr_location_find(start_addr);
+      if (location.name) {
+        fprintf(log, "[%s:tid %ld:compartment %d]", location.name, (long)location.tid, location.compartment);
       }
     }
 
--- a/runtime/libia2/include/permissive_mode.h
+++ b/runtime/libia2/include/permissive_mode.h
@@ -471,10 +471,6 @@ __attribute__((constructor)) void permissive_mode_init(void) {
 // so we need to free what `getline` allocated with `__real_free`.
 typeof(IA2_IGNORE(free)) __real_free;
 
-extern uintptr_t ia2_tls_addr_compartment1_first;
-extern uintptr_t ia2_tls_addr_compartment1_second;
-extern uintptr_t ia2_tls_addrs[IA2_MAX_COMPARTMENTS];
-
 extern uintptr_t (*partition_alloc_thread_isolated_pool_base_address)[IA2_MAX_COMPARTMENTS];
 
 void log_memory_map(void) {
@@ -519,21 +515,17 @@ void log_memory_map(void) {
       fprintf(log, "%s", path);
     } else {
       // No path, try to identify it.
-      if (start_addr == ia2_tls_addr_compartment1_first || start_addr == ia2_tls_addr_compartment1_second) {
-        fprintf(log, "[tls:tid ?:compartment 1]");
-      }
-      for (size_t pkey = 0; pkey < IA2_MAX_COMPARTMENTS; pkey++) {
-        if (start_addr == ia2_tls_addrs[pkey]) {
-          fprintf(log, "[tls:tid ?:compartment %zu]", pkey);
-          break;
-        }
-        if (partition_alloc_thread_isolated_pool_base_address && start_addr == (*partition_alloc_thread_isolated_pool_base_address)[pkey]) {
-          fprintf(log, "[heap:compartment %zu]", pkey);
-        }
-      }
       const struct ia2_addr_location location = ia2_addr_location_find(start_addr);
       if (location.name) {
         fprintf(log, "[%s:tid %ld:compartment %d]", location.name, (long)location.tid, location.compartment);
+      }
+      if (partition_alloc_thread_isolated_pool_base_address) {
+        for (size_t pkey = 0; pkey < IA2_MAX_COMPARTMENTS; pkey++) {
+          if (start_addr == (*partition_alloc_thread_isolated_pool_base_address)[pkey]) {
+            fprintf(log, "[heap:compartment %zu]", pkey);
+            break;
+          }
+        }
       }
     }
 
--- a/runtime/libia2/init.c
+++ b/runtime/libia2/init.c
@@ -1,5 +1,7 @@
 #include "ia2.h"
 #include "ia2_internal.h"
+#include "ia2_threads.h"
+
 #include <sys/auxv.h>
 #include <sys/prctl.h>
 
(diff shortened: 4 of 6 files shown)


runtime/libia2/init.c
```diff
@@ -1,7 +1,9 @@
 #include "ia2.h"
 #include "ia2_internal.h"
+// #include "ia2_allocator.h"
```
kkysen, line 3: How am I supposed to include `ia2_allocator`?  That's where `shared_realloc` seems to be defined.

runtime/libia2/init.c
```diff
@@ -13,7 +15,45 @@
...
+      return &ia2_thread_data_ptr[i];
+    }
+  }
+  if (ia2_thread_data_cap == 0) {
+    // Start capacity at 16 instead of 1, 2, 4, 8, 16.
+    ia2_thread_data_cap = 16;
+  }
+  if (ia2_thread_data_len == ia2_thread_data_cap) {
+    ia2_thread_data_cap = ia2_thread_data_cap == 0 ? 16 : ia2_thread_data_cap * 2;
+    ia2_thread_data_ptr = realloc(ia2_thread_data_ptr, ia2_thread_data_cap);
```
kkysen, line 45: This should be a `shared_realloc`, right, just like the globals are `IA2_SHARED_DATA`?

kkysen:
I'm getting these new errors about TLS sections not being adjacent anymore for all of the tests, but I'm not sure what's causing it.  Do `__thread` globals have to be declared differently or something?

```
[38/113] Linking C shared library tests/three_keys_minimal/unpadded/liblib_2.so
FAILED: tests/three_keys_minimal/unpadded/liblib_2.so
cd /home/kkysen/work/rust/ia2/build/x86_64/tests/three_keys_minimal && test ! -s /home/kkysen/work/rust/ia2/build/x86_64/tests/three_keys_minimal/three_keys_minimal_call_gates_2.objcopy || /home/kkysen/.linuxbrew/Cellar/llvm/19.1.4/bin/llvm-objcopy --redefine-syms=/home/kkysen/work/rust/ia2/build/x86_64/tests/three_keys_minimal/three_keys_minimal_call_gates_2.objcopy /home/kkysen/work/rust/ia2/build/x86_64/tests/three_keys_minimal/CMakeFiles/lib_2_unpadded.dir/./lib_2.c.o && cd /home/kkysen/work/rust/ia2/build/x86_64 && /home/kkysen/.linuxbrew/Cellar/llvm/19.1.4/bin/clang -fPIC -g  -fsanitize=undefined -Wl,--export-dynamic -Wl,@/home/kkysen/work/rust/ia2/build/x86_64/tests/three_keys_minimal/three_keys_minimal_call_gates_2.ld -pthread -Wl,--wrap=pthread_create -Wl,--wrap=main -Wl,-z,now -Wl,-z,relro -Wl,-T/home/kkysen/work/rust/ia2/build/x86_64/runtime/libia2/padding.ld -Wl,--dynamic-list=/home/kkysen/work/rust/ia2/build/x86_64/runtime/libia2/dynsym.syms -Wl,-wrap,calloc -Wl,-wrap,free -Wl,-wrap,malloc -Wl,-wrap,memalign -Wl,-wrap,posix_memalign -Wl,-wrap,pvalloc -Wl,-wrap,realloc -Wl,-wrap,valloc -Wl,-wrap,malloc_usable_size -Wl,-wrap,realpath -Wl,-wrap,strdup -Wl,-wrap,strndup -Wl,-wrap,getcwd -Wl,-wrap,asprintf -Wl,-wrap,vasprintf -shared -Wl,-soname,liblib_2.so -o tests/three_keys_minimal/unpadded/liblib_2.so tests/three_keys_minimal/CMakeFiles/lib_2_unpadded.dir/lib_2.c.o  -Wl,-rpath,/home/kkysen/work/rust/ia2/build/x86_64/runtime/partition-alloc:/home/kkysen/work/rust/ia2/build/x86_64/tests/three_keys_minimal:/home/kkysen/work/rust/ia2/build/x86_64/runtime/libia2  -ldl  runtime/libia2/liblibia2.a  runtime/partition-alloc/libpartition-alloc.so  tests/three_keys_minimal/libthree_keys_minimal_call_gates.so  runtime/libia2/liblibia2.a  -ldl  runtime/type-registry//debug/libtype_registry.a  runtime/libia2/libc.so.6 && :
/usr/bin/ld: tests/three_keys_minimal/unpadded/liblib_2.so: TLS sections are not adjacent:
/usr/bin/ld:        TLS: .tdata
/usr/bin/ld:        TLS: .tbss
/usr/bin/ld:    non-TLS: .init_array
/usr/bin/ld:    non-TLS: .fini_array
/usr/bin/ld:    non-TLS: .data.rel.ro
/usr/bin/ld:    non-TLS: .dynamic
/usr/bin/ld:    non-TLS: .got
/usr/bin/ld:    non-TLS: .data
/usr/bin/ld:        TLS: ia2_shared_data
/usr/bin/ld: map sections to segments failed: bad value
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[40/113] Linking C shared library tests/rewrite_fn_ptr_eq/librewrite_fn_ptr_eq_call_gates.so
ninja: build stopped: subcommand failed.
```

runtime/libia2/init.c
```diff
@@ -13,7 +15,45 @@
 extern __thread void *ia2_stackptr_0[PAGE_SIZE / sizeof(void *)]
     __attribute__((aligned(4096)));
 
-uintptr_t ia2_stack_addrs[MAX_PKEYS] IA2_SHARED_DATA = {0};
+__thread uintptr_t ia2_stack_addrs[MAX_PKEYS] IA2_SHARED_DATA = {0};
```
fw-immunant, line 18: `IA2_SHARED_DATA` does not work with thread locals at present, see #249.

fw-immunant:
The section error comes from using `IA2_SHARED_DATA` with a thread-local; per #249 we don't yet support thread-local shared data. In particular, our current approach involves putting shared data in its own section, but this approach can't work for thread-locals, which already must be in the `.tdata` or `.tbss` section, which is the root of the linker error above.

runtime/libia2/init.c
kkysen: I switched things to avoid this now.

runtime/libia2/init.c
ayrtonm: @fw-immunant what do you think about calling `__real_realloc` here instead of `shared_realloc`? I feel very uneasy about creating a dependency between the core libia2.a stuff and the partition alloc shim given our experience with firefox.

runtime/libia2/include/permissive_mode.h
```diff
@@ -470,7 +470,13 @@ __attribute__((constructor)) void permissive_mode_init(void) {
 // so we need to free what `getline` allocated with `__real_free`.
 void __real_free(void *ptr);
 
-extern uintptr_t ia2_stack_addrs[MAX_PKEYS];
+typedef struct {
+  pid_t tid;
+  uintptr_t stack_addrs[MAX_PKEYS];
+} IA2ThreadData;
```
ayrtonm, line 476: We should not define this struct twice so they don't get out of sync. I'm aware this is also a potential issue with the types in the variable declarations you used for labeling the other parts, but if we're defining an entire struct it's more likely we'll end up changing it so it's a bigger issue.

kkysen: Yeah, I'd definitely rather do that.  Is there a good place to put this?  I'd like to move the other `extern`s there, too.

kkysen (commented):
Per offline discussion, we don't want to create a dependency between libia2 and partition-alloc, and we also don't need to worry at this point about an unbounded number of threads.  So we can use a much simpler approach with a statically allocated or `mmap`ped array.

Also, I'll add locking to this, but we won't yet cleanup memory after threads exit (see https://github.com/immunant/IA2-Phase2/issues/546#issuecomment-2874022424).

runtime/libia2/include/permissive_mode.h
kkysen: Moved to `ia2_threads.h` and `threads.c`.  The `ia2_*` is because `threads.h` clashes.

runtime/libia2/include/ia2_threads.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
```
ayrtonm, line 1: If this is only intended to be used in ia2 sources then maybe let's put it in runtime/libia2.

runtime/libia2/threads.c
```diff
@@ -96,3 +94,100 @@ int __wrap_pthread_create(pthread_t *restrict thread,
...
+    .ia2_thread_data = {0},
+};
+
+struct ia2_thread_data *ia2_thread_data_get_current_thread(void) {
+  return ia2_all_threads_data_lookup(&threads, gettid());
+}
+
+struct ia2_addr_location ia2_addr_location_find(const uintptr_t addr) {
+  return ia2_all_threads_data_find_addr(&threads, addr);
+}
```
fw-immunant, line 193: Should all be behind `#ifdef IA2_DEBUG_LOG`.

tests/permissive_mode/permissive_mode.c
```diff
@@ -26,3 +28,24 @@ Test(permissive_mode, main) {
...
+}
+
+Test(permissive_mode, multithreaded) {
+  pthread_t threads[10];
+  for (size_t i = 0; i < 10; i++) {
+#if IA2_ENABLE
+    pthread_create(&threads[i], NULL, start_thread, NULL);
+#endif
+  }
+  // Exit before joining threads.
```
fw-immunant, line 49: Is there a particular reason not to join threads?

fw-immunant (changes requested):

runtime/libia2/include/ia2_threads.h
```diff
@@ -0,0 +1,52 @@
...
+};
+
+struct ia2_addr_location {
+  /// A descriptive name of what this address points to.
+  /// For example, "stack".
+  ///
+  /// `NULL` if unknown.
+  const char *name;
+
+  /// The thread ID of the thread this address is on.
```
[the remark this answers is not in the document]
ayrtonm: Maybe "the thread this address belongs to" would be clearer.

runtime/libia2/include/ia2_threads.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include "ia2.h"
+
+struct ia2_thread_data {
+  /// The addresses of each compartment's stack for this thread.
+  ///
+  /// This data is shared, so it should not be trusted for use as a pointer,
+  /// but it can be used best effort as a provenance-less address.
```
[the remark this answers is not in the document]
fw-immunant: It's not really a matter of provenance (which only exists at the abstract-machine level) as much as it is potential tampering (which occurs at the operational/machine level).

tests/permissive_mode/permissive_mode.c
ayrtonm: Presumably because if we were cleaning up thread stacks properly then joining everything means we wouldn't print any thread stacks or TLS at the end.

runtime/libia2/threads.c
ayrtonm: Agreed though `IA2_DEBUG_LOG` is more for making the runtime verbose so I'm not sure we should gate this under that same flag.

kkysen: This is making debug logs more verbose.  Isn't that what the flag is for?  Otherwise it's not named very well.

tests/permissive_mode/permissive_mode.c
kkysen: Yeah, exactly.  I wanted to see what the `/proc/self/maps` state is with all of the threads still existing.

fw-immunant: Makes sense, it would be nice if the comment also explained that intent.

ayrtonm:
We might want to eventually replace tid in the memory map with either `dladdr(fn_ptr_passed_to_pthread_create)` and/or `pthread_getname_np` if it's been set, though it doesn't have to be in this PR.

runtime/libia2/include/ia2_threads.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include "ia2.h"
+
+struct ia2_thread_data {
```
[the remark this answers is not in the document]
kkysen: It's only about the data layout right now, but that's not what it's meant to be entirely.  For example, we may also store the `pthread_t`, the start function, the thread name, the `pthread_create` `arg` argument, etc.  The last argument of `pthread_create` is also only colloquially called the data arg.  The source and man page call it `__arg` and `arg`.

runtime/libia2/include/ia2_threads.h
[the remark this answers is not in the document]
kkysen: > This is only true when we create an instance of this struct annotated as `IA2_SHARED_DATA`. That is the case presently, but isn't really a property of the struct type itself, but the one instance we create. And this holds for all fields of that instance, not just this one.

It's not, but it's meant as documentation, and a user may not look at the final declaration that contains this field, so I thought it's best to put the warning on the field that someone is mostly likely to check, especially when all instances of this struct are indeed `IA2_SHARED_DATA`.

runtime/libia2/include/ia2_threads.h
```diff
@@ -0,0 +1,52 @@
...
+  uintptr_t tls_addr_compartment1_first;
+  uintptr_t tls_addr_compartment1_second;
+};
+
+struct ia2_addr_location {
+  /// A descriptive name of what this address points to.
+  /// For example, "stack".
+  ///
+  /// `NULL` if unknown.
+  const char *name;
```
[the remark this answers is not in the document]
kkysen: I don't see what advantage that gives, just more boilerplate.  This is only meant to be checked by the user (printed out or examined in a debugger), not the programmer.

runtime/libia2/ia2.c
```diff
@@ -267,6 +262,8 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
 
   const int pkey = search_args->pkey;
 
+  struct ia2_thread_data *const thread_data = ia2_thread_data_get_current_thread();
+
```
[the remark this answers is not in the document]
kkysen: I thought @ayrtonm had said it's only a problem if we trust this data.

runtime/libia2/ia2.c
```diff
@@ -321,7 +318,9 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
           printf("ia2_mprotect_with_tag failed: %s\n", strerror(errno));
           exit(-1);
         }
-        ia2_tls_addr_compartment1_first = (uintptr_t)start_round_down;
+        if (thread_data) {
```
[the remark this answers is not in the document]
kkysen: The memory map logging is meant to be somewhat resilient to other things blowing up and still not crashing when logging or in a debugger.  So when things go wrong, I'd rather do nothing/best effort (`NULL`) than crash in a debugger.

runtime/libia2/include/ia2_threads.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include "ia2.h"
```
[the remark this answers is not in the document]
fw-immunant: Sure, a different flag is probably better than stretching the meaning of `IA2_DEBUG_LOG`.

runtime/libia2/include/ia2_threads.h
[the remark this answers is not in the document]
fw-immunant: If it's general information about the thread, `metadata` would be better than simply `data`.

runtime/libia2/include/ia2_threads.h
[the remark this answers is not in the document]
fw-immunant: It isn't a big deal, feel free to ignore.

runtime/libia2/include/ia2_threads.h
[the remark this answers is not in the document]
kkysen: Yeah that's a better name. I'll use that.

runtime/libia2/ia2.c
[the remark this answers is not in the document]
ayrtonm: I think @fw-immunant meant only maintain it to reduce the amount of things the TCB does and its overhead.

runtime/libia2/include/ia2_threads.h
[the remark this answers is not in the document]
ayrtonm: Agree that we should add a new flag instead of using `IA2_DEBUG_LOG`. Maybe `IA2_TRACK_MEMORY`?

runtime/libia2/threads.c
```diff
@@ -96,3 +94,100 @@ int __wrap_pthread_create(pthread_t *restrict thread,
...
+  size_t num_threads;
+  pid_t tids[MAX_THREADS_SUPPORTED];
+  struct ia2_thread_data ia2_thread_data[MAX_THREADS_SUPPORTED];
+};
+
+#define array_len(a) (sizeof(a) / sizeof(*(a)))
+
+struct ia2_thread_data *ia2_all_threads_data_lookup(struct ia2_all_threads_data *const this, const pid_t tid) {
+  struct ia2_thread_data *data = NULL;
+  if (pthread_mutex_lock(&this->lock) == -1) {
```
ayrtonm, line 114: What is `-1` here and should we check for the other error codes?

kkysen: It just says it returns -1 on error like a lot of libc functions do.

runtime/libia2/ia2.c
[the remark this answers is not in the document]
ayrtonm: If this returns `NULL` because the process is in a bad state I think it's better to complain loudly early on. If you notice an issue and allow the program to keep running the root problem will usually become harder to debug.

In the case that this returns `NULL` because we spun up too many threads it's fine to do what you're doing now.

runtime/libia2/include/ia2_threads.h
ayrtonm: If this is difficult because permissive_mode.h is a header, it's fine to leave it as-is btw.

runtime/libia2/threads.c
```diff
@@ -96,3 +94,100 @@ int __wrap_pthread_create(pthread_t *restrict thread,
   thread_thunk->data = data;
   return __real_pthread_create(thread, attr, ia2_thread_begin, thread_thunk);
 }
+
+// It's much simpler to only support a static number of created threads,
+// especially because we want to have very few dependencies.
+// If a program needs more threads, you can just increase this number.
+#define MAX_THREADS_SUPPORTED 4096
```
ayrtonm, line 101: We already limit the number of threads you can spin up. It's `IA2_MAX_THREADS` macro defined in ia2_internal.h

kkysen: Ahh okay, I'll use that then.

ayrtonm: @fw-immunant `IA2_MAX_THREADS` isn't explicitly referenced anywhere but we do limit the number of threads we can create right?

runtime/libia2/ia2.c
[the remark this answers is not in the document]
ayrtonm: Actually if we're already limiting the number of threads we can spin up, then I don't think there's ever a reason for this to return `NULL` other than the process being in a bad state. So we should probably just assert that it's non-null.

runtime/libia2/include/ia2_threads.h
kkysen: > If this is difficult because permissive_mode.h is a header, it's fine to leave it as-is btw.

Yeah, I need to include it in `permissive_mode.h` so it's difficult to do this.

runtime/libia2/include/ia2_threads.h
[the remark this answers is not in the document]
kkysen: Fixed in fbc1ccd18.

runtime/libia2/include/ia2_threads.h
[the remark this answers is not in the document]
kkysen: > It's not really a matter of provenance (which only exists at the abstract-machine level) as much as it is potential tampering (which occurs at the operational/machine level).

Fixed in 45a47f22f.

runtime/libia2/include/ia2_threads.h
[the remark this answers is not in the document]
kkysen: Fixed in 93a23e4ad.

tests/permissive_mode/permissive_mode.c
kkysen: Fixed in 3f9ef14c9.

runtime/libia2/threads.c
kkysen: Fixed in ce350b114.  I misread the man page.  It says it returns 0 on success, so I'm just checking `!= 0` now.

runtime/libia2/threads.c
kkysen: Fixed in 09ee68344.

runtime/libia2/include/ia2_threads.h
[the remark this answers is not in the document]
kkysen: Moving this to #564.

runtime/libia2/ia2.c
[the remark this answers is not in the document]
kkysen: Moving to #564.  Fixed in 22698cff7 and de173f93d.

runtime/libia2/threads.c
kkysen: Moved to #564.  Fixed in 22698cff7 and de173f93d.  I left a dummy impl of `ia2_addr_location_find` outside of `#if IA2_DEBUG_LOG` since it's used from `permissive_mode.h` and that's included in non-libia2 code.  But if we fix that, we can fix this here, too, although I'd prefer to do that in a separate PR where we change `permissive_mode.h` to not just be an `#include` anymore.

runtime/libia2/ia2.c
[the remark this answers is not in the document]
kkysen: > If this returns `NULL` because the process is in a bad state I think it's better to complain loudly early on. If you notice an issue and allow the program to keep running, the root problem will usually become harder to debug.

I added warnings if the pthread APIs return unexpected errors, but I'm worried you might run into one of these while debugging (e.x. you paused the debugger while a lock was held), and you don't want the debugger to just crash then.

runtime/libia2/include/ia2_threads.h
```diff
@@ -0,0 +1,52 @@
...
+  uintptr_t stack_addrs[IA2_MAX_COMPARTMENTS];
+
+  /// The addresses of each compartment's TLS region for this thread,
+  /// except for compartment 1, which has split TLS regions (see below).
+  uintptr_t tls_addrs[IA2_MAX_COMPARTMENTS];
+
+  /// The TLS region is split only for the first compartment,
+  /// so we need two addresses for just that one.
+  uintptr_t tls_addr_compartment1_first;
+  uintptr_t tls_addr_compartment1_second;
```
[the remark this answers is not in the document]
ayrtonm: It's compartment 1's TLS region that's split. I think what @fw-immunant  meant by "the split TLS region for compartment 0" was that we're splitting it to define a TLS variable in compartment 0. Specifically [this variable definition](https://github.com/immunant/IA2-Phase2/blob/main/runtime/libia2/include/ia2_internal.h#L419) which ends up in the main binary's TLS segment but we leave unprotected.

runtime/libia2/threads.c
fw-immunant: `IA2_MAX_THREADS` is not actually a limit we ever needed. I think it was added as part of work towards a different design of per-compartment stacks that didn't use TLS, but TLS ultimately worked better and that never shipped. I would say we should just remove the `#define`, but if we're going to use it in permissive mode maybe we should rename it to clarify that it only affects permissive mode.

kkysen: Moved to #571.

runtime/libia2/ia2.c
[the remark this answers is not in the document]
ayrtonm: > e.x. you paused the debugger while a lock was held

This reasoning doesn't make sense. Sure you could pause while holding the lock, but you shouldn't call this function (`protect_tls_pages`) from gdb. We should only be calling the print memory map function. I'd say this shouldn't be a blocker for merging this PR, but I'm still strongly in favor of catching errors early if we get into a bad state.

ayrtonm (approved):

runtime/libia2/ia2.c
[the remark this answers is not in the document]
kkysen: You're right about for this function (`protect_tls_pages`).  I was thinking about the print memory map function we want to call from a debugger.  I'll work on improving the error situation there and abort on errors unless it might be called from a debugger.

PR #558, opened 2025-05-12
rewriter: fix iteration through pre/post-condition fns using `.equal_range` instead of `.find`

kkysen:


fw-immunant (approved):
Right, this is needed to handle multiple pre/post conditions. LGTM.

PR #559, opened 2025-05-12
tests/pre_post_condition: add tests with fns with odd numbers of args

kkysen:
Fixes https://github.com/immunant/IA2-Phase2/pull/511#discussion_r2079684166.


ayrtonm:
CI failure should be fixed by #558

Issue #548, opened 2025-05-02
Move all runtime initialization code out of headers


ayrtonm:
As part of making the runtime easier to audit and debug we should most (or possibly all) code out of headers. This mainly applies to everything called by `INIT_RUNTIME` but we may also want to move `ia2_compartment_init.inc` to a proper `.c` file and consolidate all compartment initialization to one place. For the former the main drawback against this is that having it in a header makes it easier to ensure we emit code for the exact number of compartments we expect (i.e. we emit no unused `wrpkru`s to other compartment which may be abused as gadgets). However defining code in headers has significant drawbacks because it makes it much harder to reason about what DSOs code is located in and the control flow taken during init and thread init. While it would be very nice to avoid emitting unnecessary `wrpkru`s we should keep the runtime init code as simple as possible then look into potential ways of ensuring we don't create potential gadgets. The latter could be done through required compile flags when building libia2.a (e.g. `-DNUM_COMPARTMENTS`) or possibly removing the gadgets after initialization (e.g. unmapping the .text pages containing unused `wrpkru`s or temporarily mapping them as RWX and overwriting the instructions).

Issue #543, opened 2025-05-02
Allow making a shared library the main compartment


ayrtonm:
In Firefox we had to invoke the `INIT_RUNTIME` macro in a source file that gets built into libxul.so rather than the main binary. This moved some TLS definitions (e.g. ia2_stackptr_1) which required changing how they were accessed (e.g. from `__wrap_main` and the callgates). We should modify the way the runtime is initialized and the call gate code generator so it works regardless of which DSO the initial compartment is. This can be done as part of #320 and as we move code out of ia2_internal.h.

PR #561, opened 2025-05-12
Move compartment init to centralized location

### LINKED ISSUE #320: Merge `INIT_RUNTIME(n)` into `__wrap_main`
Interposing the C runtime's initialization in two ways is redundant and just adds noise to user-facing code/build system changes. Since we need to interpose `main` we should just keep `__wrap_main` and call everything included by `INIT_RUNTIME(n)` from there if possible. A potential minor issue is how to specify the number of pkeys that will be used (is this even necessary?).

Also `__wrap_main` assumes that the main binary uses pkey 1. This is a reasonable, but not well-documented assumption. We should add this to the docs and possibly add a way to validate this at build or run-time.

ayrtonm:
This WIP switches how compartments are initialized. Instead of using constructors this modifies `__wrap_main` to call a user-defined `ia2_start` function where libraries are their corresponding pkeys are specified with just function calls. The idea is to have something more similar to how pledge works where the code that runs before `main` (mostly ours) is part of the trusted code base then after `ia2_start` the program can try to do anything. This approach avoids the need for superfluous compartment transitions when protecting each compartment's memory during initialization (though some may still be necessary during thread creation). Since `ia2_start` must also `dlopen` the libraries passed we use that to get the various addresses that are passed as arguments to `protect_pages`. Closes #320 and #543 and will address part of #548.

@fw-immunant for now I just updated the two_keys_minimal test and `#if 0`ed out the previous `init_pkey` to test locally and make sure it still works.


runtime/libia2/main.c
```diff
@@ -14,6 +14,18 @@ __asm__(
 #if defined(__x86_64__)
     "pushq %rbp\n"
     "movq %rsp, %rbp\n"
+
+    // Call ia2_start making sure to preserve/restore the original arguments to main
+    "pushq %rdi\n"
+    "pushq %rsi\n"
+    "pushq %rdx\n"
+    "subq $8, %rsp\n"
+    "callq ia2_start\n"
```
ayrtonm, line 23: We may want to call whatever will call `ia2_set_up_tags` (once) here instead of directly calling the user-defined function.

ayrtonm: done (switched from calling `ia2_main` to `ia2_start`)

ayrtonm:
I just realized that since `ia2_main` is user-defined the order that compartments are initialized in is still variable. Afaict it shouldn't matter but just in case, we could have `ia2_protect_memory` just store its arguments then initialize them in a well-defined order. There are parts of initialization that do need to happen before others but with the way I have it now, it shouldn't matter what order we call `ia2_protect_memory` in.

kkysen:
> Closes #320 and #543 and will address part of #548.

Btw you need to having one of the closing keywords in front of each issue referenced (#543 is missing here).  That's why I usually put them in bullet points, which also previews the issue name, which I find more convenient to read.

ayrtonm:
From talking with @fw-immunant offline I think it might be a bit difficult to remove the rest of the code in `INIT_RUNTIME` (e.g. having something like the equivalent of `INIT_RUNTIME(15)` in libia2.a may create references to non-existent `ia2_stackptr_N` thread-locals and trigger ld errors).

Instead we could keep it and just rename it to more closely match what it does now. We could also use it to avoid the `dlsym` calls where we replace the last char (e.g. have it also expand to functions like `void *get_thread_local_ia2_stackptr(int n)`). This way we avoid having code that affects the order of operations during runtime initialization in headers and also avoid `dlsym`ing variable symbol names.

runtime/libia2/destructors.c
```diff
@@ -0,0 +1,104 @@
...
+  for (int i = 0; i < ehdr->e_phnum; i++) {
+    if (ubsan_access_phdr_type(phdr, i) == PT_DYNAMIC) {
+      dynamic_phdr = &phdr[i];
+      break;
+    }
+  }
+  if (!dynamic_phdr) {
+    return;
+  }
+  Elf64_Dyn *dynamic =
```
ahomescu, line 29: Nit: insert an empty line here to make this function more readable?

runtime/libia2/destructors.c
```diff
@@ -0,0 +1,104 @@
...
+      (Elf64_Dyn *)((uintptr_t)ehdr + dynamic_phdr->p_vaddr);
+  void *rewrite_base = NULL;
+  void *finalizers_region =
+      (void *)IA2_ROUND_DOWN((uintptr_t)finalizers, PAGE_SIZE);
+  size_t finalizers_region_size =
+      IA2_ROUND_DOWN(((uintptr_t)finalizers) + sizeof(finalizers) + PAGE_SIZE,
```
ahomescu, line 35: Is this a round-up? In that case, should it be `(uintptr_t)finalizers) + sizeof(finalizers) + PAGE_SIZE - 1`?

runtime/libia2/destructors.c
```diff
@@ -0,0 +1,104 @@
...
+  res = mprotect(finalizers_region, finalizers_region_size,
+                 PROT_READ | PROT_WRITE);
+  if (res != 0) {
+    fprintf(stderr, "Failed to set relro section writable: %d\n", res);
+    exit(1);
+  }
+  for (size_t i = 0; i < dynamic_phdr->p_memsz / sizeof(Elf64_Dyn); i++) {
+    if (dynamic[i].d_tag == DT_FINI || dynamic[i].d_tag == DT_FINI_ARRAY ||
+        dynamic[i].d_tag == DT_FINI_ARRAYSZ) {
+      uintptr_t dynamic_entry_addr = (uintptr_t)&dynamic[i].d_un.d_val;
```
ahomescu, line 47: Why is this `&d_val` instead of the value itself?

runtime/libia2/ia2.c
```diff
@@ -344,19 +344,25 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
 
   struct PhdrSearchArgs *search_args = (struct PhdrSearchArgs *)data;
 
-  size_t cur_pkey = ia2_get_tag();
+  size_t cur_pkey = search_args->pkey;//ia2_get_tag();
```
ahomescu, line 347: I'd either remove the comment, or expand it to explain why it used to be `ia2_get_tag` and you're changing it.

runtime/libia2/include/ia2_compartment_init.inc
```diff
@@ -118,15 +51,9 @@ __attribute__((visibility("default"))) void COMPARTMENT_IDENT(init_tls)(void) {
...
-struct FinalizerInfo {
-  uint64_t fini_offset;
-  uint64_t fini_array_offset;
-  uint64_t fini_array_size;
-};
-
 /* Filled in by runtime initialization with contents of  .fini_array and
  * .fini */
-static volatile struct FinalizerInfo finalizers
+volatile struct FinalizerInfo COMPARTMENT_IDENT(finalizers)
```
ahomescu, line 56: Do we need a visibility here?

runtime/libia2/init.c
```diff
@@ -1,16 +1,25 @@
...
+#endif
+#include <dlfcn.h>
+#include <elf.h>
+#include "ia2.h"
 #include "ia2_internal.h"
 #include <sys/auxv.h>
 #include <sys/prctl.h>
 
+/* Pass to mmap to signal end of program init */
+#define IA2_FINISH_INIT_MAGIC 0x1a21face1a21faceULL
```
ahomescu, line 12: This isn't exactly a "magic", but a random invalid pointer you're passing to `mmap` (it could still collide because of ASLR, but I guess the probability of that is low).

runtime/libia2/destructors.c
```diff
@@ -0,0 +1,104 @@
...
+          fprintf(stderr, "Failed to set dynamic section writable: %d\n", res);
+          exit(1);
+        }
+      }
+      if (dynamic[i].d_tag == DT_FINI) {
+        finalizers->fini_offset = dynamic[i].d_un.d_val;
+        dynamic[i].d_un.d_val =
+            (Elf64_Xword)dtor_callgate -
+            (Elf64_Xword)ehdr;
+      } else if (dynamic[i].d_tag == DT_FINI_ARRAY) {
```
ahomescu, line 80: What if you have both `DT_FINI` and `DT_FINI_ARRAY`?

runtime/libia2/include/ia2.h
```diff
@@ -174,6 +174,13 @@ uint32_t ia2_get_pkru();
 /// Returns the current compartment pkey
 size_t ia2_get_pkey();
 
+/// lib must be a single DSO or "main" for the executable
```
ahomescu, line 177: Using `NULL` for the executable might be more efficient than a hard-coded string.

runtime/libia2/include/ia2_internal.h
```diff
@@ -174,8 +174,19 @@ struct PhdrSearchArgs {
...
 
+struct FinalizerInfo {
+  uint64_t fini_offset;
+  uint64_t fini_array_offset;
+  uint64_t fini_array_size;
+};
+
+void ia2_setup_destructors(const Elf64_Ehdr *ehdr, int pkey, void *dtor_callgate, void *dtor_ptr, struct FinalizerInfo *finalizers);
+
+
```
ahomescu, line 188: What are the extra empty lines for?

runtime/libia2/include/ia2.h
```diff
@@ -174,6 +174,13 @@ uint32_t ia2_get_pkru();
...
+/// extra_libraries must be a semicolon-separated list of libraries
```
ahomescu, line 178: Where are these coming from? Could they be an array of pointers to strings instead?

runtime/libia2/init.c
```diff
@@ -97,3 +100,134 @@ __attribute__((__noreturn__)) void ia2_reinit_stack_err(int i) {
...
+        if (!handle) {
+            printf("%s: failed to dlopen %s for compartment %d\n", __func__, libs, compartment);
+            // TODO: use actual error codes
+            exit(-1);
+        }
+        // TODO: use a symbol that will always be defined. This is specific to tests/two_keys_minimal/plugin.c
+        dso_sym = "start_plugin";
+    }
+
+    void *dso_addr = dlsym(handle, dso_sym);
```
ahomescu, line 126: Could `LD_PRELOAD` give you the wrong symbol here (and everywhere you call `dlsym`)?

ahomescu: It might make sense to make all these symbols at least `protected` visibility, so they're harder to override.

thedataking:
> Looks like I'll need to find a workaround for `RTLD_DI_PHDR` since donna has an older glibc w/o it.

I've been holding off on upgrading `donna` because of IA2-Phase2. Very happy to roll to 24.04 but I wonder if we risk breaking something else that way. LMK what you think is best.

runtime/libia2/destructors.c
ayrtonm: same as above

runtime/libia2/destructors.c
ayrtonm: same as above

runtime/libia2/include/ia2_compartment_init.inc
ayrtonm: same as above

runtime/libia2/include/ia2_internal.h
ayrtonm: Removed

fw-immunant (approved):
Looks good in general, scattered comments inline but no functional problems. Let's finally get this merged!

runtime/libia2/init.c
```diff
@@ -108,3 +111,140 @@ __attribute__((__noreturn__)) void ia2_reinit_stack_err(int i) {
...
+ * Compartment 0 does not need to be protected so we don't need to store config info. This variable
+ * will only be accessed from the compartment defining ia2_main since it's the only one that should
+ * call ia2_register_compartment so other copies of it won't be referenced if libia2 is statically
+ * linked into all compartments.
+ */
+static struct CompartmentConfig user_config[IA2_MAX_COMPARTMENTS] IA2_SHARED_DATA = { 0 };
+
+/*
+ * Stores the main DSO and extra libraries (if any) for the specified compartment. This should only
+ * be called for protected compartments (calls specifying compartment 0 are no-ops).
```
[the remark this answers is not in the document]
ayrtonm: done

runtime/libia2/include/ia2.h
```diff
@@ -201,6 +201,19 @@ size_t ia2_get_tag();
 /// Returns the current compartment number
 size_t ia2_get_compartment();
 
+/// Registers the DSOs associated with a given protected compartment for use in ia2_main.
+///
+/// `lib` must be a DSO or "main" for the main executable.`extra_libraries` must be a
+/// semicolon-separated list of libraries. The pointers must remain valid after ia2_main returns.
```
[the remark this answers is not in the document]
ayrtonm: done

runtime/libia2/include/ia2.h
```diff
@@ -201,6 +201,19 @@ size_t ia2_get_tag();
 /// Returns the current compartment number
 size_t ia2_get_compartment();
 
+/// Registers the DSOs associated with a given protected compartment for use in ia2_main.
+///
+/// `lib` must be a DSO or "main" for the main executable.`extra_libraries` must be a
```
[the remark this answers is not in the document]
ayrtonm: done

runtime/libia2/include/ia2.h
```diff
@@ -201,6 +201,19 @@ size_t ia2_get_tag();
 /// Returns the current compartment number
 size_t ia2_get_compartment();
 
+/// Registers the DSOs associated with a given protected compartment for use in ia2_main.
```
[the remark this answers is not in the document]
ayrtonm: No, it should be two sentences.

runtime/libia2/include/ia2_internal.h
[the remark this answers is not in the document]
ayrtonm: done

PR #564, opened 2025-05-15
permissive_mode: only store shared addrs in `IA2_DEBUG_LOG` mode and warn about using them as ptrs

### LINKED ISSUE #554: Only store `IA2_SHARED_DATA` stack, TLS, etc. data in permissive mode/`IA2_DEBUG_LOG` mode
See https://github.com/immunant/IA2-Phase2/pull/421#discussion_r2076033377.

> Because this is `IA2_SHARED_DATA`, this is definitely not secure. That's fine if it's just for logging in permissive mode, but we should have a comment to that effect so nothing security-critical starts relying on it.

Moreover, we shouldn't store this insecure information in the first place outside of permissive mode.

kkysen:
* Fixes #554.

### DIFF
--- a/runtime/libia2/include/ia2_threads.h
+++ b/runtime/libia2/include/ia2_threads.h
@@ -2,6 +2,8 @@
 
 #include "ia2.h"
 
+#if IA2_DEBUG_LOG
+
 /// The data here is shared, so it should not be trusted for use as a pointer,
 /// but it can be used best effort for non-trusted purposes.
 struct ia2_thread_metadata {
@@ -27,6 +29,17 @@ struct ia2_thread_metadata {
   uintptr_t tls_addr_compartment1_second;
 };
 
+/// Find the `struct ia2_thread_metadata*` for the current thread,
+/// adding (but not allocating) one if there isn't one yet.
+/// If there is no memory for more or an error, `NULL` is returned.
+/// This is a purely lookup and/or additive operation,
+/// so the lifetime of the returned `struct ia2_thread_metadata*` is infinite,
+/// and since it's thread-specific,
+/// it is thread-safe to read and write.
+struct ia2_thread_metadata *ia2_thread_metadata_get_current_thread(void);
+
+#endif
+
 struct ia2_addr_location {
   /// A descriptive name of what this address points to.
   /// For example, "stack".
@@ -49,15 +62,6 @@ struct ia2_addr_location {
   /// `-1` if unknown.
   int compartment;
 };
-
-/// Find the `struct ia2_thread_metadata*` for the current thread,
-/// adding (but not allocating) one if there isn't one yet.
-/// If there is no memory for more or an error, `NULL` is returned.
-/// This is a purely lookup and/or additive operation,
-/// so the lifetime of the returned `struct ia2_thread_metadata*` is infinite,
-/// and since it's thread-specific,
-/// it is thread-safe to read and write.
-struct ia2_thread_metadata *ia2_thread_metadata_get_current_thread(void);
 
 /// Find the `ia2_addr_location` of `addr`.
 /// If it is not found or there is an error,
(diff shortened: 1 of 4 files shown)


fw-immunant (changes requested):
I think it would be better to have `ia2_all_threads_data_lookup` always return non-`NULL` (conceptually, it doesn't make much sense for such a function to fail) and `#ifdef` out the uses that fill it in.

kkysen:
> I think it would be better to have `ia2_all_threads_data_lookup` always return non-`NULL` (conceptually, it doesn't make much sense for such a function to fail) and `#ifdef` out the uses that fill it in.

It'll always return null if there's no more memory though.

fw-immunant:
> > I think it would be better to have `ia2_all_threads_data_lookup` always return non-`NULL` (conceptually, it doesn't make much sense for such a function to fail) and `#ifdef` out the uses that fill it in.
> 
> It'll always return null if there's no more memory though.

Aborting if anything unexpected (such as OOM) happens seems preferable to complicating callers.

runtime/libia2/include/ia2_threads.h
```diff
@@ -2,6 +2,8 @@
 
 #include "ia2.h"
 
+#if IA2_DEBUG_LOG
```
kkysen, line 5: So it seems like you both agree we shouldn't use `IA2_DEBUG_LOG` for this (https://github.com/immunant/IA2-Phase2/pull/557#discussion_r2093574063), even though it sounds pretty much exactly like what it's supposed to do.  The more defines we add, though, the more combinations we need to test and the more a user has to understand, even though these two uses are very similar.

ayrtonm: `IA2_DEBUG_LOG` can get very verbose if you spin up lots of threads and those messages are intertwined with whatever messages the program normally prints out to stderr/stdout. The idea here is to have a mode that doesn't add extra messages to the logs but allows you to get a better picture of what's in the address space.

ayrtonm: Ideally you shouldn't have to set a flag to enable the memory tracking, but since it serializes thread creation and limits the number of threads this isn't an option. It's also a good idea to do as little as possible in the TCB for something that would run in production so I think another flag here is acceptable.

ayrtonm:
To allow enabling this per-test I was thinking of something like replacing
```
#if IA2_DEBUG_LOG
if (thread_metadata) {
    thread_metadata->tls_addrs[pkey] = (uintptr_t)start_round_down;
}
#endif
```
with
```
ia2_track_tls_addrs(pkey, (uintptr_t)start_round_down);
```
where `ia2_track_tls_addrs` is declared as
```
void nop(void) { }
__attribute__((weak, alias("nop"))) void ia2_track_tls_addrs(int pkey, uintptr_t addr);
```
in the source files where it's called and defined as
```
void ia2_track_tls_addrs(int pkey, uintptr_t addr) {
    thread_metadata->tls_addrs[pkey] = addr;
}
```
either in a header that we `#include` to enable memory tracking or possibly in `INIT_RUNTIME`. The latter may be harder to do since the `INIT_RUNTIME` macro expansion itself can't contain `#if IA2_TRACK_MEMORY`. @fw-immunant @kkysen thoughts on this?

ayrtonm:
Also I should point out that in CI only `IA2_DEBUG` (adding an `assert(rdpkru)` to the callgates) is in the test matrix. We unconditionally set `IA2_DEBUG_LOG`  (which is poorly named and should be renamed `IA2_VERBOSE` or similar). Unconditionally setting or disabling the memory tracking in CI is also an option though unlike the verbose mode flag the behavior when memory tracking is enabled or disabled differs enough that I'm leaning against that.

runtime/libia2/include/ia2_threads.h
kkysen: Okay, I opened #572 for this.

fw-immunant (approved):
LGTM, this is just making code that maintains additional debugging metadata conditionally-compiled.

PR #566, opened 2025-05-16
ci: use more specific `rustup toolchain install` command

### LINKED ISSUE #565: ci: `rustup toolchain install stable-x86_64-unknown-linux-gnu` broken after 1.87.0 release due to `rls-preview` component
After Rust 1.87.0 was released today, CI has been failing on this error:

```
> rustup toolchain install stable-x86_64-unknown-linux-gnu
  shell: /usr/bin/bash -e {0}
info: syncing channel updates for 'stable-x86_6[4](https://github.com/immunant/IA2-Phase2/actions/runs/15059592127/job/42332109143?pr=557#step:2:5)-unknown-linux-gnu'
info: latest update on 2025-05-15, rust version 1.87.0 (17067e9ac 202[5](https://github.com/immunant/IA2-Phase2/actions/runs/15059592127/job/42332109143?pr=557#step:2:6)-05-09)
error: component 'rls-preview' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel 'stable'
One or many components listed above might have been permanently removed from newer versions
of the official Rust distribution due to deprecation.

You can find the list of removed components at
<https://rust-lang.github.io/rustup/devel/concepts/components.html#previous-components>.

If you are updating an existing toolchain, after determining the deprecated component(s)
in question, please remove them with a command such as:

    rustup component remove --toolchain stable <COMPONENT>...

After that, you should be able to continue with the update as usual.
```

https://github.com/immunant/IA2-Phase2/actions/runs/15059592127/job/42332109143?pr=557

kkysen:
* Fixes #565.

This is the same ultimate `rustup toolchain install` command that `actions-rust-lang/setup-rust-toolchain@v1` uses, but that action is much slower (~70 s vs. ~17 s previously for the whole workflow), so we just run the main `rustup toolchain install` command directly.

I fixed #565 (likely) by `rm -rf ~/.rustup` in a temporary CI step, but this also adds `--profile minimal --no-self-update --allow-downgrade`, which should hopefully help with not running into situations like #565 again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,9 @@ jobs:
         ia2-tracer: [ON, OFF]
     steps:
       - name: Setup Rust
-        run: rustup toolchain install stable-x86_64-unknown-linux-gnu
+        # We don't use actions-rust-lang/setup-rust-toolchain because it's slower
+        # (~17 s vs. ~70 s) and we need to sequentially run this 32 times.
+        run: rustup toolchain install stable --target x86_64-unknown-linux-gnu --profile minimal --no-self-update --allow-downgrade
       - name: Check out code
         uses: actions/checkout@v4
       - name: Init b63 submodule # Can't use checkout@v4 because it unconditionally inits all


fw-immunant:
This makes every instance of the `test-on-pkru-enabled-host` job go from ~17s to ~70s. Is there any other way to fix this that would be faster? Is this actually necessary after removing the old `.rustup`?

kkysen:
> This makes every instance of the `test-on-pkru-enabled-host` job go from ~17s to ~70s. Is there any other way to fix this that would be faster? Is this actually necessary after removing the old `.rustup`?

Ah yeah, that's bad.  I didn't realize.  I changed to running `rustup toolchain install` directly again, but with all of the extra minimal flags the action passes, too.  And I also left a comment explaining why we do this for perf.

fw-immunant (approved):
Great, thanks for figuring out how to speed this back up.

PR #570, opened 2025-05-19
library-only mode (sans FF-specific work)

fw-immunant:
We still need to merge the additional changes required for FF compartmentalization, but triage for those is ongoing. These changes shouldn't break anything upstream and are needed for library-only mode for any target application.

### DIFF
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -1175,7 +1175,7 @@ auto ValidateDirectory = CLI::Validator(
     },
     "Directory", "path of a directory");
 
-const std::map<std::string, int> arch_map{{"x86", (int)Arch::X86}, {"aarch64", (int)Arch::Aarch64}};
+const std::map<std::string, Arch> arch_map{{"x86", Arch::X86}, {"aarch64", Arch::Aarch64}};
 
 std::string CompilationDbPath;
 std::string LibraryFilesFile;
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -1302,8 +1302,14 @@ int main(int argc, const char **argv) {
 
   // Collect mapping of filenames to pkeys and used pkeys
   std::set<Pkey> pkeys_used;
+  std::set<std::string> LibraryFilesSet(LibraryFiles.begin(), LibraryFiles.end());
   for (auto s : SourceFiles) {
-    auto pkey = pkey_from_commands(get_commands, s);
+    std::optional<Pkey> pkey;
+    if (LibraryOnlyMode) {
+      pkey = LibraryFilesSet.contains(s) ? 0 : 1;
+    } else {
+      pkey = pkey_from_commands(get_commands, s);
+    }
     if (!pkey) {
       return -1;
     }
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -1231,6 +1231,28 @@ int main(int argc, const char **argv) {
   }
   CompilationDatabase &comp_db = *MaybeCmds;
 
+  // Library-only mode file groups
+  llvm::SmallVector<llvm::StringRef> LibraryFiles;
+  llvm::SmallVector<llvm::StringRef> RewriteFiles;
+
+  // Read library and rewrite filenames from input files for library-only mode
+  if (LibraryOnlyMode) {
+    assert(SourceFiles.empty());
+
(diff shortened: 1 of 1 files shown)


fw-immunant:
Note: split the C++-tolerance changes out into separate PR.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -1165,14 +1256,31 @@ auto ValidateDirectory = CLI::Validator(
...
+
 int main(int argc, const char **argv) {
   CLI::App app{"IA2 Source Rewriter"};
+  auto LibFilesOpt =
+      app.add_option("--library-files", LibraryFilesFile, "Path to file listing untrusted library sources")
+          ->group(LibOnlyGroup)
+          ->option_text("FILE")
+          ->check(CLI::ExistingFile);
+  auto RwFilesOpt =
+      app.add_option("--rewrite-files", RewriteFilesFile, "Path to file listing application sources to rewrite")
```
ahomescu, line 1277: Why do you need this to be separate from `source-files`? Could you use that in library-only mode?

tools/rewriter/SourceRewriter.cpp
```diff
@@ -1207,6 +1315,36 @@ int main(int argc, const char **argv) {
   }
   CompilationDatabase &comp_db = *MaybeCmds;
 
+  std::set<std::string> LibraryFilesSet;
+
+  // Read library and rewrite filenames from input files for library-only mode
+  if (LibraryOnlyMode) {
+    assert(SourceFiles.empty());
```
ahomescu, line 1322: Does the order between `LibraryFiles` and `RewriteFiles` matter? If not, you could get rid of the latter (per my comment above) and use `SourceFiles` in its place. You could copy its original contents to `RewriteFilesSet` first, then add in `LibraryFiles`.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -172,6 +197,48 @@ static bool ignore_function(const clang::Decl &decl,
...
+  }
+
+  Filename filename = get_filename(*loc, sm);
+
+  bool is_empty = filename.empty();
+  if (is_empty) {
+    return true;
+  }
+
+  // if (filename.find("include/fmt") != std::string::npos) {
```
ahomescu, line 222: Will you need this eventually? If not, you could delete the whole thing.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -172,6 +197,48 @@ static bool ignore_function(const clang::Decl &decl,
...
+  if (filename.starts_with(RootDirectory) && !filename.starts_with(OutputDirectory)) {
+    llvm::errs() << "internal error: querying if we should modify file under root directory (this should not happen): " << filename << "\n";
+    exit(1);
+  }
+
+  if (!filename.starts_with(OutputDirectory)) {
+    return true;
+  }
+
+  return false;
```
ahomescu, line 239:
```suggestion
  return !filename.starts_with(OutputDirectory);
```

tools/rewriter/SourceRewriter.cpp
```diff
@@ -1165,14 +1256,31 @@ auto ValidateDirectory = CLI::Validator(
...
+std::string RewriteFilesFile;
 std::vector<std::string> SourceFiles;
 std::vector<std::string> ExtraArgs;
 
+auto LibOnlyGroup = "Library-only mode";
+
 int main(int argc, const char **argv) {
   CLI::App app{"IA2 Source Rewriter"};
+  auto LibFilesOpt =
+      app.add_option("--library-files", LibraryFilesFile, "Path to file listing untrusted library sources")
```
ahomescu, line 1272: Since the argument is the path to a file containing the library files (and not the actual files themselves), could it have a better name like `--library-files-list-file`?

Alternatively, why not just pass in the files directly? If the command line becomes too long, there are other options: `xargs`, or we could add more generic support for `@file` that other Linux tools support.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -1257,8 +1395,14 @@ int main(int argc, const char **argv) {
   // Collect mapping of filenames to pkeys and used pkeys
   std::set<Pkey> pkeys_used;
   for (auto s : SourceFiles) {
-    auto pkey = pkey_from_commands(get_commands, s);
+    std::optional<Pkey> pkey;
+    if (LibraryOnlyMode) {
+      pkey = LibraryFilesSet.contains(s) ? 0 : 1;
```
ahomescu, line 1400: Since you already have `default_pkey` defined, could `1` here be that instead?

Issue #572, opened 2025-05-20
Split `IA2_DEBUG_LOG` into `IA2_VERBOSE` and `IA2_DEBUG_MEMORY`


kkysen:
Right now, `IA2_DEBUG_LOG` is used for verbose logging for debugging purposes, and #564 adds the memory map tracking and labeling to it as well.  However, the verbose logging is quite verbose and so `IA2_DEBUG_LOG` shouldn't be used for all debug logging like the memory maps.  So we can rename `IA2_DEBUG_LOG` to `IA2_VERBOSE` for the existing uses to be more descriptive, and move the memory map code to `IA2_DEBUG_MEMORY` to allow turning that on more precisely.

See https://github.com/immunant/IA2-Phase2/pull/564#discussion_r2095373615 and https://github.com/immunant/IA2-Phase2/pull/557#discussion_r2093574063.

If we test all combinations of these flags in CI, CI runtimes will explode exponentially, as they are currently run sequentially on our self-hosted runner.  However, we currently always run with `IA2_DEBUG_LOG` on rather than testing both, and we can start with doing the same for `IA2_DEBUG_MEMORY`, or at least for the permissive mode tests (although we want to isolate those in #568 as well).  If we're able to run the self-hosted CI jobs in parallel (they are individually pretty fast), we might be able to test more combinations in CI.

PR #573, opened 2025-05-21
libia2/memory_maps: split memory maps out from permissive mode

### LINKED ISSUE #568: permissive_mode: detach `/proc/self/maps` labeling from permissive mode
Printing and labeling the `/proc/self/maps` memory map is quite useful as a general debugging tool and there's no reason it needs to be tied to permissive mode.  We should expose a function, e.g. `print_memory_map`, that we can call from `gdb` as well as continuing to print it to a log in permissive mode.  This also allows us to view the labeled memory map while debugging permissive mode itself, which can be helpful in debugging Firefox.

kkysen:
* Fixes #568.

This splits memory maps (`include/ia2_memory_maps.h`, `memory_maps.h`, `memory_maps.c`) out of permissive mode (`include/permissive_mode.h`, `include/ia2_threads.h`, `threads.c`).

Previously, `permissive_mode.h` contained `log_memory_maps`, a local function that printed the labeled memory maps to the log file.  Now this is moved to the public IA2 function `ia2_log_memory_maps` in `ia2_memory_maps.h`, which can log to any `FILE*`, including `stderr` in a debugger.

Now that the whole public function is limited to just `ia2_log_memory_maps`, all of the details previously exposed in `ia2_threads.h` can be made private within `libia2` again, moved to `memory_maps.{c,h}`.

This pretty much only moves code around.  It also moves a lot more code behind `IA2_DEBUG_LOG`, since only `ia2_log_memory_maps` is exported instead of `ia2_addr_location_find`.  So without `IA2_DEBUG_LOG` (which I'll soon change in #572), only an unlabeled memory map is logged.

Other fixes, like handling locking and errors when potentially called in a debugger better, are being left to later.


runtime/libia2/memory_maps.c
```diff
@@ -0,0 +1,198 @@
...
+  return ia2_all_threads_metadata_lookup(&threads);
+}
+
+struct ia2_addr_location ia2_addr_location_find(const uintptr_t addr) {
+  return ia2_all_threads_metadata_find_addr(&threads, addr);
+}
+
+extern uintptr_t (*partition_alloc_thread_isolated_pool_base_address)[IA2_MAX_COMPARTMENTS];
+
+void label_memory_map(FILE *log, uintptr_t start_addr) {
```
ayrtonm, line 118: nit: should be `static` or prefixed with `ia2_` (probably the former)

ayrtonm (approved):
LGTM

PR #574, opened 2025-05-22
libia2: split `IA2_DEBUG_LOG` into `IA2_VERBOSE` and `IA2_DEBUG_MEMORY`

### LINKED ISSUE #572: Split `IA2_DEBUG_LOG` into `IA2_VERBOSE` and `IA2_DEBUG_MEMORY`

kkysen:
* Fixes #572.

`IA2_VERBOSE` and `IA2_DEBUG_MEMORY` and more precise and descriptive names, and it can be useful to enable one but not the other.

We also don't test with either of them on (previously `IA2_DEBUG_LOG`) in CI, so this enables that.  To avoid exponentially increasing CI times (because the self-hosted runner runs jobs sequentially), this only tests with them on rather than off, too.

### DIFF
--- a/runtime/libia2/memory_maps.c
+++ b/runtime/libia2/memory_maps.c
@@ -3,7 +3,7 @@
 
 // Only enable this code that stores these addresses when debug logging is enabled.
 // This reduces the trusted codebase and avoids runtime overhead.
-#if IA2_DEBUG_LOG
+#if IA2_DEBUG_MEMORY
 
 struct ia2_all_threads_metadata {
   pthread_mutex_t lock;
@@ -136,11 +136,11 @@ static void label_memory_map(FILE *log, uintptr_t start_addr) {
   }
 }
 
-#else // IA2_DEBUG_LOG
+#else // IA2_DEBUG_MEMORY
 
 static void label_memory_map(FILE *log, uintptr_t start_addr) {}
 
-#endif // IA2_DEBUG_LOG
+#endif // IA2_DEBUG_MEMORY
 
 // `getline` calls `malloc` inside of `libc`,
 // but we wrap `malloc` with `__wrap_malloc`,
--- a/runtime/libia2/memory_maps.c
+++ b/runtime/libia2/memory_maps.c
@@ -138,7 +138,7 @@ static void label_memory_map(FILE *log, uintptr_t start_addr) {
 
 #else // IA2_DEBUG_LOG
 
-void label_memory_map(FILE *log, uintptr_t start_addr) {}
+static void label_memory_map(FILE *log, uintptr_t start_addr) {}
 
 #endif // IA2_DEBUG_LOG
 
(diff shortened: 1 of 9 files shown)


runtime/libia2/memory_maps.c
```diff
@@ -136,11 +136,11 @@ static void label_memory_map(FILE *log, uintptr_t start_addr) {
   }
 }
 
-#else // IA2_DEBUG_LOG
+#else // IA2_DEBUG_MEMORY
 
-void label_memory_map(FILE *log, uintptr_t start_addr) {}
+static void label_memory_map(FILE *log, uintptr_t start_addr) {}
```
ayrtonm, line 141: Why can't `ia2_log_memory_maps` be defined as the no-op for `!IA2_DEBUG_MEMORY` instead of this function?

ayrtonm (approved):

runtime/libia2/memory_maps.c
ayrtonm: If it's to support the config `permissive_mode + !IA2_DEBUG_MEMORY` I think it might make sense to require `IA2_DEBUG_MEMORY` to use permissive mode. @fw-immunant any thoughts?

fw-immunant: Permissive mode is already somewhat invasive, so I don't think it's likely to be a problem for it to require `IA2_DEBUG_MEMORY`. But I don't have a strong opinion either way here.

kkysen: > Why can't `ia2_log_memory_maps` be defined as the no-op for `!IA2_DEBUG_MEMORY` instead of this function?

Without `IA2_DEBUG_MEMORY`, it just prints what's in `/proc/self/maps` (with some columns removed), which it doesn't need to track any memory maps to do, which add overhead, etc.  So I thought it's good to still print that without `IA2_DEBUG_MEMORY`.

PR #575, opened 2025-05-22
libia2/memory_maps: make `IA2_MAX_THREADS` private and move it to `memory_maps.c` behind `IA2_DEBUG_MEMORY`

### LINKED ISSUE #571: Rename `IA2_MAX_THREADS` and scope it to memory debugging flag
`IA2_MAX_THREADS` was added in 40314f9 as part of work toward a different threading design, but the limited isn't needed or enforced anymore.  The `/proc/maps/self` memory map labeling in #557 uses it now, however, so we should rename it and scope it to the flag that the memory map labeling/debugging/tracking ends up under.

See https://github.com/immunant/IA2-Phase2/pull/557#discussion_r2095889891.

kkysen:
* Fixes #571.

`IA2_MAX_THREADS` was once used as part of a different design of per-compartment stacks, but we never actually used that non-TLS design, so there is no limit to the number of threads.  However, the memory maps code uses it now, so this just moves it there.  Thus, it's no longer public in `ia2_internal.h` and behind `IA2_DEBUG_MEMORY` in a `*.c` file.

~~Unlike what #571 and https://github.com/immunant/IA2-Phase2/pull/557#discussion_r2095889891 say, this doesn't actually rename it.  I didn't really see a reason to do so now that it's clearly behind `IA2_DEBUG_MEMORY` and in `memory_maps.c`, so it's not visible outside that file.~~

It's renamed to just `MAX_THREADS` now.  It's not a public export anymore so it doesn't need the `IA2_*` prefix anymore.

### DIFF
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -202,7 +202,6 @@ asm(".macro movz_shifted_tag_x18 tag\n"
 #define errno_s (strerror_l(errno, uselocale((locale_t)0)))
 
 #define PTRS_PER_PAGE (PAGE_SIZE / sizeof(void *))
-#define IA2_MAX_THREADS (PTRS_PER_PAGE)
 
 #define IA2_ROUND_DOWN(x, y) ((x) & ~((y)-1))
 
--- a/runtime/libia2/memory_maps.c
+++ b/runtime/libia2/memory_maps.c
@@ -5,6 +5,8 @@
 // This reduces the trusted codebase and avoids runtime overhead.
 #if IA2_DEBUG_MEMORY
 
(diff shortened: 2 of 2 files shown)


fw-immunant (approved):
I would suggest renaming to `MAX_THREADS`, as it's not describing a limitation of IA2 itself, but just an implementation detail of the memory map thread tracking. But this does avoid exporting a misleading #define (even internally), so LGTM.

kkysen:
> I would suggest renaming to `MAX_THREADS`, as it's not describing a limitation of IA2 itself, but just an implementation detail of the memory map thread tracking. But this does avoid exporting a misleading #define (even internally), so LGTM.

Okay I'll rename it, too, then.

PR #576, opened 2025-05-22
libia2/memory_maps: label start fn

kkysen:
* Fixes #567.

<details>
  <summary>An example <code>mpk_log_*</code>:</summary>

```
  start addr-end addr     perms offset  path
176260000000-176260001000 ---p 00000000 [heap:compartment 1]
176260001000-176260002000 rw-p 00000000
176260002000-176260004000 ---p 00000000
176260004000-176260008000 rw-p 00000000
176260008000-176270000000 ---p 00000000
200bfffff000-201000000000 ---p 00000000
3c1c00000000-3c2000000000 ---p 00000000
5f22d9d50000-5f22d9d5a000 r--p 00000000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5f22d9d5a000-5f22d9d8f000 r-xp 0000a000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5f22d9d8f000-5f22d9da4000 r--p 0003f000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5f22d9da5000-5f22d9da6000 r--p 00054000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5f22d9da6000-5f22d9dab000 rw-p 00055000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5f22d9dab000-5f22d9e42000 rw-p 0005a000 /home/kkysen/work/rust/ia2/build/x86_64/tests/permissive_mode/permissive_mode
5f22d9e42000-5f22da784000 rw-p 00000000
5f22e41aa000-5f22e41cb000 rw-p 00000000 [heap]
74aec0c00000-74aec1000000 rw-p 00000000 [stack:compartment 0:tid 1390230 (thread permissive_mode) (start fn start_thread)]
74aec1000000-74aec1400000 rw-p 00000000 [stack:compartment 1:tid 1390230 (thread permissive_mode) (start fn start_thread)]
74aec1400000-74aec1800000 rw-p 00000000 [stack:compartment 0:tid 1390229 (thread permissive_mode) (start fn start_thread)]
74aec1800000-74aec1c00000 rw-p 00000000 [stack:compartment 1:tid 1390229 (thread permissive_mode) (start fn start_thread)]
74aec1c00000-74aec1c01000 ---p 00000000
[...]

PR #578, opened 2025-05-22
cmake: don't make call gates link `libia2` just for its include dirs

kkysen:
Instead, just add the include directories explicitly.

This was causing issues with #576, causing link errors for `__real_free`. See <https://github.com/immunant/IA2-Phase2/pull/576#issuecomment-2902013380>.


fw-immunant (approved):
LGTM. Technically a self-review but I'll count this as an r+ from @kkysen too.

ayrtonm (approved):

PR #579, opened 2025-05-22
libia2/memory_maps: handle `sscanf` failure

kkysen:
Fixes https://github.com/immunant/IA2-Phase2/pull/555/files#r2102990872.


fw-immunant (approved):
LGTM.

PR #580, opened 2025-05-22
libia2: add `-Wl,--wrap=free` for when we link `libia2` but not the `partition-alloc` shim

kkysen:
We use `__real_free` in `memory_maps.c`, so we need this now.


fw-immunant (changes requested):
Add a comment linking this to our usage of `__real_free`.

fw-immunant (approved):

PR #582, opened 2025-05-25
tests/memory_maps: add a memory maps test with multiple compartments and threads

kkysen:
This differs from the existing permissive mode test in that it just prints the labeled memory maps to `stdout` instead of saving them to a log file. This is more convenient during testing.
It also tests more compartments.

### DIFF
--- a/tests/memory_maps/CMakeLists.txt
+++ b/tests/memory_maps/CMakeLists.txt
@@ -1,0 +1,26 @@
+define_shared_lib(
+  LIBNAME 2
+  SRCS 2.c
+  PKEY 2
+)
+
+define_shared_lib(
+  LIBNAME 3
+  SRCS 3.c
+  PKEY 3
+)
+
+define_shared_lib(
+  LIBNAME 4
+  SRCS 4.c
+  PKEY 4
+)
+
+define_test(
+  SRCS main.c
+  PKEY 1
+  LIBS 2 3 4
+  NEEDS_LD_WRAP
+  CRITERION_TEST
+  WITHOUT_SANDBOX # tracer forbids reading `/proc/self/maps`
+)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -70,6 +70,8 @@ add_subdirectory(tls_protected)
 
 add_subdirectory(threads)
 add_subdirectory(protected_threads)
+add_subdirectory(memory_maps)
+
 # The following tests are not supported on ARM64 yet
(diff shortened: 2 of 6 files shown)


ayrtonm:
Was this just failing because of the `define_ia2_wrapper` in the cmake file?

kkysen:
> Was this just failing because of the `define_ia2_wrapper` in the cmake file?

No, it was failing before, too, although on much fewer CI jobs.  I removed the `define_ia2_wrapper` now, though, and CI is re-running (seems to have paused the first time for some reason, re-running again now).

tests/memory_maps/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+  PKEY 4
+)
+
+define_test(
+  SRCS main.c
+  PKEY 1
+  NEEDS_LD_WRAP
+  CRITERION_TEST
+  LIBS 2 3 4
+)
```
kkysen, line 25: @fw-immunant, do I need `WITHOUT_SANDBOX` here, like `permissive_mode/CMakeLists.txt` has?  `fopen("/proc/self/maps", "r")` is failing with `ia2-sandbox`.

kkysen: Okay now it's still hanging indefinitely in CI even with `WITHOUT_SANDBOX`, which fixed it locally.

tests/memory_maps/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+  PKEY 4
+)
+
+define_test(
+  SRCS main.c
+  PKEY 1
+  LIBS 2 3 4
+  NEEDS_LD_WRAP
+  CRITERION_TEST
+  WITHOUT_SANDBOX
```
fw-immunant, line 25: Is memory-map debugging incompatible with the tracer? Document here whether that's fundamental or blocked on fixing a particular bug.

kkysen: Hmm, I'm not sure, but the `fopen("/proc/self/maps", "r")` is failing.

fw-immunant: Ok, tracked at #594.

kkysen: Is this good to merge then?  And we'll fix the tracer separately?

fw-immunant: Yep, go for it.

PR #586, opened 2025-05-28
runtime/libia2: Simplify use of IA2_VERBOSE

ayrtonm:
This simplifies most uses of IA2_VERBOSE to make logging easier and more uniform. This commit intentionally ignores the existing use of IA2_VERBOSE in ia2_internal.h since (ideally) the static function defined in that header should be move to a source file at some point.

### DIFF
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -442,3 +442,9 @@ __attribute__((__noreturn__)) void ia2_reinit_stack_err(int i);
     REPEATB##n(setup_destructors_for_compartment, nop_macro);                  \
     mark_init_finished();                                                      \
   }
+
+#if IA2_VERBOSE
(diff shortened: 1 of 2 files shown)


runtime/libia2/include/ia2_internal.h
```diff
@@ -442,3 +442,9 @@ __attribute__((__noreturn__)) void ia2_reinit_stack_err(int i);
     REPEATB##n(setup_destructors_for_compartment, nop_macro);                  \
     mark_init_finished();                                                      \
   }
+
+#if IA2_VERBOSE
+#define ia2_log(fmt, ...) fprintf(stdout, "%s:" fmt, __func__, __VA_ARGS__)
```
kkysen, line 447: Do we want to print file and line, too?

kkysen (approved):

PR #590, opened 2025-06-09
Fix CI

fw-immunant:
There wasn't too much post-reinstall fallout but something is failing to find LLVM's `lit`. Hopefully this fixes it.


.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
             -DCMAKE_C_FLAGS="-fuse-ld=${{ matrix.linker }}"              \
             -DClang_DIR=$Clang_DIR                                       \
             -DLLVM_DIR=$LLVM_DIR                                         \
-            -DLLVM_EXTERNAL_LIT=$HOME/.local/bin/lit                     \
+            -DLLVM_EXTERNAL_LIT=/usr/lib/llvm-18/build/utils/lit/lit.py  \
```
kkysen, line 48: Do we want to hardcode this path?

Issue #287, opened 2023-09-22
TLS destructors


rinon:
We need a test that includes thread-local destructors. I think that compartment-specific thread-local destructors are going to fail, as we don't wrap these destructors to switch into the right compartment first.

PR #593, opened 2025-06-10
libia2: free compartment stacks on thread termination

### LINKED ISSUE #546: Deallocate compartment stacks when a thread terminates

kkysen:
* Fixes #546.

I did this with `pthread_key_create` and `pthread_setspecific`, as this makes the implementation very simple and should handle all ways thread terminate (`return`, `pthread_exit`, `pthread_cancel`).

`pthread_key_create` creates a TLS key with a destructor function that's called whenever a thread terminates (however that happens).  This is used to `munmap` all the thread stacks.  Since multiple stacks are used per thread and `pthread_key_create` can only store a single ptr as its data, the thread stacks are stored in a separate `__thread` TLS array and `pthread_setspecific` is used to simply store a non-`NULL` value as the data so that the destructor will eventually be run.

### DIFF
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -415,8 +415,6 @@ __attribute__((__noreturn__)) void ia2_reinit_stack_err(int i);
 #endif
 /* clang-format on */
 
-void ia2_create_thread_keys(void);
-
 #define _IA2_INIT_RUNTIME(n)                                                   \
   __attribute__((visibility("default"))) int ia2_n_pkeys_to_alloc = n;                                                \
   __attribute__((visibility("default"))) __thread void *ia2_stackptr_0[PAGE_SIZE / sizeof(void *)]                    \
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -415,7 +415,7 @@ __attribute__((__noreturn__)) void ia2_reinit_stack_err(int i);
 #endif
 /* clang-format on */
 
-void create_thread_keys(void);
+void ia2_create_thread_keys(void);
 
 #define _IA2_INIT_RUNTIME(n)                                                   \
   __attribute__((visibility("default"))) int ia2_n_pkeys_to_alloc = n;                                                \
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -415,6 +415,8 @@ __attribute__((__noreturn__)) void ia2_reinit_stack_err(int i);
 #endif
 /* clang-format on */
 
+void create_thread_keys(void);
+
 #define _IA2_INIT_RUNTIME(n)                                                   \
   __attribute__((visibility("default"))) int ia2_n_pkeys_to_alloc = n;                                                \
   __attribute__((visibility("default"))) __thread void *ia2_stackptr_0[PAGE_SIZE / sizeof(void *)]                    \
--- a/runtime/libia2/init.c
+++ b/runtime/libia2/init.c
@@ -58,7 +58,7 @@ static void thread_stacks_destructor(void *_unused) {
 
 /// Create `thread_stacks_key`.
 /// This should be called once per process at the very beginning, currently in `ia2_init`.
-void create_thread_keys(void) {
+void ia2_create_thread_keys(void) {
   const int result = pthread_key_create(&thread_stacks_key, thread_stacks_destructor);
   if (result != 0) {
     fprintf(stderr, "pthread_key_create failed: %s\n", strerrorname_np(result));
@@ -291,7 +291,7 @@ void ia2_start(void) {
     ia2_setup_destructors();
     /* Set up global resources. */
     ia2_set_up_tags();
-    create_thread_keys();
+    ia2_create_thread_keys();
     verify_tls_padding();
     /* allocate an unprotected stack for the untrusted compartment */
     allocate_stack_0();
--- a/runtime/libia2/init.c
+++ b/runtime/libia2/init.c
@@ -27,13 +27,21 @@ static pthread_key_t thread_stacks_key IA2_SHARED_DATA;
 __thread void *stacks[IA2_MAX_COMPARTMENTS] = {0};
 
 void thread_stacks_destructor(void *_unused) {
+#if IA2_VERBOSE
+  char thread_name[16] = {0};
+  if (pthread_getname_np(pthread_self(), thread_name, sizeof(thread_name)) != 0) {
+    ia2_log("pthread_getname_np failed on thread %ld\n", (long)gettid());
+    thread_name[0] = '?';
+  }
+#endif
+
   for (size_t compartment = 0; compartment < IA2_MAX_COMPARTMENTS; compartment++) {
     void *const stack = stacks[compartment];
     if (!stack) {
       continue;
     }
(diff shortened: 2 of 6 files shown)


fw-immunant:
See #287 re: the testing we need.

kkysen (commented):
> > Re [#546 (comment)](https://github.com/immunant/IA2-Phase2/issues/546#issuecomment-2956539580), this does seem to work locally and doesn't run into issues with the thread destructors being in the wrong compartment.
> 
> This is very surprising to me, because I don't think we perform any compartment dispatch, and I don't see how this code would determine which compartment to switch to. Did you check which compartment each thread-specific destructor runs in? Does this mean we no longer need to fix #546 separately?

I checked and you're right, `munmap` is called from different compartments as the stack it is freeing.  This works because all of the current thread tests (`threads`, `protected_threads`, `permissive_mode`, `memory_maps`) run `WITHOUT_SANDBOX`, as the tracer fails on all of them.  So it's not catching the `munmap` from the wrong compartment here either since it's just not running.

runtime/libia2/include/ia2_internal.h
```diff
@@ -424,6 +424,8 @@ __attribute__((__noreturn__)) void ia2_reinit_stack_err(int i);
 #endif
 /* clang-format on */
 
+void create_thread_keys(void);
```
ayrtonm, line 427: nit: maybe namespace this since it's not static (i.e. `ia2_create_thread_keys`)

runtime/libia2/init.c
```diff
@@ -15,6 +16,50 @@
...
+static pthread_key_t thread_stacks_key IA2_SHARED_DATA;
+
+/// The compartment stacks for each thread.
+static __thread void *stacks[IA2_MAX_COMPARTMENTS] = {0};
+
+/// The destructor for `thread_stacks_key`.
+/// It deallocates all of the compartment stacks for the current thread being destructed.
+static void thread_stacks_destructor(void *_unused) {
+#if IA2_VERBOSE
+  char thread_name[16] = {0};
```
ayrtonm, line 31: Please add a comment explaining the 16. It wasn't immediately obvious to me that that wasn't an arbitrary value.

runtime/libia2/init.c
```diff
@@ -15,6 +16,50 @@
...
+#endif
+
+  for (size_t compartment = 0; compartment < IA2_MAX_COMPARTMENTS; compartment++) {
+    void *const stack = stacks[compartment];
+    if (!stack) {
+      continue;
+    }
+    ia2_log("deallocating stack for compartment %zu on thread %ld (%s): %p..%p\n",
+            compartment, (long)gettid(), thread_name, stack, stack + STACK_SIZE);
+    if (munmap(stack, STACK_SIZE) == -1) {
```
ayrtonm, line 46: nit: compare with `MAP_FAILED` instead of `-1`

runtime/libia2/init.c
```diff
@@ -32,12 +77,26 @@ char *allocate_stack(int i) {
...
   }
 
+  ia2_log("allocating stack for compartment %d on thread %ld: %p..%p\n", i, (long)gettid(), stack, stack + STACK_SIZE);
 #if IA2_DEBUG_MEMORY
   struct ia2_thread_metadata *const thread_metadata = ia2_thread_metadata_get_current_thread();
   if (thread_metadata) {
     thread_metadata->stack_addrs[i] = (uintptr_t)stack;
   }
 #endif
+  stacks[i] = stack;
```
ayrtonm, line 87: nit: Maybe add a sanity check that `stacks[i]` is NULL?

runtime/libia2/init.c
```diff
@@ -32,12 +77,26 @@ char *allocate_stack(int i) {
...
+  // The value set doesn't matter here as long as it's non-`NULL`.
+  // `allocate_stack` is called for each compartment
+  // when a new thread is created, so we just need to set a non-`NULL` value
+  // (which triggers a destructor running on thread termination),
+  // but it doesn't really matter what value it is,
+  // since the destructor `thread_stacks_destructor`
+  // just uses the TLS global `stack` directly.
+  const int result = pthread_setspecific(thread_stacks_key, (void *)stacks);
```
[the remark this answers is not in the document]
kkysen: Yeah, it's fine.  Only the first call makes a difference, as all we care about is if the value is `NULL` or not.  We never use the value.

runtime/libia2/init.c
```diff
@@ -15,6 +16,50 @@
 extern __thread void *ia2_stackptr_0[PAGE_SIZE / sizeof(void *)]
     __attribute__((aligned(4096)));
 
+/// The TLS pthread key for storing thread stacks.
+/// The value of the data set with `pthread_setspecific` is not used;
+/// it just matters if it was set (to non-`NULL`) or not.
+static pthread_key_t thread_stacks_key IA2_SHARED_DATA;
+
+/// The compartment stacks for each thread.
+static __thread void *stacks[IA2_MAX_COMPARTMENTS] = {0};
```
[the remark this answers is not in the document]
ayrtonm: I think this will trigger a compartment violation in the case where a thread terminates while not in compartment 1 (i.e. the same compartment this static variable is defined in), but that's ok for now. We can merge something that works first then look at making it work with compartments w/o making everything shared.

ayrtonm (approved):
LGTM after fixing some minor nits

runtime/libia2/init.c
ayrtonm: oh I misread the man page.

runtime/libia2/include/ia2_internal.h
kkysen: Fixed in 1d36dae3c.

runtime/libia2/init.c
kkysen: Fixed in c66f38cda.  Also fixed it in `memory_maps.c` where I did the same thing.

runtime/libia2/init.c
```diff
@@ -15,6 +16,50 @@
 extern __thread void *ia2_stackptr_0[PAGE_SIZE / sizeof(void *)]
     __attribute__((aligned(4096)));
 
+/// The TLS pthread key for storing thread stacks.
+/// The value of the data set with `pthread_setspecific` is not used;
+/// it just matters if it was set (to non-`NULL`) or not.
+static pthread_key_t thread_stacks_key IA2_SHARED_DATA;
+
+/// The compartment stacks for each thread.
```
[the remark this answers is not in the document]
kkysen: Fixed in c1c5d4ee5.

runtime/libia2/init.c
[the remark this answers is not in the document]
kkysen: Clarified in 9616dde65.

runtime/libia2/thread_name.h
```diff
@@ -0,0 +1,22 @@
...
+#include <string.h>
+#include <unistd.h>
+
+#define THREAD_NAME_MAX_LEN 16
+
+struct thread_name {
+  char name[THREAD_NAME_MAX_LEN];
+};
+
+static inline struct thread_name thread_name_get(pthread_t thread) {
```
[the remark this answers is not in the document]
kkysen: Fixed in 0fb55d6.

PR #596, opened 2025-06-13
test_runner: make `cr_assert` unconditional and use instead of `assert`

kkysen:
Pulled out of #524.

@fw-immunant, this is done instead of `assert` since `assert` sometimes is more like `debug_assert!`, right?


fw-immunant:
Merged as part of #524.

fw-immunant:
> @fw-immunant, this is done instead of `assert` since `assert` sometimes is more like `debug_assert!`, right?

Yes, C `assert` is the equivalent of Rust `debug_assert!` in that they are both skipped in release builds. C doesn't have an "official" notion of what a release build is; `assert` changes its behavior based on whether the `NDEBUG` macro is defined.

PR #598, opened 2025-06-14
type-registry: use `panic_abort!` instead of `panic!`

kkysen:
`panic!`, even with `panic = "abort"`, still goes through the panic handler machinery, which unfortunately uses TLS, which doesn't work with IA2 correctly.

This changes all panics to use `panic_abort!` instead, which behaves like `panic!`, printing the panic message, location info, and backtrace, but then immediately and directly `abort()`s instead of going through the panic machinery.

The alternative is to make `type-registry` `#![no_std]` and define our own `#[panic_handler]`, which can work well, and is more thorough, but then we can't use a ton of things from `std`, like locks.

Thus, we'd have to implement our own locks, and this is tricky. The simplest is a spin lock (there's a `spin` crate already, too), but that's not ideal in userspace.  Wrapping `pthread` is an option, but this is annoyingly difficult since `pthread_mutex_t` and `pthread_rwlock_t` can't be moved, so we'd need to `Box` them, which means we can't `const` initialize them. And existing implementations like `std` and `parking_lot` are in `std` or depend on it, so we can't use them even if they could've just depended only on `futex` and `park`.

### DIFF
--- a/runtime/type-registry/src/lib.rs
+++ b/runtime/type-registry/src/lib.rs
@@ -8,6 +8,10 @@ use std::hash::DefaultHasher;
 use std::ptr;
 use std::sync::RwLock;
 
+use crate::panic::panic_abort;
+
+mod panic;
+
 pub type Ptr = *const ();
 
 /// A pointer's address without provenance, used purely as an address for comparisons.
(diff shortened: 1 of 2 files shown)


fw-immunant:
Superseded by commits in #524. This approach doesn't necessarily change the behavior of all panics, so it's more robust to define our own panic handler.

PR #599, opened 2025-06-14
tests/terminating_threads: add a test for various combinations of ways threads can terminate

kkysen:
This tests various combinations of threads terminating, including a matrix of:
* 1, 2, and 11 threads (including the main thread)
* threads terminating by `return`, `exit(0)`, `abort()`, `pthread_exit(0)` (including the main thread) or `pause()`ing or `sleep`ing
* threads being `pthread_join`ed, `pthread_cancel`ed, or through process termination

However, `abort()` currently incurs a ~1 second penalty for some reason, so that's disabled. Moreover, `pthread_exit` causes `SIGILL` and `pthread_cancel` causes `SIGSEGV`, so these are disabled, too.

Furthermore, just like the other existing thread tests (`threads` and `protected_threads`), `terminating_threads` runs `WITHOUT_SANDBOX`, as the tracer fails on all of them.

I'll add tests covering `atexit`, `pthread_cleanup_push`, `pthread_key_create`, and other process/thread termination cleanup functions next in a subsequent PR.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -72,6 +72,7 @@ add_subdirectory(threads)
 add_subdirectory(protected_threads)
 add_subdirectory(memory_maps)
 add_subdirectory(signals)
+add_subdirectory(terminating_threads)
 
 # The following tests are not supported on ARM64 yet
 if (NOT LIBIA2_AARCH64)
--- a/tests/terminating_threads/CMakeLists.txt
+++ b/tests/terminating_threads/CMakeLists.txt
@@ -1,0 +1,12 @@
+define_shared_lib(
+  SRCS lib.c
+  PKEY 2
+)
+
+define_test(
+  SRCS main.c
+  PKEY 1
+  NEEDS_LD_WRAP
+  CRITERION_TEST
+  WITHOUT_SANDBOX # other thread tests require this, too
+)
--- a/tests/terminating_threads/lib.c
+++ b/tests/terminating_threads/lib.c
@@ -1,0 +1,4 @@
+#include <ia2_test_runner.h>
+
+#define IA2_COMPARTMENT 2
+#include <ia2_compartment_init.inc>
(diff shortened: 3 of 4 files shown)


fw-immunant:
In order to test thread cleanup, it would be good to somehow assert that the per-compartment stacks for the thread have been deallocated. We could do this via test cases that save the stack pointer from a thread and then attempt to dereference it from the corresponding compartment in the main thread after joining the thread.

kkysen (commented):
> In order to test thread cleanup, it would be good to somehow assert that the per-compartment stacks for the thread have been deallocated. We could do this via test cases that save the stack pointer from a thread and then attempt to dereference it from the corresponding compartment in the main thread after joining the thread.

So I could do this for a single non-main thread, but how would I test this for multiple non-main threads?  The test is destructive.

kkysen:
I think I can use `mincore` instead of reading `/proc/self/maps` to do it more simply in a non-destructive way.  It seems to be working.

kkysen:
I'm not sure what could be causing this error:

https://github.com/immunant/IA2-Phase2/actions/runs/15779474376/job/44481333621#step:19:8947

> running suite 'terminating_threads' test 'threads_11_other_threads_return', expecting exit status 0 (EXIT_SUCCESS)...
> double free or corruption (out)
> qemu: uncaught target signal 6 (Aborted) - core dumped
> killed by signal 6 (SIGABRT), but expected exit status 0

There's no allocation in `tests/terminating_threads`, and I don't see any invalid writes (maybe I'm missing them).

randomPoison (approved):
These tests look reasonable to me. You may still want to check for stack deallocation as Frances suggested, but that seems like something that could be done in a follow up PR.

kkysen:
> You may still want to check for stack deallocation as Frances suggested, but that seems like something that could be done in a follow up PR.

Yeah, that's in another PR (which I didn't open yet because the PR tree is too complex).

PR #604, opened 2025-06-20
test_runner: make `cr_assert` `assert` if `!NDEBUG` as `assert` prints source info

kkysen:


fw-immunant (approved):
Sure, more-informative debug output is always good.

PR #609, opened 2025-06-26
type-registry: some `no_std` improvements/fixes

kkysen:
This fixes some things in the `no_std` implementation of the `type-registry`, specifically:

* We use `libc` functions from the `libc` crate instead of declaring them ourselves (sometimes slightly wrong).
* We use `spin::RwLock` instead of rolling our own spin-based `Mutex`, which hasn't been checked for correctness, and is less ideal than a `RwLock`, which it originally was.
* ~~We use `hashbrown::HashMap` instead of `alloc::collections::BTreeMap`.  We would normally use `std::collections::HashMap`, but it hasn't yet been moved to `alloc`, and it internally uses `hashbrown` anyways.  Moreover, we use `fnv` as the hasher, as that's `const fn` compatible.~~
* Some other little cleanups.

### DIFF
--- a/runtime/type-registry/Cargo.lock
+++ b/runtime/type-registry/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "autocfg"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
+
+[[package]]
 name = "libc"
 version = "0.2.174"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


kkysen (commented):
> Miostly looks good, but I don't know what the motivation is for the `HashMap` change. If the only advantage is asymptotic performance, I think it makes more sense to keep our dependency list shorter instead.

It's mainly for performance (not just asymptotic) and because that's what it was originally and I didn't think we should change to `BTreeMap`.  For the dependencies, `fnv` is really the only new one but it's quite small.  `hashbrown` is what `std` uses.

kkysen:
I can cut that commit if you really prefer, but I think it's better with it.

fw-immunant:
> I can cut that commit if you really prefer, but I think it's better with it.

I'd rather drop it; sans that commit, let's merge this.

PR #615, opened 2025-07-21
tests/terminating_threads: `#if 0` out broken tests instead of using `exit(0)`, which has bugs itself

kkysen:
`exit` does cleanups, so it may have some issues that may be causing some of the CI heisenbugs.  `#if 0` these tests out seems to help CI avoid errors now.  Note that `_exit` was still causing errors.


ayrtonm:
Please just use `#if 0`. I see no potential downside to it.

PR #616, opened 2025-07-25
tests/terminating_threads: skip `threads_11_other_threads_return` on aarch64

kkysen:
Debugging CI.

#615 didn't work 😭.


kkysen (commented):
> We should just disable this test on ARM

That's what I'm trying to do, though.

ayrtonm:
If it's still flaky after disabling individual `Test` invocations disable the whole thing for ARM in `tests/CMakeLists.txt`.

PR #585, opened 2025-05-27
(WIP) libia2/memory_maps: redo thread support

kkysen:
Still debugging...

This reformulates thread support for memory maps.  This gets rid of locks and just uses an atomic counter in exchange for guaranteeing each thread registers only once to start.

PR #619, opened 2025-07-28
memory_maps: move `ia2_threads_metadata` global definition to `INIT_RUNTIME` to have a once-mapped definition

kkysen:
Without this, `liblibia2.a` contains the `ia2_threads_metadata`, which is then linked into each compartment's `*.so`, causing each compartment to reference a different version of the global, causing lots of errors and segfaults.

This also requires moving the type and other definitions from `memory_maps.{h,c}` to `ia2.h`.

This is split off of #585 as it's really a separate issue that was surfaced in trying to implement #585.  Also, this is now rebased after #561 was merged, as #585 was waiting on #561 to merge first for a cleaner implementation.

### DIFF
--- a/runtime/libia2/include/ia2_internal.h
+++ b/runtime/libia2/include/ia2_internal.h
@@ -13,6 +13,8 @@ struct dl_phdr_info;
 #define _GNU_SOURCE
 #endif
 
+#include "ia2_common.h"
+
 #include <assert.h>
 #include <errno.h>
 #include <link.h>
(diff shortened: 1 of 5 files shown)


runtime/libia2/include/ia2_internal.h
```diff
@@ -294,38 +293,41 @@ asm(".macro movz_shifted_tag_x18 tag\n"
...
-        : "x9");                                                               \
-    return out;                                                                \
+#define return_stackptr_if_compartment(compartment)               \
+  if (tag == compartment) {                                       \
+    void *out;                                                    \
+    __asm__ volatile(                                             \
+        "mrs x9, tpidr_el0\n"                                     \
+        "adrp %0, :gottprel:ia2_stackptr_" #compartment           \
+        "\n"                                                      \
+        "ldr %0, [%0, #:gottprel_lo12:ia2_stackptr_" #compartment \
```
ayrtonm, line 318: The clang-format on should go below this

ayrtonm (approved):

kkysen (commented):
> I do agree that we only want a single copy of that variable in the address space to make things easier to understand, but in my opinion the way to do that is to make libia2.a a shared library. Stuffing more variables into a macro makes things harder to understand if anything, so if this PR isn't solving a specific problem I don't think we should merge it.

Oh, sure, we could definitely make libia2 a shared library instead. I just thought we decided that wasn't worth it. Happy to change.

ayrtonm:
> I just thought we decided that wasn't worth it.

It definitely is lower priority relative to closing the PRs we have open, the tracer stuff and compartmentalizing glibc (especially since the latter two are big holes in the whole sandbox), but I think it would be worthwhile to make things easier to understand/audit.

We should open an issue for this and first enumerate everything that libia2.a and the rewriter-generated sources provide. Then figure out what we can put into a shared library and what we'd need to link into each DSO (and find a reasonably ergonomic way to do that).

PR #628, opened 2025-08-12
WIP runtime/libia2: Protect RW segments for libc and ld.so

ayrtonm:
This is currently expected to crash most programs since calls into libc from compartments other than pkey = 1 do not change compartments.

### DIFF
--- a/runtime/libia2/ia2.c
+++ b/runtime/libia2/ia2.c
@@ -423,6 +423,16 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
   return 0;
 }
 
+static bool is_syslib(struct dl_phdr_info *info, int32_t pkey) {
+  const char *libname = basename(info->dlpi_name);
+  bool is_libc = !strcmp(libname, "libc.so.6");
+  bool is_ldso = !strcmp(libname, "ld-linux-x86-64.so.2");
+  // TODO: Check for C++ standard library
+  // Always put C standard library and dynamic linker in compartment 1. This is an arbitrary choice.
+  const int32_t syslib_pkey = 1;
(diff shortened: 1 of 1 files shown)


runtime/libia2/main.c
```diff
@@ -1,5 +1,9 @@
 #include "ia2.h"
 
+void ia2_compartment_destructor_1(void);
```
ayrtonm, line 3: TODO: Move this somewhere else/generate it from the rewriter.

runtime/libia2/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 4.0)
 project(libia2)
 
-add_library(libia2 ia2.c init.c threads.c main.c exit.c memory_maps.c thread_name.c)
+add_library(libia2 ia2.c init.c threads.c main.c memory_maps.c thread_name.c)
```
ayrtonm, line 4: TODO: Do we want to keep the option of exiting the process in the untrusted compartment?

tests/tls_protected/main.c
```diff
@@ -70,3 +70,9 @@ Test(tls_protected, no_access_lib_secret) {
 Test(tls_protected, access_lib_secret) {
   run_test(true);
 }
+
+#if IA2_REWRITING
+void *__tls_get_addr(size_t m, size_t offset) {
```
ayrtonm, line 75: This could be autogenerated by the rewriter w/o requiring that it appears in user-code, but this was the easiest way to test this. This function is defined in ld.so which is in the same compartment as main.c and the call to `__tls_get_addr` in library.c is required for the rewriter to realize it needs to generate callgates to call it from that compartment. It then adds the `-Wl,--wrap` flag for this function and uses that when linking library.o.

PR #637, opened 2025-08-19
Increase the size of the thread-isolated address pools

randomPoison:
This was something I encountered when working on Firefox. With our changes to partition-alloc, all of our allocations go through their thread-isolated address pools, which only have 256 MB of address space, causing OOM issues when running Firefox. Since we only use the thread-isolated pools, I think giving them the max pool size is reasonable.


ayrtonm:
This seems fine. I'd only say remove the commented out previous value and add a note that we only use this pool in the PA directory's README (ideally with a link to one of the chromium pages that explains what the pools are).

ayrtonm (approved):

PR #642, opened 2025-09-30
libc/ld compartmentalization (minus dynamic allocations)

oinoom:


runtime/libia2/ia2.c
```diff
@@ -436,6 +444,10 @@ int protect_pages(struct dl_phdr_info *info, size_t size, void *data) {
     search_args->found_library_count++;
   }
 
+  if (syslib) {
+    printf("IA2: Protecting system library %s in compartment 1\n", libname);
```
ayrtonm, line 448: You can use `ia2_log` which works just like printf but automatically prefixes the function name

tools/rewriter/GenCallAsm.cpp
```diff
@@ -1207,6 +1217,36 @@ std::string emit_asm_wrapper(
 
   emit_prologue(aw, caller_pkey, target_pkey, arch);
 
+  if (trace_target_pkey.has_value()) {
+    if (arch == Arch::X86) {
+      add_asm_line(aw, ".ifdef IA2_TRACE_EXIT");
```
ayrtonm, line 1222: I don't think this works. `.ifdef` expects an ELF symbol (e.g. global var or function) named `IA2_TRACE_EXIT` but I think we only define a preprocessor macro named `IA2_TRACE_EXIT`

runtime/libia2/exit_callgates.c
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Restores the stack pointer (if it was switched) and the caller's PKRU. Every
+// `ia2_callgate_enter` must be paired with exactly one call to this function; a
+// missing or double call will leave the thread with an invalid stack or the
+// wrong PKRU.
+void ia2_callgate_exit(ia2_callgate_cookie cookie) {
+#if defined(__x86_64__)
+  if (cookie.saved_sp) {
+    __asm__ volatile("mov %0, %%rsp" : : "r"(cookie.saved_sp) : "memory");
```
ayrtonm, line 70: This line can also cause problems.

runtime/libia2/exit_callgates_x86_64.S
```diff
@@ -0,0 +1,82 @@
...
+    rdpkru
+    cmpl %edi, %eax
+    je 4f
+    ud2
+4:
+#endif
+
+    ret
+
+    .size ia2_callgate_enter, .-ia2_callgate_enter
```
ayrtonm, line 81: nit: This needs to be at the end of `ia2_callgate_enter`. In the second arg to the [`.size` directive](https://sourceware.org/binutils/docs/as/Size.html) "`.`" means "the address where the assembler will emit code next" so we calculate the size by subtracting the starting address of `ia2_callgate_enter` from that. That's what `.-ia2_callgate_enter` is doing. Doesn't really matter though since this only affects some ELF debug info and not codegen.

runtime/libia2/exit_callgates_x86_64.S
```diff
@@ -0,0 +1,82 @@
...
+// PKRU mask for the exit compartment: allow access to pkey 0 and pkey 1 only.
+// PKRU uses 2 bits per protection key (Access Disable, Write Disable).
+// Result: 0xfffffff0 (pkey 0 and 1 accessible, pkeys 2-15 blocked)
+#define IA2_EXIT_PKRU (~((3U << (2 * IA2_EXIT_COMPARTMENT_PKEY)) | 3U))
+
+    .text
+    .p2align 4,,15
+    .globl ia2_callgate_enter
+    .type ia2_callgate_enter,@function
+ia2_callgate_enter:
```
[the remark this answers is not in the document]
ayrtonm: Did you need to increase the `subq rsp` size? I think pushq decrements rsp as well?

runtime/libia2/exit_callgates_x86_64.c
```diff
@@ -0,0 +1,121 @@
...
+#define IA2_EXIT_COMPARTMENT_PKEY 1
+
+// PKRU mask for the exit compartment: allow access to pkey 0 and pkey 1 only.
+// PKRU uses 2 bits per protection key (Access Disable, Write Disable).
+// Result: 0xfffffff0 (pkey 0 and 1 accessible, pkeys 2-15 blocked)
+#define IA2_EXIT_PKRU (~((3U << (2 * IA2_EXIT_COMPARTMENT_PKEY)) | 3U))
+
+// Stringify helper macros for converting macro values to assembly immediates
+#define STR(x) #x
+#define XSTR(x) STR(x)
```
ayrtonm: Aren't these and some of the macros above defined in one of our headers? we should re-use those macros if possible.

runtime/libia2/exit_callgates_x86_64.c
```diff
@@ -0,0 +1,121 @@
...
+
+__asm__(
+    ".text\n"
+    ".p2align 4,,15\n"
```
ayrtonm: I don't think the asm code here needs to be aligned. You can probably remove it or you should add a comment explaining why this is here.

runtime/libia2/include/ia2_exit_callgates.h
```diff
@@ -0,0 +1,35 @@
...
+  uint32_t saved_pkru;
+  void *saved_sp;
+} ia2_callgate_cookie;
+
+/**
+ * Switches into the exit (libc) compartment and returns a cookie capturing the
+ * caller's PKRU and stack pointer so that the original compartment can resume
+ * once the exit call gate completes.
+ */
+ia2_callgate_cookie ia2_callgate_enter(void);
```
ayrtonm: If this and `ia2_callgate_exit` won't be called from C code then it doesn't a declaration.

runtime/libia2/include/ia2_internal.h
```diff
@@ -18,12 +18,17 @@ struct dl_phdr_info;
...
 #include <locale.h>
 #include <stdbool.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
 #include <unistd.h>
 
+#ifdef IA2_LIBC_COMPARTMENT
```
ayrtonm: All these `#ifdef IA2_LIBC_COMPARTMENT` should probably be `#if defined(IA2_LIBC_COMPARTMENT)` (it would make a difference when passing in `-DIA2_LIBC_COMPARTMENT=0`)

ayrtonm: I realize there's not much of a material difference on x86, but I think it more closely describes what you meant

tools/rewriter/SourceRewriter.cpp
```diff
@@ -1548,6 +1653,29 @@ int main(int argc, const char **argv) {
...
+    if (direct_call_wrappers.contains(ldso_fn)) {
+      // Always set pkey for ld.so functions to compartment 1
+      fn_decl_pass.fn_pkeys[ldso_fn] = 1; // ld.so functions run in compartment 1
+
+      if (!fn_decl_pass.fn_signatures.contains(ldso_fn)) {
+        FnSignature ldso_sig;
+        if (ldso_fn == "_dl_debug_state") {
+          // void _dl_debug_state(void) - simple function with no parameters
+          ldso_sig.api.ret.name = ""; // Return value has no name
+          ldso_sig.api.ret.type = 0; // Use TypeId 0 for void (basic type)
```
ayrtonm: is TypeId 0 -> void something we define or a clang thing?

tools/rewriter/GenCallAsm.cpp
```diff
@@ -683,15 +683,31 @@ static void emit_copy_args(AsmWriter &aw, const std::vector<ArgLocation> &args,
   }
 }
 
-static void emit_set_pkru(AsmWriter &aw, uint32_t target_pkey, Arch arch) {
+static void emit_set_pkru(AsmWriter &aw, uint32_t target_pkey, Arch arch, std::optional<uint32_t> union_pkey = std::nullopt) {
   if (arch == Arch::X86) {
-    // Change pkru to the compartment's value
-    add_comment_line(aw, "Set PKRU to the compartment's value");
+#ifdef IA2_LIBC_COMPARTMENT
```
ayrtonm: nit: it'd be nice to turn this compile-time flag into a run-time flag so there's only one build on the rewriter. You can use the `Arch` enum as a guide on how to do this.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -1596,10 +1740,41 @@ int main(int argc, const char **argv) {
...
+
+#ifdef IA2_LIBC_COMPARTMENT
+    const bool is_libc_compartment = compartment_pkey == IA2_LIBC_COMPARTMENT;
+
+    if (is_libc_compartment) {
+      // __wrap___cxa_finalize already switches into pkey 1, so the exit-compartment
+      // destructor runs with the right PKRU/stack. emit_asm_wrapper would assert on
+      // caller == target, yet ia2_compartment_init.inc still points both
+      // compartment_destructor_ptr (line 95) and the DT_FINI rewrites (lines 167-172)
+      // at __wrap_ia2_compartment_destructor_1. Keep the symbol alive with a single jmp.
```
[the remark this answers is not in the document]
oinoom: We’re already in the right compartment when the exit destructor runs, so emitting a call gate would just trip the caller == target assertion. The runtime still rewrites DT_FINI/.fini_array to call _wrap_ia2_compartment_destructor_1, so I kept the symbol by emitting a trivial wrapper that just jumps to the real destructor without switching compartments.

tools/rewriter/SourceRewriter.cpp
```diff
@@ -1576,6 +1704,12 @@ int main(int argc, const char **argv) {
                    << ") since its definition was not found by FnDecl pass\n";
       continue;
     }
+
+    // Skip generating wrapper if caller and target are in the same compartment
```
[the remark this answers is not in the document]
oinoom: It's because we're now iterating over the system headers and wrapping what we see. I've added a comment

ayrtonm:
A few minor nits, but otherwise looks fine

PR #649, opened 2025-12-11
auto-wrap dl functions

oinoom:
Adds functionality to wrap dynamic linker/libc functions automatically using the rewriter, as well as tests to confirm it works

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -32,6 +32,7 @@ if (NOT LIBIA2_AARCH64 AND NOT IA2_LIBC_COMPARTMENT)
 endif()
 
 if (IA2_LIBC_COMPARTMENT)
+    # Libc compartment tests only (x86-64 only, enforced by cmake_dependent_option)
     add_subdirectory(libc_default_compartment)
     add_subdirectory(dl_debug_test)
     add_subdirectory(dl_debug_getpwnam)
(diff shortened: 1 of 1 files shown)


cmake/ia2.cmake
```diff
@@ -304,6 +304,32 @@ function(add_ia2_call_gates NAME)
...
   endforeach()
 
+  # When --libc-compartment is enabled (and not on AArch64), append the loader
+  # stub TU (tools/rewriter/ldso_autowrap_stubs.c). It includes the public dl*
+  # headers so FnDecl/DetermineAbi can learn real ABI signatures for loader
+  # entrypoints from system headers instead of hard-coding them. We also add the
+  # stub to exactly one _compile_commands target so clang tooling sees a compile
+  # command for it.
+  set(LDSO_STUB_SOURCES "")
+  if(NOT LIBIA2_AARCH64)
```
ayrtonm: Can we just check the libc compartmentalization cmake flag instead of `NOT LIBIA2_AARCH64`?

ayrtonm (approved):
Rewriter changes and static verification tests look good to me

tests/three_keys_libc_autowrap/main.c
```diff
@@ -0,0 +1,23 @@
...
+extern void lib_a_noop(void);
+extern void lib_b_noop(void);
+
+void ia2_main(void) {
+    ia2_register_compartment("main", 1, NULL);
+    ia2_register_compartment("liblib_a.so", 2, NULL);
+    ia2_register_compartment("liblib_b.so", 3, NULL);
+}
+
+Test(three_keys_libc_autowrap, cross_compartment_calls) {
```
ayrtonm: Can we add another `Test` macro invocation that actually makes use of the generated callgates. For now it doesn't have to be something extensive like dlopen, but at least calling `dlsym`/`dladdr` from different compartments and verifying it works dynamically would be good.

PR #650, opened 2025-12-11
Cmake.improve

oinoom:


tests/CMakeLists.txt
```diff
@@ -91,18 +87,6 @@ if (NOT LIBIA2_AARCH64)
     add_subdirectory(two_shared_ranges)
     add_subdirectory(trusted_indirect)
 
-    if(IA2_LIBC_COMPARTMENT)
-        add_subdirectory(libc_default_compartment)
```
ayrtonm: Did you mean to drop these tests?

oinoom: yes they were added above, this was redundant

tests/CMakeLists.txt
```diff
@@ -32,10 +32,6 @@ if (NOT LIBIA2_AARCH64 AND NOT IA2_LIBC_COMPARTMENT)
 endif()
 
 if (IA2_LIBC_COMPARTMENT)
-    if (LIBIA2_AARCH64)
-        message(FATAL_ERROR "IA2_LIBC_COMPARTMENT is not supported on AArch64")
-    endif()
-
```
ayrtonm: The `return()` on line 48 needs to be removed. It's not doing anything right now, but could cause cmake to do the wrong thing depending on how this CMakeLists.txt gets included.

tests/CMakeLists.txt
ayrtonm: nvm, I see now that they were previously duplicated

ayrtonm (approved):
LGTM after addressing the two open comments.

PR #653, opened 2025-12-18
Add bootstrap shim and iconv test

oinoom:
Add LD_PRELOAD bootstrap shim that intercepts GLIBC_PRIVATE loader symbols and forwards them through loader gate wrappers.

Update dl_debug_test with iconv-based compartmentalization tests.


ayrtonm:
I see why this is necessary, but which DSO exactly calls __libc_dlopen_mode after ia2_start (the main binary, libc.so or ld.so)? I'm wondering if we can do something like redefine those ELF symbols instead of using LD_PRELOAD.

PR #659, opened 2026-02-02
Add --wrap=__cxa_finalize to rewriter-generated .ld files

oinoom:
When libc compartmentalization is enabled, the rewriter now adds --wrap=__cxa_finalize to all compartments' .ld files. This ensures that calls to __cxa_finalize are redirected to __wrap___cxa_finalize in libia2.a, which properly switches PKRU to enable pkey 1 access.

This fixes exit-time crashes in compartmentalized applications where __cxa_finalize (called during destructor execution) needs to access libc's internal data structures (e.g., __exit_funcs_lock) that are tagged with pkey 1.

Previously, the --wrap=__cxa_finalize flag was only set via CMake's INTERFACE property on libia2, which doesn't propagate to meson-based projects like dav1d that use the rewriter-generated .ld files.

### DIFF
--- a/tools/rewriter/SourceRewriter.cpp
+++ b/tools/rewriter/SourceRewriter.cpp
@@ -1843,6 +1843,17 @@ int main(int argc, const char **argv) {
                   "--wrap="s + fn_name + '\n', ".ld");
   }
 
+  // When libc compartmentalization is enabled, add --wrap=__cxa_finalize to all
+  // compartments' .ld files. The __wrap___cxa_finalize in libia2.a switches PKRU
+  // to allow pkey 1 access before calling the real __cxa_finalize, which is
+  // necessary because __cxa_finalize accesses libc's internal data structures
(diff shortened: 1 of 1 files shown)


fw-immunant (commented):
Does this fully supersede #658?

Should this replace the existing handling for this flag in `runtime/libia2/CMakeLists.txt`?

Otherwise, this fix makes sense. It's a bit messy that we have these flags flowing through the build system both as required flags and in the `.ld` files, but it would probably be even more confusing to have meson having to reason about whether libia2 was compiled with glibc compartmentalization.

oinoom:
> Does this fully supersede https://github.com/immunant/IA2-Phase2/pull/658?

@fw-immunant yes it fully supersedes it.

> Should this replace the existing handling for this flag in runtime/libia2/CMakeLists.txt?

AFAIK it's idempotent and harmless to keep both. It's primarily in the new place for external builds, which need it.

PR #666, opened 2026-02-10
Fix tracer pkey re-protection during init for libc compartment support

oinoom:
### Background

When `IA2_LIBC_COMPARTMENT` is enabled, the custom glibc's `__minimal_malloc` protects its entire heap (including TLS blocks) with pkey 1. During IA2 initialization, `protect_tls_pages` re-protects each compartment's TLS with the correct per-compartment pkey. The kernel handles this correctly via VMA splitting, but the tracer's Rust memory map tracking rejected the operation with overly strict monotonicity checks.

- **C policy** (`is_op_permitted` in `track_memory_map.c`): already allows compartment 0 to `pkey_mprotect` any memory during init
- **Rust map tracking** (`memory_map_pkey_mprotect_region` in `lib.rs`): rejected it because the region was already owned by another pkey or already `pkey_mprotect`'d

### Changes

Relax the Rust monotonicity checks before `init_finished`, matching the existing pattern for regular `mprotect`. During init:
- Allow re-protecting memory owned by another compartment (needed because the loader's pkey 1 heap contains TLS for all compartments)
- Allow repeated `pkey_mprotect` on the same region
- Update `owner_pkey` to the new pkey during init
- Remove the commented-out cmake workaround that disabled the sandbox for all libc compartment tests
- Run `loader_minimal_malloc` without the sandbox since it reads `/proc/self/smaps`, which is not allowed by landlock (the sandbox's filesystem walk skips `/proc/self` because it is a symlink)

### DIFF
--- a/cmake/define-test.cmake
+++ b/cmake/define-test.cmake
@@ -114,11 +114,6 @@ function(define_test)
         set(DEFINE_TEST_WITHOUT_SANDBOX TRUE)
     endif()
 
-    # Custom loader with MPK-protected heap is incompatible with sandbox
-    # (sandbox doesn't expect pre-existing pkey-protected regions)
-    # if (LIBIA2_REBUILD_GLIBC AND NOT LIBIA2_AARCH64)
-    #     set(DEFINE_TEST_WITHOUT_SANDBOX TRUE)
-    # endif()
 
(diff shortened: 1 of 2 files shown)


fw-immunant (approved):
CMake change needs cleanup, but otherwise looks good.

PR #671, opened 2026-03-05
libia2: keep TCB TLS page shared across compartments

oinoom:
### Problem

IA2 retags writable memory, including `PT_TLS`, to per-compartment pkeys. On x86_64, that model is too strict for TLS pages that hold process ABI state rather than compartment-private state.

In particular, the thread-pointer/TCB page (`__builtin_thread_pointer()`, `%fs` ABI state) can be accessed in normal execution and transition paths where current PKRU does not yet match the compartment that originally tagged the page. When that page is left compartment-tagged, this can fault in real workloads (including decode-path crashes where stack-protector canary reads use `%fs`).

### Fix

`protect_tls_pages()` now uses an x86_64 split policy:

- Default: tag module `PT_TLS` ranges with the target compartment pkey
- Shared carve-out 1: `ia2_stackptr_0` page (existing call-gate stack slot behavior)
- Shared carve-out 2: thread-pointer/TCB page (page containing `%fs` ABI state)

Implementation details:

- Maintain a bounded carve-out set (`MAX_SHARED_TLS_PAGES = 2`) and only add carve-outs that overlap the current module TLS range
- Sort carve-out pages and protect non-shared regions using a cursor walk to avoid overlap and ordering errors
- Restore shared carve-out pages to `pkey 0` during process-init path (`ia2_get_compartment() == 0`)
- Skip redundant post-init per-thread re-tagging to `pkey 0`; those pages are already shared and repeated retagging from compartment 1 can violate tracer policy
- Keep the explicit startup hardening in `ia2_start()` that re-applies TCB sharing after compartment setup
- Preserve pre-existing AArch64 behavior; TCB carve-out logic is x86_64-specific

### Similar Patterns in IA2

- Special-case shared TLS behavior already existed for `ia2_stackptr_0` in `protect_tls_pages()` (see `runtime/libia2/ia2.c`, `protect_tls_pages`, around lines 466-473)
- `protect_pages()` already uses a "protect-by-default with explicit shared carve-outs" model for writable ranges (see `runtime/libia2/ia2.c`, `protect_pages`, shared-range collection around lines 603-632, and overlap trimming before `ia2_mprotect_with_tag` around lines 685-723)
- RELRO and explicit shared sections are already handled as policy exceptions to default tagging (see `runtime/libia2/ia2.c`, explicit `shared_sections` handling around lines 605-632 and `PT_GNU_RELRO` handling around lines 634-645)

### DIFF
--- a/runtime/libia2/ia2.c
+++ b/runtime/libia2/ia2.c
@@ -390,6 +390,63 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
       exit(-1);
     }
 
+    // Look for the untrusted stack pointer, in case this lib defines it.
+    extern __thread void *ia2_stackptr_0;
+    uint64_t untrusted_stackptr_addr = (uint64_t)&ia2_stackptr_0;
+    if ((untrusted_stackptr_addr & 0xFFF) != 0) {
+      printf("address of ia2_stackptr_0 (%p) is not page-aligned\n",
+             (void *)untrusted_stackptr_addr);
+      exit(-1);
+    }
+#if defined(__aarch64__)
+    // Preserve the pre-existing AArch64 behavior exactly. The TCB carve-out
+    // changes are x86_64-specific and should not alter ARM TLS tagging.
+    if (untrusted_stackptr_addr >= start && untrusted_stackptr_addr < end) {
+      // The TLS region should only be split for compartment 1.
+      assert(pkey == 1);
+
+      // Protect TLS region start to the beginning of the untrusted region.
+      if (untrusted_stackptr_addr > start_round_down) {
+        int mprotect_err = ia2_mprotect_with_tag(
+            (void *)start_round_down, untrusted_stackptr_addr - start_round_down,
+            PROT_READ | PROT_WRITE, pkey);
+        if (mprotect_err != 0) {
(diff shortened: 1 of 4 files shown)


runtime/libia2/ia2.c
```diff
@@ -111,6 +136,10 @@ size_t ia2_get_compartment(void) {
   return ia2_get_tag();
 }
 
+// x18-tagged aarch64 builds do not currently use the x86_64 TCB/%fs ABI path.
+// Keep this symbol for cross-arch call-site symmetry.
+void ia2_unprotect_thread_pointer_page(void) {}
```
ayrtonm, line 141: Might be better to remove this no-op since it could easily get missed later on and instead just add `#if defined(__x86_64__)` around this function's callsite.

runtime/libia2/ia2.c
```diff
@@ -333,6 +362,8 @@ int protect_tls_pages(struct dl_phdr_info *info, size_t size, void *data) {
   struct ia2_thread_metadata *const thread_metadata = ia2_thread_metadata_get_for_current_thread();
 #endif
 
+  enum { MAX_SHARED_TLS_PAGES = 2 };
```
ayrtonm, line 365: nit: I don't think an enum provides any benefit for a single value so maybe make this `const int` or whatever type to avoid the implicit backing type

Issue #667, opened 2026-02-10
Intermittent `three_keys_minimal` test failure under tracer


fw-immunant:
This happened on CI for both #655 and #595. I've managed to reproduce it locally on `main` (62017114678a04caf1d6db9a53f5ac81c48a359f) by running the test repeatedly: `while ctest -R three_keys_minimal; do :; done`.

We see the tracer first forbidding a component from mmapping over an existing region held by compartment 1, then the tracer aborting (taking its OOM codepath) when the `mmap` call fails:

```
forbidden operation requested: compartment 3 mmap (321000000000, 17179869184, prot=0, flags=22, fd=-1)
region pkey: 1
forbidden operation requested: compartment 3 mmap (321000000000, 17179869184, prot=0, flags=22, fd=-1)
region pkey: 1
[FATAL:partition_address_space.cc(76)] Check failed: false. 
```

Full output:
```
ia2 warning: rebuild runtime to use landlock features from ABI version 7 instead of 5
attempting to mprotect unmapped memory
attempting to mprotect unmapped memory
attempting to mprotect unmapped memory
attempting to mprotect unmapped memory
attempting to pkey_mprotect unmapped memory
attempting to pkey_mprotect unmapped memory
running suite 'three_keys_minimal' test 'direct_calls', expecting exit status 0 (EXIT_SUCCESS)...
ia2_start: initializing ia2 runtime
ia2_register_compartment: registered main for compartment #1
ia2_register_compartment: registered liblib_2.so for compartment #2
ia2_register_compartment: registered liblib_1.so for compartment #3
allocate_stack: allocating stack for compartment 0 on thread 903229: 0x7ffff6e00000..0x7ffff7200000
ia2_setup_compartment: protecting memory for compartment 1
ia2_mprotect_with_tag(addr=0x7ffff6a00000, len=4194304, prot=3, tag=1)
allocate_stack: allocating stack for compartment 1 on thread 903229: 0x7ffff6a00000..0x7ffff6e00000
protect_pages: protecting library: 
protect_pages: Shared range 0: 0x5555555e0000-0x555555608000
protect_pages: Segment 4: 0x5555555d3000-0x5555555d4000
protect_pages: Shared range 1 overlaps start of segment 4, adjusting start to 0x5555555d4000
protect_pages: Segment 5: 0x5555555d4000-0x555555f4a000
protect_pages: Shared range 0 overlaps end of segment 5, adjusting end to 0x5555555e0000
[...]

PR #675, opened 2026-03-20
tracer: allow non-fixed mmap hints; add deterministic regression test

oinoom:
## Summary

This PR fixes a flaky failure in `three_keys_minimal` #667 by correcting tracer mmap policy for non-`MAP_FIXED` mappings and adds a deterministic regression test that fails before the fix and passes after it

## Problem

`three_keys_minimal` intermittently failed in CI with tracer output like:

- `forbidden operation requested: compartment 2/3 mmap (..., prot=0, flags=22, fd=-1)`
- `region pkey: 1`
- followed by partition-alloc fatal behavior (for example `Check failed: false`) and test failure

Observed failing pattern:

- `flags=22` is `0x22` (`MAP_PRIVATE | MAP_ANONYMOUS`) and does **not** include `MAP_FIXED`
- The requested address is therefore a hint, not a strict placement requirement

## Root Cause

In `runtime/tracer/track_memory_map.c`, `EVENT_MMAP` policy only auto-allowed non-fixed mmaps when `addr == 0`:

- Non-`MAP_FIXED` + `addr==0`: allowed
- Non-`MAP_FIXED` + `addr!=0`: overlap-checked as if the hinted address were authoritative

That is incorrect for non-fixed mappings. The kernel may ignore a non-fixed hint and place the mapping elsewhere. Treating the hint as mandatory caused false denials whenever the hint overlapped memory owned by another compartment.

## Fix

In `is_op_permitted(EVENT_MMAP)`, allow all non-`MAP_FIXED` mmaps unconditionally:

- Old behavior: allow only if `start == 0` and not `MAP_FIXED`
- New behavior: allow whenever not `MAP_FIXED`

Why this remains safe in tracer implementation:

- For `mmap`, the tracer uses the returned address from syscall exit, not the entry-time hint: it runs to syscall exit (`runtime/tracer/track_memory_map.c:1175`), reads registers (`runtime/tracer/track_memory_map.c:1227`), and sets `info->range.start = reg_retval` for `EVENT_MMAP` (`runtime/tracer/track_memory_map.c:447`, `runtime/tracer/track_memory_map.c:451`)
- The tracked memory map is then updated using that returned address
- If map update cannot be represented consistently, tracer fails closed with an error

## Deterministic Regression Test

Added a deterministic regression test in `three_keys_minimal`:

- New helper in compartment 2:
  - `lib_2_mmap_nonfixed_hint(void *hint)` in `tests/three_keys_minimal/lib_2.c`
  - Declaration in `tests/three_keys_minimal/include/lib_2/lib_2.h`
- New test:
  - `nonfixed_mmap_hint_foreign_compartment` in `tests/three_keys_minimal/main.c`
  - Uses a hint address inside main compartment private mapping (`main_get_static()` page)
  - Issues non-`MAP_FIXED` anonymous mmap and expects success

This test is deterministic because the hint intentionally targets foreign-compartment-owned space, which old policy always denied and new policy always allows.

### DIFF
--- a/runtime/tracer/track_memory_map.c
+++ b/runtime/tracer/track_memory_map.c
@@ -43,8 +43,10 @@ static bool is_op_permitted(struct memory_map *map, int event,
                             union event_info *info) {
   switch (event) {
   case EVENT_MMAP:
-    // non-FIXED maps of NULL don't need to check for overlap
-    if (info->mmap.range.start == 0 && !(info->mmap.flags & MAP_FIXED)) {
+    /* Non-MAP_FIXED mmap addresses are just hints; the kernel can place the
+     * mapping elsewhere. So we must not reject them based on overlap at the
+     * hinted address (which can be spuriously in another compartment). */
+    if (!(info->mmap.flags & MAP_FIXED)) {
       return true;
     }
     if (memory_map_all_overlapping_regions_have_pkey(map, info->mmap.range,
--- a/runtime/tracer/track_memory_map.c
+++ b/runtime/tracer/track_memory_map.c
@@ -43,9 +43,9 @@ static bool is_op_permitted(struct memory_map *map, int event,
                             union event_info *info) {
   switch (event) {
   case EVENT_MMAP:
-    /* Non-MAP_FIXED mmap addresses are just hints; the kernel can place the
-     * mapping elsewhere. So we must not reject them based on overlap at the
-     * hinted address (which can be spuriously in another compartment). */
+    /* A non-MAP_FIXED mmap hint must not replace an existing mapping. Since
+     * the hinted address is not binding, we must not reject it based on
+     * overlap at that hinted location (which may be in another compartment). */
     if (!(info->mmap.flags & MAP_FIXED)) {
       return true;
     }
--- a/tests/three_keys_minimal/include/lib_2/lib_2.h
+++ b/tests/three_keys_minimal/include/lib_2/lib_2.h
@@ -3,3 +3,15 @@
 #include "lib.h"
 
 DECLARE_LIB(2, 1);
+
+/**
+ * Issue a non-MAP_FIXED anonymous mmap using an address hint from another
+ * compartment.
+ *
+ * This is used by the regression test for tracer mmap policy. The kernel may
+ * ignore a non-fixed hint and place the mapping elsewhere, so this operation
+ * should be allowed by policy.
+ *
+ * @return 0 on success, or -errno on failure.
+ */
+int lib_2_mmap_nonfixed_hint(void *hint);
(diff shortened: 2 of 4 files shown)


runtime/tracer/track_memory_map.c
```diff
@@ -43,8 +43,10 @@ static bool is_op_permitted(struct memory_map *map, int event,
                             union event_info *info) {
   switch (event) {
   case EVENT_MMAP:
-    // non-FIXED maps of NULL don't need to check for overlap
-    if (info->mmap.range.start == 0 && !(info->mmap.flags & MAP_FIXED)) {
+    /* Non-MAP_FIXED mmap addresses are just hints; the kernel can place the
+     * mapping elsewhere. So we must not reject them based on overlap at the
```
fw-immunant: Specifically, what matters is that the kernel will *not* overwrite an existing mapping when `MAP_FIXED` is not passed. E.g., man(3) mmap says:
> When MAP_FIXED is not set, the implementation uses _addr_ in an implementation-defined manner to arrive at _pa_.  The _pa_ so chosen shall be an area of the address space that the implementation deems suitable for a mapping of _len_ bytes to the file. All implementations interpret an _addr_ value of 0 as granting the implementation complete freedom in selecting _pa_, subject to constraints described below. A non-zero value of _addr_ is taken to be a suggestion of a process address near which the mapping should be placed. **When the implementation selects a value for _pa_, it never places a mapping at address 0, nor does it replace any extant mapping.**

What's critical is this promise, not the kernel's freedom of choice.

tests/three_keys_minimal/include/lib_2/lib_2.h
```diff
@@ -3,3 +3,15 @@
...
 
 DECLARE_LIB(2, 1);
+
+/**
+ * Issue a non-MAP_FIXED anonymous mmap using an address hint from another
+ * compartment.
+ *
+ * This is used by the regression test for tracer mmap policy. The kernel may
+ * ignore a non-fixed hint and place the mapping elsewhere, so this operation
+ * should be allowed by policy.
```
fw-immunant: Again, the important thing is that this is guaranteed not to disrupt an existing mapping, not that the kernel may place the mapping elsewhere.

fw-immunant (changes requested):
Some reasoning in comments needs to be corrected, but this fix is much preferable to what I had prototyped in #667, and looks to be more agnostic to the particular case being exercised by PartitionAlloc here.

oinoom:
@fw-immunant I changed the commentary in https://github.com/immunant/IA2-Phase2/pull/675/commits/5f7c968178d5812261ba910466df66a820cb68f5, let me know if that doesn't cut it.

fw-immunant (approved):

Issue #283, opened 2023-09-20
DTV can be modified by any compartment


fw-immunant:
This came up a few weeks ago but I wanted to get an issue filed for visibility.

TLS is accessed via the dynamic thread vector (DTV), whose address lives at `%fs:DTV_OFFSET` on x86_64; this vector contains pointers to the start of the TLS region for each module (which means each compartment for IA2).

The DTV is allocated during `pthread_create` before the new thread starts executing. If partition-alloc does this, it will tie the DTV to the compartment that starts the new thread, allowing it to corrupt TLS base addresses of other comparmtents.

From Slack:
> I think we can probably get away with making the DTV immutable after thread creation--the prerequisites here aiui are going to be loading modules upfront (which we already require) and eagerly allocating TLS blocks on thread startup (to have glibc set up all entries of the DTV) which we do with a loop around `__tls_get_addr`. when creating threads we may have to do a dance like "have glibc allocate all TLS blocks, make DTV immutable, check if it looks like it should, abort if not"... we'll be duplicating some glibc internals (and depending on matching them) but we kind of already depend on glibc implementation details for our TLS protection anyway

This gives us the outline for our design:
1. add allocator flag to force page-align and page-size
    - how do we prevent cross-thread and cross-compartment interference with this flag?
3. set flag around `pthread_create` to force alignment of dtv allocation
4. `pkey_mprotect` dtv to compartment 0 read-only
5. validate the DTV and abort if not as expected
    - how do we know what the DTV should look like? one possibility is to back up the main thread's DTV to a known read-only region and compare against that. new threads' DTVs will store different addresses but the %fs-relative offsets of those addresses should match those from the main thread's DTV.

fw-immunant:
> **how do we prevent cross-thread and cross-compartment interference with this flag?**

I believe this is the only open question we need to answer to implement this.

PR #677, opened 2026-03-31
Targeted TCB Carveout 

oinoom:
# Problem
IA2 retags writable memory (including PT_TLS-backed TLS pages) to compartment pkeys.
On x86_64, the page containing `__builtin_thread_pointer()` (the TCB page addressed through `%fs`) holds process ABI state that can be accessed while different compartment PKRU values are active.

That ABI state includes stack-protector canary reads such as `%fs:0x28`, which occur in normal function prologues/epilogues and are not tied to a single compartment's policy boundary.
If this page remains compartment-tagged, code can fault when execution reaches one of these ABI reads under a PKRU that cannot access that page.

Observed on `main` during single-thread dav1d decode:
- crash site: `dav1d_ref_create+16`
- faulting instruction: `mov %fs:0x28,%rax`
- fault class: PKRU permission failure on `%fs`-relative TCB access

# Fix In This Branch
This branch applies a minimal x86_64-only startup hardening change (and replaces #671):
- Keep existing `protect_tls_pages()` logic unchanged.
- Add `ia2_unprotect_thread_pointer_page()` that:
  - computes `tcb_page = round_down(__builtin_thread_pointer(), PAGE_SIZE)`
  - retags exactly that one page to shared pkey 0 via `ia2_mprotect_with_tag()`
- Call that helper in `ia2_start()` after compartment setup/TLS retagging has run, before init completion.
- Keep AArch64 behavior unchanged (no-op stub for symbol parity).

# What This Branch Accomplishes
- Removes the first strict-mode `dav1d` decode blocker on `main` (the `%fs:0x28` TCB access fault).
- Demonstrates forward progress to the next independent blocker (`__tls_get_addr` / `_rtld_local`).

# How This Differs From #671 
#671 widened runtime TLS policy in `protect_tls_pages()` (bounded multi-page carveout set, sorted carveout walk, per-thread carveout handling, redundant-retag avoidance).

This branch intentionally does **not** do that.
It is a narrower replacement branch that changes only startup TP-page treatment and leaves `protect_tls_pages()` as in `main`.

Practical effect:
- broader branch: attempts a generalized shared-TLS carveout policy
- this branch: fixes only the x86_64 TP-page visibility blocker from first principles

### DIFF
--- a/runtime/libia2/ia2.c
+++ b/runtime/libia2/ia2.c
@@ -129,8 +129,6 @@ size_t ia2_get_compartment(void) {
   return ia2_get_tag();
 }
 
-void ia2_unprotect_thread_pointer_page(void) {}
-
 // TODO: insert_tag could probably be cleaned up a bit, but I'm not sure if the
 // generated code could be simplified since addg encodes the tag as an imm field
 #define _addg(out_ptr, in_ptr, tag) \
--- a/runtime/libia2/ia2.c
+++ b/runtime/libia2/ia2.c
@@ -97,6 +97,24 @@ size_t ia2_get_compartment(void) {
   }
 }
 
+/*
+ * Keep the page containing the x86_64 thread pointer shared.
+ *
+ * Stack-protector and other ABI-sensitive accesses read %fs-relative data
+ * (for example, %fs:0x28) and must remain valid across compartment PKRU
+ * transitions.
+ */
+void ia2_unprotect_thread_pointer_page(void) {
+  uintptr_t tcb_page =
+      IA2_ROUND_DOWN((uintptr_t)__builtin_thread_pointer(), PAGE_SIZE);
+  int err = ia2_mprotect_with_tag(
+      (void *)tcb_page, PAGE_SIZE, PROT_READ | PROT_WRITE, 0);
+  if (err != 0) {
+    printf("ia2_mprotect_with_tag failed: %s\n", strerror(errno));
+    exit(-1);
+  }
+}
+
 #elif defined(__aarch64__)
 
 static size_t ia2_get_x18(void) {
@@ -110,6 +128,8 @@ size_t ia2_get_tag(void) __attribute__((alias("ia2_get_x18")));
 size_t ia2_get_compartment(void) {
   return ia2_get_tag();
 }
+
+void ia2_unprotect_thread_pointer_page(void) {}
 
 // TODO: insert_tag could probably be cleaned up a bit, but I'm not sure if the
 // generated code could be simplified since addg encodes the tag as an imm field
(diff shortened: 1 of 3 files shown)


ayrtonm:
I think this is a better approach than #671.

runtime/libia2/ia2.c
```diff
@@ -111,6 +129,8 @@ size_t ia2_get_compartment(void) {
   return ia2_get_tag();
 }
 
+void ia2_unprotect_thread_pointer_page(void) {}
```
ayrtonm: Might be better to remove this no-op since it could easily get missed later on and instead just add `#if defined(__x86_64__)` around this function's callsite.

ayrtonm (approved):
LGTM with a small change

fw-immunant:
What else lives on this page? We haven't seen a problematic interaction here before, so presumably this only happens when stack canaries are enabled (does the build tooling for dav1d do this someplace obvious?). Shouldn't we make this carveout at least conditional so that we aren't widening our attack surface when stack canaries are not used?

If data that we really do need protected lives on this page, maybe we could do something like swallowing segfaults on this particular address in the tracer.

oinoom:
@fw-immunant 

> What else lives on this page?

Things like `tcb` pointer, `dtv` pointer, `stack_guard`, and `pointer_guard`

> presumably this only happens when stack canaries are enabled 

`libdav1d.so` shows `%fs:0x28` accesses and `__stack_chk_fail` without this change, so yes this seems stack canary related.

> does the build tooling for dav1d do this someplace obvious?

It doesn't seem to do this explicitly, the stack canary usage seems to be a default (e.g. my local `cc` injects `-fstack-protector-strong` on its own). 

> Shouldn't we make this carveout at least conditional so that we aren't widening our attack surface when stack canaries are not used?

I'm not sure what this should be conditional on, usage of the stack canary?

> maybe we could do something like swallowing segfaults on this particular address in the tracer.

is this an alternative to a conditional carve out?

fw-immunant:
> @fw-immunant
> 
> > What else lives on this page?
> 
> Things like `tcb` pointer, `dtv` pointer, `stack_guard`, and `pointer_guard`

These are definitely things that we don't want a compromised compartment to be able to modify for other compartments, so we need to do something here. One possibility is sketched in #283, which specifically discusses the DTV but probably goes for TCB as the whole, too.

> > does the build tooling for dav1d do this someplace obvious?
> 
> It doesn't seem to do this explicitly, the stack canary usage seems to be a default (e.g. my local `cc` injects `-fstack-protector-strong` on its own).

Interesting... on Arch, I locally see `--enable-default-ssp` in `gcc -v`, but I don't see it on donna. Despite that flag being present on Arch, I haven't seen test failures locally, so I'm curious why this is an issue for dav1d but not our tests. Maybe dav1d includes functions that meet the criteria for stack protectors, but our tests don't (gcc docs mention stack arrays)?

> > Shouldn't we make this carveout at least conditional so that we aren't widening our attack surface when stack canaries are not used?
> 
> I'm not sure what this should be conditional on, usage of the stack canary?

Yeah, if the only problematic accesses here are from a non-IA2 mitigation then we shouldn't simply assume that mitigation is enabled but have some explicit compatibility with it enabled exactly when necessary.

> 
> > maybe we could do something like swallowing segfaults on this particular address in the tracer.
> 
> is this an alternative to a conditional carve out?

Yes, assuming it works. But if it's sufficient to make this region immutable as suggested in #283, then doing that would probably be preferable.

